PR #7, opened 2019-02-01
PMC Submitter Not Imported Only

jhoanmanuelms:
- Properly process the errors in submission to log the problem instead of failing all the process
- Add a property to indicate the source file that will be used to submit
- Add logic to process only the submissions for a specific resource file that haven't been submitted yet
- Mark the submission as submitted in the Mongo DB once the process is finished
- Fix XML serializer unit test
- Remove unnecessary code

### DIFF
--- a/scheduler/pmc-processor-task/src/main/kotlin/ac/uk/ebi/pmc/data/MongoRepository.kt
+++ b/scheduler/pmc-processor-task/src/main/kotlin/ac/uk/ebi/pmc/data/MongoRepository.kt
@@ -5,10 +5,12 @@ import ac.uk.ebi.pmc.data.docs.FileDoc
 import ac.uk.ebi.pmc.data.docs.SubmissionDoc
 import com.mongodb.async.client.FindIterable
 import com.mongodb.async.client.MongoClient
-import com.mongodb.client.model.Filters
+import com.mongodb.client.model.Filters.and
+import com.mongodb.client.model.Filters.eq
 import org.bson.types.ObjectId
 import org.litote.kmongo.coroutine.findOne
 import org.litote.kmongo.coroutine.insertOne
+import org.litote.kmongo.coroutine.updateOne
 import java.io.File
 
 private const val SUBMISSION_COLLECTION = "submissions"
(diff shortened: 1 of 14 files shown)


scheduler/pmc-processor-task/src/main/kotlin/ac/uk/ebi/pmc/data/MongoRepository.kt
```diff
@@ -19,36 +21,32 @@ class MongoRepository(
     private val dataBase: String,
     private val mongoClient: MongoClient
 ) {
-    fun getAllSubmissions(): FindIterable<SubmissionDoc> {
-        val database = mongoClient.getDatabase(dataBase)
-        val collection = database.getCollection(SUBMISSION_COLLECTION, SubmissionDoc::class.java)
+    fun getNotImportedSubmissionsBySourceFile(sourceFile: String): FindIterable<SubmissionDoc> =
```
Juan-EBI: I think this method signature can be

```
findSubmission(source:String, Imported = false)
```

IMHO find plus kotlin named parameters make clear enought what does method do

scheduler/pmc-processor-task/src/test/kotlin/ac/uk/ebi/pmc/importer.data/MongoDocServiceTest.kt
```diff
@@ -36,7 +36,7 @@ internal class MongoDocServiceTest(
     @Test
     fun `save submission with file`(): Unit = runBlocking {
         coEvery { dataRepository.saveSubFile(exampleFile, submission.accNo) } returns fileId
-        coEvery { dataRepository.save(ofType(SubmissionDoc::class)) } returns Unit
+        coEvery { dataRepository.save(ofType(SubmissionDoc::class)) } returns null
```
Juan-EBI: why null now instead of void?

scheduler/pmc-processor-task/src/main/kotlin/ac/uk/ebi/pmc/data/MongoRepository.kt
```diff
@@ -19,36 +21,32 @@ class MongoRepository(
...
-        return ids.map { collection.findOne(Filters.eq("_id", it))!! }.toList()
+        return ids.map { collection.findOne(eq("_id", it))!! }.toList()
     }
+
+    private fun getFilesCollection() = getCollection(FILES_COLLECTION, FileDoc::class.java)
+
+    private fun getSubmissionCollection() = getCollection(SUBMISSION_COLLECTION, SubmissionDoc::class.java)
+
+    private fun <A, B : Class<A>> getCollection(name: String, documentClass: B) =
+            mongoClient.getDatabase(dataBase).getCollection(name, documentClass)
```
Juan-EBI: this one is the perfect case for kotlin reified

```
inline fun <reified T> getCollection(name:String)
```

in that way it can be called like

getCollection<Submission/Doc>("Submissions")

Juan-EBI:
please check minor comments before merge

scheduler/pmc-processor-task/src/test/kotlin/ac/uk/ebi/pmc/importer.data/MongoDocServiceTest.kt
jhoanmanuelms: Because I changed the save method to be a one-liner so I have to use the Mongok `save` method returning type which is `Void?`.

Since it's a nullable type, I just return null for this tests case.

PR #18, opened 2019-02-08
Feature/base pmc feature

Juan-EBI:
- refactor to use 3 diferent task : load, process, submit


scheduler/pmc-processor-task/src/main/kotlin/ac/uk/ebi/pmc/config/AppConfig.kt
```diff
@@ -32,11 +36,22 @@ class AppConfig {
...
+    @Bean
+    fun submissionFileRepository(client: MongoClient) = SubFileRepository(client.getCollection("eubioimag", "files"))
+
+    @Bean
+    fun submissionDocService(
+        serializationService: SerializationService,
+        errorsRepository: ErrorsRepository,
+        submissionRepository: SubRepository,
+        submissionFileRepository: SubFileRepository
+    ) =
```
jhoanmanuelms: same line <?>

scheduler/scheduler/README.md
```diff
@@ -1,3 +1,3 @@
 # Scheduler
-Schedule job to be executed into the lsf cluster. Jobs are basically java application that run for a limited time and execute specific task.
- 
\ No newline at end of file
+Provide a rest api that allow the schedule jobs to be executed into the lsf cluster. Jobs are basically java application that run for a limited time and execute specific task.
```
jhoanmanuelms: I'd suggest: that allows to schedule jobs that will be executed

scheduler/pmc-processor-task/src/main/kotlin/ac/uk/ebi/pmc/PmcProcessorApp.kt
```diff
@@ -12,22 +13,29 @@ import java.io.File
...
+        pmcSubmissionLoader: PmcSubmissionLoader
+    ) =
+        TaskExecutor(properties, importer, submitter, pmcSubmissionLoader)
 }
 
 class TaskExecutor(
     private val properties: PmcImporterProperties,
-    private val batchImporter: PmcBatchImporter,
-    private val pmcBatchSubmitter: PmcBatchSubmitter
+    private val pmcSubProcessor: PmcSubProcessor,
```
jhoanmanuelms: This is just something I was thinking but, I see that we use indistinctively "Sub" and "Submission" to refer to the same thing. We should pick one or the other and stick to it (I'd prefer if we go with the latter).

I just say this because, for example, in this case, this variable called `pmcSubProcessor` might refer to a processor for PMC submissions but it might also be interpreted as a PMC sub-process (as in a process that's part of another process).

The same thing could apply for other parts of the code. I know I might be overcomplicating here but better take this decision now that we're refactoring 😅

PR #24, opened 2019-02-14
Files Table Template Transpiler

jhoanmanuelms:
- Create a component to transpile from a files table template to its equivalent page tab in the desired format
- Refactor TSV serializer to handle specific elements instead of full submission


Juan-EBI:
looks good to me please consider:

1. Add kdoc to main method explaining behavior.
2. Remove interfaces and abstracts class , I do not see our-self in the short-mid time adding several implementation and best moment to add code is when needed.

jhoanmanuelms:
I applied your comments 😬

PR #30, opened 2019-02-25
TSV Serializer Tests

jhoanmanuelms:
- Add tests for TSV serializer
- Add value attributes to table headers

### DIFF
--- a/commons/commons-bio/src/main/kotlin/ebi/ac/uk/model/TableModel.kt
+++ b/commons/commons-bio/src/main/kotlin/ebi/ac/uk/model/TableModel.kt
@@ -1,6 +1,7 @@
 package ebi.ac.uk.model
 
 import ebi.ac.uk.base.isNotBlank
+import ebi.ac.uk.model.constants.SectionFields
 import ebi.ac.uk.model.constants.TableFields
 import ebi.ac.uk.model.extensions.parentAccNo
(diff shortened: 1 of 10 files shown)


commons/commons-bio/src/main/kotlin/ebi/ac/uk/model/TableModel.kt
```diff
@@ -68,11 +69,16 @@ abstract class Row<T>(val original: T) {
 
     override fun hashCode() = Objects.hash(original)
 
-    fun headers() = attributes.map { Header(it.name, it.nameAttrs.map(AttributeDetail::name)) }
+    fun headers() = attributes.map {
+        Header(it.name, it.nameAttrs.map(AttributeDetail::name), it.valueAttrs.map(AttributeDetail::name))
```
Juan-EBI: I am wondering if is worth it to create attribute extension properties to simplify code here. something like 

```
    val nameDetailsNames: List<String>
        get() = nameAttrs.map(AttributeDetail::name)
```

so it becomes

` Header(it.name, it.nameDetailsNames, it.valueDetailsNames`

jhoanmanuelms: Done

PR #38, opened 2019-03-12
Section Extended

jhoanmanuelms:
- Refactor submission model to use library file as a string element for simple model and as a LibraryFile object for the extended model
- Remove library file object from serialization since it's now an attribute
- Adjust unit tests

### DIFF
--- a/commons/commons-bio/src/main/kotlin/ebi/ac/uk/model/ExtendedSection.kt
+++ b/commons/commons-bio/src/main/kotlin/ebi/ac/uk/model/ExtendedSection.kt
@@ -1,0 +1,42 @@
+package ebi.ac.uk.model
+
+import arrow.core.Either
+import ebi.ac.uk.model.extensions.libraryFile
+import ebi.ac.uk.util.collections.addLeft
+import java.util.Objects
+
+class ExtendedSection(type: String) : Section(type) {
+    var libraryFile: LibraryFile? = null
+    var extendedSections: MutableList<Either<ExtendedSection, SectionsTable>> = mutableListOf()
+
+    constructor(section: Section) : this(section.type) {
+        accNo = section.accNo
+        files = section.files
+        links = section.links
+        sections = section.sections
+        attributes = section.attributes
+
+        section.libraryFile?.let { libraryFile = LibraryFile(it) }
+        section.sections.forEach { sect ->
+            sect.fold({ extendedSections.addLeft(ExtendedSection(it)) }, { addSectionTable(it) })
+        }
+    }
+
+    fun addReferencedFile(file: File) = libraryFile?.addFile(file)
+
+    override fun equals(other: Any?) = when {
+        other !is ExtendedSection -> false
+        other === this -> true
+        else -> Objects.equals(type, other.type)
+            .and(Objects.equals(accNo, other.accNo))
+            .and(Objects.equals(files, other.files))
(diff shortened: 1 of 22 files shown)


commons/commons-bio/src/main/kotlin/ebi/ac/uk/model/extensions/SectionExt.kt
```diff
@@ -1,10 +1,32 @@
...
+        value?.let { this[SectionFields.LIB_FILE] = it }
+    }
 
-fun Section.allReferencedFiles() = libraryFile?.referencedFiles?.toList() ?: listOf()
+fun Section.allFiles() = files.map { it.fold({ file -> listOf(file) }, { table -> table.elements }) }.flatten()
 
 fun Section.allSections() =
     sections.map { it.fold({ section -> listOf(section) }, { table -> table.elements }) }.flatten()
+
+fun ExtendedSection.allReferencedFiles(): List<File> {
```
Juan-EBI: I think is worth to seprate `SectionExt` from `ExtendedSectionExt`

commons/commons-bio/src/main/kotlin/ebi/ac/uk/model/extensions/SubExt.kt
```diff
@@ -46,6 +47,8 @@ var Submission.title: String?
 
 fun Submission.allFiles() = section.allFiles() + section.allSections().flatMap { it.allFiles() }
 fun Submission.getSectionByType(name: String) = section.allSections().first { it.type == name }
-fun Submission.libFileSections() = (section.allSections() + section).filterNot { it.libraryFile == null }
-fun Submission.allReferencedFiles() =
-    section.allReferencedFiles() + section.allSections().flatMap { it.allReferencedFiles() }
+fun Submission.libFileSections() = (section.allSections() + section).filterNot { it.libraryFile.isNullOrBlank() }
+
+fun ExtendedSubmission.allExtendedSections() = extendedSection.allExtendedSections() + extendedSection
```
Juan-EBI: Same here I think we can separate extension by base type

commons/commons-bio/src/main/kotlin/ebi/ac/uk/model/ExtendedSection.kt
```diff
@@ -0,0 +1,42 @@
...
+        sections = section.sections
+        attributes = section.attributes
+
+        section.libraryFile?.let { libraryFile = LibraryFile(it) }
+        section.sections.forEach { sect ->
+            sect.fold({ extendedSections.addLeft(ExtendedSection(it)) }, { addSectionTable(it) })
+        }
+    }
+
+    fun addReferencedFile(file: File) = libraryFile?.addFile(file)
```
Juan-EBI: this one looks more like an extension

commons/commons-bio/src/main/kotlin/ebi/ac/uk/model/extensions/SectionExt.kt
jhoanmanuelms: Done

commons/commons-bio/src/main/kotlin/ebi/ac/uk/model/extensions/SubExt.kt
jhoanmanuelms: Done

PR #40, opened 2019-03-14
 Library File Formats Support

jhoanmanuelms:
- Add support for library file in all formats
- Use the extended section instead of the regular section for persistence
- Remove hardcoded library file in TSV format
- Fix library file name in the submission generated files


Juan-EBI:
LGTM

PR #41, opened 2019-03-15
Persist Referenced Files

jhoanmanuelms:
- Changes to persist files referenced in library files
- Add parameter to the submission mapper to indicate whether or not to load the referenced files from the database
- Verify referenced files in the integration tests

### DIFF
--- a/submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/service/SubmissionRepository.kt
+++ b/submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/service/SubmissionRepository.kt
@@ -5,9 +5,8 @@ import ac.uk.ebi.biostd.persistence.repositories.SubmissionDataRepository
 
 class SubmissionRepository(
     private val submissionRepository: SubmissionDataRepository,
-    private val submissionDbMapper: SubmissionDbMapper = SubmissionDbMapper()
+    private val submissionDbMapper: SubmissionDbMapper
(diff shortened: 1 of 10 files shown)


submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/service/SubmissionRepository.kt
```diff
@@ -11,8 +11,8 @@ class SubmissionRepository(
     fun getByAccNo(accNo: String) =
         submissionDbMapper.toSubmission(submissionRepository.getByAccNoAndVersionGreaterThan(accNo))
 
-    fun getExtendedByAccNo(accNo: String) =
-        submissionDbMapper.toExtSubmission(submissionRepository.getByAccNoAndVersionGreaterThan(accNo))
+    fun getExtendedByAccNo(accNo: String, loadRefFiles: Boolean = false) =
+        submissionDbMapper.toExtSubmission(submissionRepository.getByAccNoAndVersionGreaterThan(accNo), loadRefFiles)
```
Juan-EBI: I think we need to change the current approach of loading files in mapper by using hibernate lazy loading as is not clear in code (as is implicit) that the system is actually performing an extra query to load data.  My suggestion is to made this part of the code explicit as latter we will also add some custom logic here (to be hable to handle huge files) so I think we should propagate to mapper a map of library files that we load here in the repository.

PR #42, opened 2019-03-19
Tags Support

jhoanmanuelms:
- Add support for tags
- Update database schema


jhoanmanuelms:
The classifier shouldn't be a separated entity but rather a property of the tag

jhoanmanuelms:
Do you mean that instead of being a `MutableList<String>`, tags would be a `MutableList<Pair>` ?

PR #44, opened 2019-03-23
JSON Files Serialization

jhoanmanuelms:
Fix files serialization in JSON format


Juan-EBI:
Only one comment,  IMHO I think the implementation needed is something like a read-only attribute that means that if get a submission with a file type or file size we override we override with calculated values/

Juan-EBI:
Only one comment,  IMHO I think the implementation needed is something like a read-only attribute that means that if get a submission with a file type or file size we override we override with calculated values.

Said that in `deserialize` method we just ignore and in serialize we calculate and set the values.

PR #53, opened 2019-05-10
Fix TSV Deserialization Issues

jhoanmanuelms:
- Allow attributes with no value in TSV tables
- Fix parent section accNo in TSV serialization
- Improve error messages for invalid elements
- Fix TSV chunk line numbers in error messages
- Improve error messages for invalid host requests


commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/tsv/deserialization/TsvSerializationContext.kt
```diff
@@ -48,13 +49,17 @@ class TsvSerializationContext {
         execute(rootSection, chunk) { rootSection.addSectionTable(chunk.asTable()) }
 
     fun addSubSectionTable(parent: String, chunk: SectionTableChunk) =
-        execute(rootSection, chunk) { sections.getValue(parent).addSectionTable(chunk.asTable()) }
+        execute(rootSection, chunk) {
+            sections.getOrElse(parent, { throw InvalidParentSectionException(parent) }).addSectionTable(chunk.asTable())
```
Juan-EBI: I think we should keep this part clean and instead create specialized Map that handles this scenario something like 

```
SectionsContext {
    getSection(parentCccNo:String)
}
```

Juan-EBI: I think we should keep this part clean and instead create specialized Map that handles this scenario something like 

```
SectionsContext {
    getSection(parentCccNo:String)
}
```

client/bio-commandline/src/main/kotlin/uk/ac/ebi/biostd/client/cli/BioStudiesCommandLine.kt
```diff
@@ -36,13 +37,14 @@ class BioStudiesCommandLine : CliktCommand(name = "PTSubmit") {
...
-            val submission = client.submitSingle(input!!.readText(), SubmissionFormat.valueOf(format), files).body!!
+            val submission =
+                client.submitSingle(input!!.readText(), SubmissionFormat.valueOf(format.toUpperCase()), files).body!!
 
             echo("SUCCESS: Submission with AccNo ${submission.accNo} was submitted")
         } catch (exception: Exception) {
             when (exception) {
                 is ResourceAccessException -> throw PrintMessage(exception.cause?.message ?: FILES_NOT_FOUND_ERROR_MSG)
-                is RestClientResponseException -> throw PrintMessage(toJson(exception))
+                is RestClientResponseException -> throw PrintMessage(formatException(exception))
```
Juan-EBI: I think we should start doing this at the client level, catch any error and return proper JSON from there. Yet another TODO

Juan-EBI:
LGTM

PR #58, opened 2019-05-17
Mixed Files Resource

jhoanmanuelms:
- Create a new files resource that combines path and list files in order to allow the users to make submissions that include both, attached and already existing files
- Add a configuration property to specify the files directory
- Add and adjust unit tests

### DIFF
--- a/submission/submission-webapp/src/main/resources/application-local.yml
+++ b/submission/submission-webapp/src/main/resources/application-local.yml
@@ -18,7 +18,8 @@ spring:
       max-request-size: 10240MB
 
 app:
-  basepath: # Set here the path to the folder you want to use as storage
+  basepath: # Path to the folder to be used as storage
+  filesDirPath: # Path to the folder to be used for the user/groups files
   security:
     tokenHash: bio-local
     environment: LOCAL
--- a/submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/handlers/FilesHandler.kt
+++ b/submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/handlers/FilesHandler.kt
@@ -1,10 +1,10 @@
 package ac.uk.ebi.biostd.submission.handlers
 
 import ac.uk.ebi.biostd.SubFormat
-import ac.uk.ebi.biostd.submission.model.ListFilesSource
+import ac.uk.ebi.biostd.submission.model.AttachedFilesSource
 import ac.uk.ebi.biostd.submission.model.MixedFilesSource
-import ac.uk.ebi.biostd.submission.model.PathFilesSource
 import ac.uk.ebi.biostd.submission.model.ResourceFile
+import ac.uk.ebi.biostd.submission.model.UserFilesSource
 import ebi.ac.uk.model.ExtendedSubmission
 import ebi.ac.uk.paths.FolderResolver
 
@@ -23,8 +23,7 @@ class FilesHandler(
      */
     fun processFiles(submission: ExtendedSubmission, files: List<ResourceFile>, format: SubFormat) {
         val userFolder = folderResolver.getUserMagicFolderPath(submission.user.id, submission.user.secretKey)
-        val userFiles = PathFilesSource(userFolder)
-        val fileSource = if (files.isEmpty()) userFiles else MixedFilesSource(ListFilesSource(files), userFiles)
+        val fileSource = MixedFilesSource(AttachedFilesSource(files), UserFilesSource(userFolder))
 
         filesValidator.validate(submission, fileSource)
         libraryFilesHandler.processLibraryFiles(submission, fileSource, format)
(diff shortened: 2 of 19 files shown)


submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/model/FilesSource.kt
```diff
@@ -15,8 +14,26 @@ interface FilesSource {
...
 
-class ListFilesSource(private val files: List<ResourceFile>) : FilesSource {
+class MixedFilesSource(
+    private val listFiles: ListFilesSource,
+    private val pathFiles: PathFilesSource
+) : FilesSource {
+    override fun exists(filePath: String) = listFiles.exists(filePath).or(pathFiles.exists(filePath))
+
+    override fun getInputStream(filePath: String) =
+        executeBySource(filePath, { listFiles.getInputStream(filePath) }, { pathFiles.getInputStream(filePath) })
```
Juan-EBI: Consider the following aproach

`= exists(filePath).fold({ listFiles.getInputStream(filePath) }, {pathFiles.getInputStream(filePath) } }})`

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/model/FilesSource.kt
```diff
@@ -15,8 +14,26 @@ interface FilesSource {
     fun readText(filePath: String): String
 }
 
-class ListFilesSource(private val files: List<ResourceFile>) : FilesSource {
+class MixedFilesSource(
+    private val listFiles: ListFilesSource,
```
Juan-EBI: `ListFilesSource` and `PathFilesSource` names were not the best maybe refactor to FileSystemSource and ProvidedFileSource

submission/submission-webapp/src/main/resources/application-local.yml
```diff
@@ -18,7 +18,8 @@ spring:
       max-request-size: 10240MB
 
 app:
-  basepath: # Set here the path to the folder you want to use as storage
+  basepath: # Path to the folder to be used as storage
```
Juan-EBI: it may worth it to specify if path should be absolute or relative

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/handlers/FilesHandler.kt
```diff
@@ -22,7 +23,8 @@ class FilesHandler(
      */
     fun processFiles(submission: ExtendedSubmission, files: List<ResourceFile>, format: SubFormat) {
         val userFolder = folderResolver.getUserMagicFolderPath(submission.user.id, submission.user.secretKey)
-        val fileSource = if (files.isEmpty()) PathFilesSource(userFolder) else ListFilesSource(files)
+        val userFiles = PathFilesSource(userFolder)
+        val fileSource = if (files.isEmpty()) userFiles else MixedFilesSource(ListFilesSource(files), userFiles)
```
Juan-EBI: why not use always `MixedFilesSource`?

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/model/FilesSource.kt
```diff
@@ -15,8 +14,26 @@ interface FilesSource {
     fun readText(filePath: String): String
 }
 
-class ListFilesSource(private val files: List<ResourceFile>) : FilesSource {
+class MixedFilesSource(
```
Juan-EBI: this one should be the generic approach so it looks ok for be to be named `FilesSourc`

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/model/FilesSource.kt
jhoanmanuelms: I was actually thinking about renaming these to `AttachedFilesSource` and `UserFilesSource`. WDYT?

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/handlers/FilesHandler.kt
jhoanmanuelms: Sounds about right

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/model/FilesSource.kt
jhoanmanuelms: Probably not, because in the attachment process, it's fine to have just `ListFileSource` one. I think it's fine that we have the general `FilesSource` interface.

Maybe this name can be improved? `FallbackFilesSource` or something like that?

Juan-EBI: `it's fine to have just ListFileSource one`  I understand that is fine but why have different behavior based on request type,  in this case it does not represent any extra effort so I think is better to have single approach as is simpler.

PR #62, opened 2019-05-24
Library File Fixes

jhoanmanuelms:
- Fix file size
- Update library file attribute name

### DIFF
--- a/submission/submitter/src/test/kotlin/ac/uk/ebi/biostd/submission/model/FilesSourceTest.kt
+++ b/submission/submitter/src/test/kotlin/ac/uk/ebi/biostd/submission/model/FilesSourceTest.kt
@@ -48,7 +48,7 @@ class FilesSourceTest(temporaryFolder: TemporaryFolder) {
 
     @Test
     fun `get user file size`() {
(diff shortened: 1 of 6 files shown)


submission/submitter/src/test/kotlin/ac/uk/ebi/biostd/submission/model/FilesSourceTest.kt
```diff
@@ -48,7 +48,7 @@ class FilesSourceTest(temporaryFolder: TemporaryFolder) {
 
     @Test
     fun `get user file size`() {
-        assertThat(testInstance.size(TEST_USER_FILE)).isGreaterThan(0)
+        assertThat(testInstance.size(TEST_USER_FILE)).isEqualTo(0)
```
Juan-EBI: I am wondering if there is an easy way to create a file of a specific size to test this properly

jhoanmanuelms: Nice catch. Done.

PR #72, opened 2019-06-13
Fix Access Tag

jhoanmanuelms:
- Private submissions shouldn't inherit the "Public" access tag if its parent if public
- Fix library file serialization
- Fix typos in configuration files

### DIFF
--- a/submission/submission-webapp/src/main/resources/application-local.yml
+++ b/submission/submission-webapp/src/main/resources/application-local.yml
@@ -18,10 +18,10 @@ spring:
       max-request-size: 10240MB
 
 app:
-  basepath: # Path to the folder to be used as storage
-  filesDirPath: # Path to the folder to be used for the user/groups files
-  tmpDirPath: # Absolute of the temp folder to store files
+  basepath: # Absolute path to the folder to be used as storage
+  tempDirPath: # Absolute of the temp folder to store files
   security:
     tokenHash: bio-local
+    filesDirPath: # Absolute path to the folder to be used for the user/groups files
     environment: LOCAL
     requireActivation: false
--- a/submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/processors/AccessTagProcessor.kt
+++ b/submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/processors/AccessTagProcessor.kt
@@ -9,7 +9,11 @@ import ebi.ac.uk.persistence.PersistenceContext
  */
 class AccessTagProcessor : SubmissionProcessor {
     override fun process(submission: ExtendedSubmission, context: PersistenceContext) {
-        submission.accessTags.addAll(context.getParentAccessTags(submission))
-        submission.accessTags.removeIf { it == "Public" && submission.releaseDate.isNullOrBlank().not() }
+        val accessTags =
+            context.getParentAccessTags(submission)
+                .toMutableList()
+                .apply { removeIf { it == "Public" && submission.releaseDate.isNullOrBlank().not() } }
+
+        submission.accessTags.addAll(accessTags)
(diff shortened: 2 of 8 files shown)


submission/submission-webapp/src/main/resources/application-local.yml
```diff
@@ -18,10 +18,10 @@ spring:
       max-request-size: 10240MB
 
 app:
-  basepath: # Path to the folder to be used as storage
-  filesDirPath: # Path to the folder to be used for the user/groups files
-  tmpDirPath: # Absolute of the temp folder to store files
+  basepath: # Absolute path to the folder to be used as storage
```
Juan-EBI: clearly not part of the scope of this ticket but this weekend I learn that we can document property class and IDEa will provide nice autocompletion for properties so if you have time we can do it there.

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/processors/AccessTagProcessor.kt
```diff
@@ -1,14 +1,15 @@
...
 import ebi.ac.uk.persistence.PersistenceContext
 
 /**
  * Add all parent submission tags to submitted submission.
  */
 class AccessTagProcessor : SubmissionProcessor {
-
     override fun process(submission: ExtendedSubmission, context: PersistenceContext) {
         submission.accessTags.addAll(context.getParentAccessTags(submission))
+        submission.accessTags.removeIf { it == "Public" && submission.releaseDate.isNullOrBlank().not() }
```
Juan-EBI: I think instead we should filter the tags that we add from parent and add a public tag if released, final results is exactly the same but in this way is more consistent with the real business cases scenarios.

jhoanmanuelms: Done

submission/submission-webapp/src/main/resources/application-local.yml
jhoanmanuelms: I think this file is more intended to run the application than to be used in the IDE. This is a good suggestion tho.

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/processors/AccessTagProcessor.kt
```diff
@@ -1,14 +1,19 @@
...
 class AccessTagProcessor : SubmissionProcessor {
-
     override fun process(submission: ExtendedSubmission, context: PersistenceContext) {
-        submission.accessTags.addAll(context.getParentAccessTags(submission))
+        val accessTags =
+            context.getParentAccessTags(submission)
+                .toMutableList()
+                .apply { removeIf { it == "Public" && submission.releaseDate.isNullOrBlank().not() } }
+
+        submission.accessTags.addAll(accessTags)
```
Juan-EBI: my suggestion was intended to something like 

```
submission.accessTags.addAll(accessTags).filter { it == "Public"}
```

and then in proper class
```
if (releaseDate > today){
     submission.addAccessTag("Public")
}
```

jhoanmanuelms: Yes, I understand this, but maybe this is a mixing of concerns. Maybe this change must be made after the processors/handlers architecture is changed.

Otherwise, we'll be setting access tags in the time's processor or checking times in the access tags processor.

jhoanmanuelms: Done

PR #73, opened 2019-06-14
TSV Comments

### LINKED ISSUE #65: Support Comments TSV
Page Tab specification supports comments for TSV submissions. This needs to be implemented in the new submitter.

According to page tab specification, any line starting with "#" should be ignored.

jhoanmanuelms:
Support for comments in TSV page tab format

Fixes #65


Juan-EBI:
Changes looks good, I just wondering what is the purpose of this feature if a comment gets deleted and lost after submit.

jhoanmanuelms:
I think this is more intended to pass information to the users, for example, in the BioImages template, comments are used to let the users know what information to place on each field.

PR #83, opened 2019-07-09
File List Streaming Deserializers

jhoanmanuelms:
Create JSON, XML and TSV deserializers that stream the file list instead of directly using string content


commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/xml/deserializer/stream/FileListXmlStreamDeserializer.kt
```diff
@@ -0,0 +1,35 @@
...
+        val reader = XMLInputFactory.newFactory().createXMLStreamReader(file.inputStream())
+        val module = JacksonXmlModule().apply {
+            setDefaultUseWrapper(false)
+            addDeserializer(PageTabFile::class.java, FileXmlStreamDeserializer())
+            addDeserializer(Attribute::class.java, AttributeXmlStreamDeserializer())
+        }
+        val mapper = XmlMapper(module)
+        reader.next()
+
+        while (reader.hasNext()) {
```
jcamiloradamesa: here and in others part of the code this "old" style iterations make the code looks pretty complicate and obfuscate logic. is true that the library provides and horrible api but we can extend it to provide a nice method. For example, here I can imagine.

`Reader.forEach(element: (ReaderType) -> Unit)`

commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/xml/deserializer/stream/FileXmlStreamDeserializer.kt
```diff
@@ -0,0 +1,35 @@
...
+import ebi.ac.uk.model.File
+import ebi.ac.uk.model.constants.AttributeFields
+import ebi.ac.uk.model.constants.FileFields
+
+internal class FileXmlStreamDeserializer : StdDeserializer<File>(File::class.java) {
+    override fun deserialize(p: JsonParser?, ctxt: DeserializationContext?): File {
+        var path = ""
+        val attributes: MutableList<Attribute> = mutableListOf()
+
+        while (p!!.nextToken() != JsonToken.END_OBJECT) {
```
jcamiloradamesa: use platform type and create iteration exxtension

commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/xml/deserializer/stream/AttributeXmlStreamDeserializer.kt
```diff
@@ -0,0 +1,32 @@
...
+
+            p.nextToken()
+
+            when (field) {
+                NAME -> name = p.text.trim()
+                VALUE -> value = p.text.trim()
+            }
+        }
+
+        value.ifBlank { throw InvalidFileListElementException("Attribute value is mandatory") }
```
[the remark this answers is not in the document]
jhoanmanuelms: I added a TODO for this

commons/commons-serialization/src/test/kotlin/ac/uk/ebi/biostd/json/deserialization/FileListJsonStreamDeserializerTest.kt
```diff
@@ -0,0 +1,62 @@
...
+const val JSON_FILE_LIST = "FileList.json"
+
+@ExtendWith(TemporaryFolderExtension::class)
+class FileListJsonStreamDeserializerTest(temporaryFolder: TemporaryFolder) {
+    private val testInstance = FileListJsonStreamDeserializer()
+    private val jsonFile = temporaryFolder.createFile(JSON_FILE_LIST)
+
+    @BeforeEach
+    fun beforeEach() {
+        val jsonFileList = jsonArray(jsonObj {
```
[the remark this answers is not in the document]
jhoanmanuelms: This is planned for a future PR as discussed before

commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/xml/deserializer/stream/FileListXmlStreamDeserializer.kt
```diff
@@ -0,0 +1,35 @@
...
+import ebi.ac.uk.model.File as PageTabFile
+import ebi.ac.uk.model.FileList
+import java.io.File
+import javax.xml.stream.XMLInputFactory
+
+internal class FileListXmlStreamDeserializer {
+    fun deserialize(file: File): FileList {
+        val referencedFiles: MutableList<PageTabFile> = mutableListOf()
+        val reader = XMLInputFactory.newFactory().createXMLStreamReader(file.inputStream())
+        val module = JacksonXmlModule().apply {
```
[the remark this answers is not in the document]
jhoanmanuelms: This is created only once when the file list starts to stream. After that, the mapper is reused to map each file.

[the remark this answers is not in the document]
Juan-EBI: e.g. we are creating a `JacksonXmlModule` each time we are serialization a file so if a submission with 10 files will be created 10 of times which is not necessary at all.

commons/commons-serialization/src/test/kotlin/ac/uk/ebi/biostd/json/deserialization/FileListJsonStreamDeserializerTest.kt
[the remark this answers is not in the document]
Juan-EBI: mm got it, lest review this because the only way I see to test API is actually working is by coding this test before integrating.

commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/ext/XMLStreamReaderExt.kt
```diff
@@ -0,0 +1,18 @@
+package ac.uk.ebi.biostd.ext
+
+import javax.xml.stream.XMLStreamReader
+
+/**
+ * Iterates over each XML node loaded by the reader and applies the given function .
+ *
+ * @param function The function to be applied to each element.
+ */
+fun XMLStreamReader.forEach(function: () -> Unit) {
```
Juan-EBI: 2 issues in this function.

1. Is not safe (user need to remember to call next and close) otherwise will work incorrectly.
2. Iteration should be over reader object so user does not need to access externally variable.

commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/ext/XMLStreamReaderExt.kt
```diff
@@ -0,0 +1,18 @@
...
+    while (hasNext()) {
+        try {
+            function()
+        } catch (exception: NoSuchElementException) {
+            break
```
Juan-EBI: we should use exceptional code "normal code", we need to identify when there is not more reader and finalize buckle is like making this.

```
var list = listOf(10, 22)
index = 0;
while(true) {
     try {
     print(list[index++])
    }
    catch (exception: NoSuchElementException)
        break;
} 
```

commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/ext/JsonParserExt.kt
```diff
@@ -0,0 +1,75 @@
+package ac.uk.ebi.biostd.ext
+
+import ac.uk.ebi.biostd.common.deserialization.stream.StreamDeserializerBuilder
+import com.fasterxml.jackson.core.JsonParseException
+import com.fasterxml.jackson.core.JsonParser
+import com.fasterxml.jackson.core.JsonToken
+
+/**
```
Juan-EBI: I would expect a single method with the call to `nexToken` which is the iterator one.

commons/commons-bio/src/main/kotlin/ebi/ac/uk/model/builders/FileBuilder.kt
```diff
@@ -0,0 +1,14 @@
+package ebi.ac.uk.model.builders
+
+import ebi.ac.uk.model.Attribute
+import ebi.ac.uk.model.File
+
+class FileBuilder {
+    var path: String? = null
+    var size: Long = 0
```
Juan-EBI: we should not populate size with the value provided by the user, this need to be calculated based on file size

jhoanmanuelms: Done

PR #84, opened 2019-07-10
Persisten mapper

Juan-EBI:
- added persistence to extended mapping support


commons/commons-util/src/main/kotlin/ebi/ac/uk/util/collections/MapExtensions.kt
```diff
@@ -0,0 +1,11 @@
+package ebi.ac.uk.util.collections
+
+fun <V, K> Map<Boolean, List<V>>.fold(whenTrue: (List<V>) -> K, whenFalse: (List<V>) -> K): List<K> {
```
jhoanmanuelms: It would be nice to add docs to the extensión functions

submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/mapping/extended/to/ToExtAttribute.kt
```diff
@@ -0,0 +1,13 @@
+package ac.uk.ebi.biostd.persistence.mapping.extended.to
+
+import ac.uk.ebi.biostd.persistence.model.Attribute
+import ac.uk.ebi.biostd.persistence.model.AttributeDetail
+import ebi.ac.uk.extended.model.ExtAttribute
+import ebi.ac.uk.extended.model.ExtAttributeDetail
+
+internal const val TO_EXT_ATTRIBUTE_EXTENSIONS = "ac.uk.ebi.biostd.persistence.mapping.extended.to.ToExtAttributeKt"
+
+internal fun Attribute.toExtAttribute(): ExtAttribute = ExtAttribute(
```
jhoanmanuelms: As a standard, I've noticed we've been using the extended class as the file name. For example, AttributeExt.kt

Juan-EBI: problem is that we have a different extension for the same class . Extension to convert to one way and extension to convert to the other side so mix them in single file looks pretty complicated.

PR #87, opened 2019-07-24
pivotal-167036604: Login - Logout is not refreshing token

Juan-EBI:


submission/submission-security/src/main/kotlin/ebi/ac/uk/security/model/TokenPayload.kt
```diff
@@ -1,3 +1,14 @@
 package ebi.ac.uk.security.model
 
-internal data class TokenPayload(val id: Long, val email: String, val fullName: String)
+import java.time.OffsetDateTime
+
+internal data class TokenPayload(
+    val id: Long,
+    val email:
+    String,
```
jhoanmanuelms: Type should be next to variable

submission/submission-security/src/test/kotlin/ebi/ac/uk/security/service/SecurityServiceTest.kt
```diff
@@ -259,4 +263,28 @@ internal class SecurityServiceTest(
...
     }
+
+    @Nested
+    inner class CheckToken {
+
+        private val securityToken = mockk<SecurityToken>()
+        private val user = mockk<User>()
+
+        @Test
+        fun `check when is not in black list`() {
```
jhoanmanuelms: We should add a test to verify every time a token is created it has a creation time which is the variation

PR #92, opened 2019-08-06
 PMC Import Past Submissions Support

jhoanmanuelms:
- Support to reload PMC missed submissions
- Add notifications URL to the properties files


jhoanmanuelms:
Changes not required yet

PR #94, opened 2019-08-12
Feature - Pivotal ID: 167762806 - Read Excel File

jhoanmanuelms:
- Implement ExcelReader in order to read page tab from excel files
- Use the corresponding reader in the command line depending on the type


Juan-EBI:
The implementation looks good but we need to add this at submission level by adding `submitXlsx` method in `SubmitResource`

PR #101, opened 2019-09-05
Fix: Pivotal ID # 168286132 - TSV Parent Acc No

jhoanmanuelms:
- Add parent accession number when adding a subsection
- Add parent accession to the TSV serialization
- Remove unnecessary "loadRefFiles" flag
- Improve unit tests

### DIFF
--- a/commons/commons-serialization/src/test/kotlin/ac/uk/ebi/biostd/tsv/TsvDeserializerTest.kt
+++ b/commons/commons-serialization/src/test/kotlin/ac/uk/ebi/biostd/tsv/TsvDeserializerTest.kt
@@ -9,6 +9,7 @@ import ac.uk.ebi.biostd.test.submissionWithFilesTable
 import ac.uk.ebi.biostd.test.submissionWithInnerSubsections
 import ac.uk.ebi.biostd.test.submissionWithInnerSubsectionsTable
 import ac.uk.ebi.biostd.test.submissionWithInvalidAttribute
+import ac.uk.ebi.biostd.test.submissionWithInvalidInnerSubsection
 import ac.uk.ebi.biostd.test.submissionWithInvalidNameAttributeDetail
 import ac.uk.ebi.biostd.test.submissionWithInvalidValueAttributeDetail
 import ac.uk.ebi.biostd.test.submissionWithLinks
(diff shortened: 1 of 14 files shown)


submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/mapping/SubmissionDbMapper.kt
```diff
@@ -45,7 +45,7 @@ import ac.uk.ebi.biostd.persistence.model.User as UserDb
 class SubmissionDbMapper {
     private val sectionMapper = DbSectionMapper()
 
-    fun toExtSubmission(submissionDb: SubmissionDb, loadRefFiles: Boolean = false) =
```
Juan-EBI: :)

submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/mapping/SubmissionDbMapper.kt
```diff
@@ -76,27 +76,25 @@ class SubmissionDbMapper {
 }
 
 private class DbSectionMapper {
-    internal fun toSection(sectionDb: SectionDb): Section =
+    internal fun toSection(sectionDb: SectionDb, parentAccNo: String? = null): Section =
```
Juan-EBI: I believe in the `SectionDb` and `ExtendedSection` parent section should be a reference to a section, not a plain string. AccNo is only a way for the user to indicate parent section but system internal should honor `parentId` .

commons/commons-serialization/src/test/kotlin/ac/uk/ebi/biostd/tsv/TsvDeserializerTest.kt
```diff
@@ -153,7 +153,10 @@ class TsvDeserializerTest {
                     Attribute("Grant Id", "No. 2015BAD27B01")),
                 sections = mutableListOf(
```
Juan-EBI: we may add a test and handle case (or todo with ticket) when parent does exist.

submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/mapping/SubmissionDbMapper.kt
jhoanmanuelms: You're right about this. I removed this part since it's actually not necessary for this ticket. 

However, the `parentAccNo` is currently defined as a string in both `Section` and `ExtendedSeciton` models so, we may consider creating a separated chore task to refactor this.

commons/commons-serialization/src/test/kotlin/ac/uk/ebi/biostd/tsv/TsvDeserializerTest.kt
jhoanmanuelms: This case is already handled by the TSV serializer. I added a unit test for this case.

PR #119, opened 2019-09-26
Pivotal ID # 168777655: RootPath Support

Juan-EBI:
https://www.pivotaltracker.com/story/show/168777655
- added rootpath source support
- refactor sources

Delivers #168777655


commons/commons-bio/src/main/kotlin/ebi/ac/uk/io/sources/ComposeFileSource.kt
```diff
@@ -0,0 +1,12 @@
+package ebi.ac.uk.io.sources
+
+class ComposeFileSource(private vararg val sources: FilesSource) : FilesSource {
+
```
jhoanmanuelms: Blank

commons/commons-bio/src/main/kotlin/ebi/ac/uk/io/sources/ComposeFileSource.kt
```diff
@@ -0,0 +1,12 @@
+package ebi.ac.uk.io.sources
+
+class ComposeFileSource(private vararg val sources: FilesSource) : FilesSource {
```
jhoanmanuelms: Composed

PR #120, opened 2019-09-27
Pivotal ID # 168727325: Project Submission Relationship

jhoanmanuelms:
- Validate that the project has the corresponding access tag
- Add access tag for the test project
- Add an endpoint and a web client method to create projects
- The corresponding access tag for the project will be automatically created
- Use the new methods in the projects API test

TODO In Next PR
- Some classes require unit testing
- Validate only superusers are able to submit projects
- The integration tests that use projects should validate tags


submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/integration/PersistenceContextImpl.kt
```diff
@@ -60,6 +62,12 @@ open class PersistenceContextImpl(
         }
     }
 
+    override fun saveAccessTag(accessTag: String) {
+        accessTagsDataRepository.save(AccessTag(name = accessTag))
```
Juan-EBI: inliner

submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/service/SubmissionRepository.kt
```diff
@@ -34,6 +34,7 @@ class SubmissionRepository(
         submissionRepository.findByTypeAndAccNo(Project.value, tags.map { it.name })
             .map { submissionDbMapper.toSubmission(it) }
 
+    // TODO this method should return the submission model, not the persistence object
```
Juan-EBI: do not get this TODO, this one is a spring data repository

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/processors/ProjectProcessor.kt
```diff
@@ -0,0 +1,11 @@
+package ac.uk.ebi.biostd.submission.processors
+
+import ebi.ac.uk.model.ExtendedSubmission
+import ebi.ac.uk.model.extensions.addAccessTag
+import ebi.ac.uk.persistence.PersistenceContext
+
+class ProjectProcessor : SubmissionProcessor {
```
Juan-EBI: Is this only applied to the project? as it implements SubmissionProcessor I would expect is applied to every submission and it does look correct them

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/validators/ProjectValidator.kt
```diff
@@ -1,12 +1,19 @@
...
+import ac.uk.ebi.biostd.submission.exceptions.ProjectAlreadyExistingException
+import ebi.ac.uk.base.ifTrue
 import ebi.ac.uk.model.ExtendedSubmission
-import ebi.ac.uk.model.extensions.attachTo
 import ebi.ac.uk.persistence.PersistenceContext
 
 class ProjectValidator : SubmissionValidator {
     override fun validate(submission: ExtendedSubmission, context: PersistenceContext) {
-        submission.attachTo?.let { context.getSubmission(it) ?: throw InvalidProjectException(it) }
+        context.isNew(submission.accNo).not().ifTrue {
```
Juan-EBI: we may need better `ifFalse` extension

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/validators/SubmissionProjectValidator.kt
```diff
@@ -0,0 +1,24 @@
...
+        submission.attachTo?.let {
+            validateProject(it, context)
+            validateProjectAccessTag(it, submission, context)
+        }
+    }
+
+    private fun validateProject(project: String, context: PersistenceContext) =
+        context.getSubmission(project) ?: throw InvalidProjectException(project)
+
+    private fun validateProjectAccessTag(project: String, submission: ExtendedSubmission, context: PersistenceContext) =
```
Juan-EBI: what is the purpose of this validation? the idea AFAIK is that the system will create access tag for the project. Users should not use access tags.

submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/integration/PersistenceContextImpl.kt
jhoanmanuelms: This is a unit return type

submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/service/SubmissionRepository.kt
jhoanmanuelms: In all the other methods, we return the submission model object but in this case, we return the persistence object. This method should also use the Submission Mapper and return a Submission model object

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/processors/ProjectProcessor.kt
jhoanmanuelms: Well, the project is a submission. In this case, this is a processor for the projects which follows the same validation/processing approach.

Maybe we could rename the interface as `SubmitterProcessor` and `SubmitterValidator`?

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/validators/ProjectValidator.kt
jhoanmanuelms: Done

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/validators/SubmissionProjectValidator.kt
jhoanmanuelms: This validation is intended for projects that were manually created and the access tag might not exist.

submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/service/SubmissionRepository.kt
Juan-EBI: ohh got it I was thinking this was `SubmissionDataRepository` and not `SubmissionRepository`

PR #125, opened 2019-10-07
Pivotal ID # 168727325: Submission API

jhoanmanuelms:
- Return model instead of persistence object in submissions repository
- Change filters to OffsetDate
- Remove unnecessary headers
- Add version in response
- Update the tests and add an integration test to the filtering by release date


submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/web/resources/SubmissionResource.kt
```diff
@@ -44,5 +41,8 @@ class SubmissionResource(
         @AuthenticationPrincipal user: SecurityUser
     ): List<SubmissionDto> = submissionWebHandler
         .getSubmissions(user, filter)
-        .map { SubmissionDto(it.accNo, it.title, it.creationTime, it.modificationTime, it.releaseTime) }
+        .map {
+            SubmissionDto(
+                it.accNo, it.title ?: "", it.version, it.creationTime, it.modificationTime, it.releaseTime)
```
Juan-EBI: I think there is one extension `orEmpty

jhoanmanuelms: Done

PR #126, opened 2019-10-07
Pivotal ID # 168982339: Classify Integration Tests

jhoanmanuelms:
- Create separated packages to classify the integration tests depending on the functionality being tested
- Organize tests in the new packages
- Rename some classes in order to better represent what they're actually testing


jhoanmanuelms:
Good suggestions. Done.

PR #128, opened 2019-10-10
Pivotal ID #169062746: Show errors when submission validation fails

awaisathar:
Added validation exception reporting


submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/SubmissionSubmitter.kt
```diff
@@ -17,10 +22,39 @@ class SubmissionSubmitter(
...
         files: FilesSource,
         context: PersistenceContext
+
     ): Submission {
-        validators.forEach { validator -> validator.validate(submission, context) }
-        processors.forEach { processor -> processor.process(submission, context) }
-        filesHandler.processFiles(submission, files)
+        val validationNodes = mutableListOf<ValidationNode>()
+
+        validators.map {
```
Juan-EBI: I think we can do this code more declarative(or functional) and keep logic clean (right now is obfuscated by error handling) which is not the purpose of this class, what we can do here is. Collect all exceptions in a `Try` and accumulate them, then we check if not empty in order to trigger the error or not.  Check https://arrow-kt.io/docs/arrow/core/try/

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/SubmissionSubmitter.kt
```diff
@@ -17,10 +22,39 @@ class SubmissionSubmitter(
         submission: ExtendedSubmission,
         files: FilesSource,
         context: PersistenceContext
+
     ): Submission {
-        validators.forEach { validator -> validator.validate(submission, context) }
-        processors.forEach { processor -> processor.process(submission, context) }
-        filesHandler.processFiles(submission, files)
+        val validationNodes = mutableListOf<ValidationNode>()
```
jhoanmanuelms: I would suggest using a context object to store the errors if they exist, something like this approach:

https://github.com/EBIBioStudies/biostudies-backend-services/blob/master/commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/tsv/deserialization/chunks/ChunkProcessor.kt#L51

And pass that object to each validator/processor.

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/SubmissionSubmitter.kt
```diff
@@ -17,10 +22,39 @@ class SubmissionSubmitter(
...
+            try {
+                it.process(submission, context)
+            } catch (e: ValidationException) {
+                ValidationNode(ValidationNodeStatus.ERROR, e.message ?: e.javaClass.name)
+            }
+        }.filterIsInstance<ValidationNode>().ifNotEmpty {
+            validationNodes.add(ValidationNode(ValidationNodeStatus.ERROR, "Processing errors", it))
+        }
+
+        try {
```
jhoanmanuelms: Then, verifying whether or not this context object contains errors and handle them properly.

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/SubmissionSubmitter.kt
awaisathar: "Try"ing to minimize code change :D

PR #129, opened 2019-10-11
Pivotal ID #168982339: Integration Tests Refactor

jhoanmanuelms:
- Refactor common functionality into the base integration test class
- Remove duplicated tests
- Use TSV as standard for all the submission tests
- Use submission objects to validate instead of comparing values
- Relocate the submission draft integration test into the corresponding package


commons/commons-bio/src/main/kotlin/ebi/ac/uk/model/Submission.kt
```diff
@@ -9,12 +9,12 @@ open class Submission(
     var accessTags: MutableList<String> = mutableListOf(),
     override var attributes: List<Attribute> = listOf()
 ) : Attributable {
-
```
Juan-EBI: nice

commons/commons-bio/src/main/kotlin/ebi/ac/uk/model/TableModel.kt
```diff
@@ -119,4 +118,7 @@ class SectionsTable(sections: List<Section> = emptyList()) : Table<Section>(sect
         override val id = t.accNo!!
         override val attributes = t.attributes
     }
+
+    fun asSectionsTable() = SectionsTable(
```
Juan-EBI: this shows the problem with wrongly used inheritance and the issues in our design, we do not know what instances have this collection and we need to check it manually.

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/assertions/AllInOneSubmissionHelper.kt
```diff
@@ -0,0 +1,49 @@
...
+import ebi.ac.uk.model.ExtendedSubmission
+import io.github.glytching.junit.extension.folder.TemporaryFolder
+import org.assertj.core.api.Assertions.assertThat
+import java.nio.file.Paths
+
+internal class AllInOneSubmissionHelper(
+    private val basePath: String,
+    private val submissionRepository: SubmissionRepository
+) {
+    internal fun createAllInOneSubmissionFiles(webClient: BioWebClient, tempFolder: TemporaryFolder) {
```
Juan-EBI: What I do not like about this method is that it only make sense if all submission structure match this one, so basically there is a relation between this and object that is not explicit in code which forces others check this and all submission code in order to make sense of it. In my opinion, methods should do a single thing but do it properly. So if this one is in charge it should create the full submission, it that way it will complete a full consistent task.

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/assertions/AllInOneSubmissionHelper.kt
```diff
@@ -0,0 +1,49 @@
...
+        tempFolder.createDirectory("Folder1/Folder2")
+
+        webClient.uploadFiles(listOf(tempFolder.createFile("DataFile1.txt"), tempFolder.createFile("DataFile2.txt")))
+        webClient.uploadFiles(listOf(tempFolder.createFile("Folder1/DataFile3.txt")), "Folder1")
+        webClient.uploadFiles(listOf(tempFolder.createFile("Folder1/Folder2/DataFile4.txt")), "Folder1/Folder2")
+    }
+
+    internal fun assertSavedSubmission(accNo: String) {
+        val extendedSubmission = submissionRepository.getExtendedByAccNo(accNo)
+        assertThat(extendedSubmission.asSubmission()).isEqualTo(allInOneSubmission(accNo))
```
Juan-EBI: I was cour equals and hascode implementation and I think we need to change them, is ok to use them to compare object basic properties as title, attributes, url but not for nested objects as this is making hibernate load the full entity three (making lazy loading basically useless) . That refactor is out of the scope of this issue but if we trust equals latter we will need to re introduce assertions

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/common/BaseIntegrationTest.kt
```diff
@@ -16,4 +25,30 @@ internal open class BaseIntegrationTest(private val tempFolder: TemporaryFolder)
         System.setProperty("app.tempDirPath", temp.absolutePath)
         System.setProperty("app.security.filesDirPath", dropbox.absolutePath)
     }
+
+    protected fun getWebClient(serverPort: Int, user: TestUser): BioWebClient {
```
Juan-EBI: IMO this method make 100% sense for this generic base class but if you check others as `submitString` they does not bellow to all test, they do help to reuse code but there is no a reason to all integration test have that method available.

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/common/BaseIntegrationTest.kt
```diff
@@ -16,4 +25,30 @@ internal open class BaseIntegrationTest(private val tempFolder: TemporaryFolder)
...
+    }
+
+    protected fun submitString(
+        webClient: BioWebClient,
+        submission: String,
+        format: SubmissionFormat = TSV,
+        files: List<File> = emptyList()
+    ) = assertSuccessfulResponse(webClient.submitSingle(submission, format, files))
+
+    protected fun submitFile(webClient: BioWebClient, submission: File, files: List<File> = emptyList()) =
```
Juan-EBI: this method does help to reuse code but actually make code harder to understand and maintain as user will need to go to this method definition to check how submission is submitted.  `webClient.submitSingle(submission, files)` is IMO way more clear even is duplicated code, as commented before IMO our final goal is no make maintainable code not to avoid duplication (if we can do both much better).

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/entities/RegularUser.kt
```diff
@@ -5,10 +5,10 @@ import ebi.ac.uk.api.security.RegisterRequest
 /**
  * Represents a bio studies regular user.
  */
-object RegularUser {
-    const val username = "Regular User"
-    const val email = "biostudies-dev@ebi.ac.uk"
-    const val password = "678910"
+object RegularUser : TestUser {
```
Juan-EBI: nice

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/files/UserFileApiTest.kt
```diff
@@ -36,10 +35,7 @@ internal class UserFileApiTest(private val tempFolder: TemporaryFolder) : BaseIn
 
         @BeforeAll
         fun init() {
-            val securityClient = SecurityWebClient.create("http://localhost:$serverPort")
-            securityClient.registerUser(SuperUser.asRegisterRequest())
-
-            webClient = securityClient.getAuthenticatedClient(SuperUser.email, SuperUser.password)
+            webClient = getWebClient(serverPort, SuperUser)
```
Juan-EBI: nice

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/MultipartFileSubmissionApiTest.kt
```diff
@@ -176,7 +139,7 @@ internal class MultipartFileSubmissionApiTest(
                         "value" to "FileList.json"
                     })
                 }
-            }
+            }.toString()
 
             val fileList = tempFolder.createFile("FileList.json").apply {
                 writeBytes(jsonArray({
```
Juan-EBI: I create an overload that accept file and content so I think we can use everywhere it looks way more clean IMO.

```
 val fileList = tempFolder.createFile("FileList.tsv").apply {
                writeBytes(tsv {
                    line("Files", "GEN")
                    line("SomeFile.txt", "ABC")
                }.toString().toByteArray())
            }
```

```
val fileList = tempFolder.createFile(
    "FileList.tsv", 
     tsv {
            line("Files", "GEN")
            line("SomeFile.txt", "ABC")
        }.toString())
```

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/common/BaseIntegrationTest.kt
jhoanmanuelms: Done

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/common/BaseIntegrationTest.kt
jhoanmanuelms: Done

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/MultipartFileSubmissionApiTest.kt
jhoanmanuelms: Done

PR #132, opened 2019-10-18
Pivotal ID # 169226913: Excel File Streaming

jhoanmanuelms:
Use streaming library to read excel files in order to handle big files


commons/commons-util/src/main/kotlin/ebi/ac/uk/util/file/ExcelReader.kt
```diff
@@ -1,13 +1,21 @@
...
-import org.apache.poi.xssf.usermodel.XSSFWorkbook
 import java.io.File
 
+const val BUFFER_SIZE = 4096
+const val ROW_CACHE_SIZE = 1000
+
 class ExcelReader {
     fun readContentAsTsv(file: File): String {
-        val sheet = XSSFWorkbook(file.inputStream()).getSheetAt(0)
+        val sheet = StreamingReader.builder()
```
Juan-EBI: don't we need to close the buffered reader?

jhoanmanuelms: Good catch. Done.

commons/commons-util/src/main/kotlin/ebi/ac/uk/util/file/ExcelReader.kt
```diff
@@ -1,13 +1,24 @@
...
     fun readContentAsTsv(file: File): String {
-        val sheet = XSSFWorkbook(file.inputStream()).getSheetAt(0)
+        val inputStream = file.inputStream()
+        val sheet = StreamingReader.builder()
+            .rowCacheSize(ROW_CACHE_SIZE)
+            .bufferSize(BUFFER_SIZE)
+            .open(inputStream)
+            .getSheetAt(0)
+
+        inputStream.close()
```
Juan-EBI: take a look of kotlin special handling for [autoclosable](https://www.baeldung.com/kotlin-try-with-resources)

jhoanmanuelms: Right. Done.

PR #134, opened 2019-10-18
Pivotal ID # 168982339: Empty AccNo

jhoanmanuelms:
- Fix the problem with the submissions without an accession number
- Create an integration test to validate the scenario

### DIFF
--- a/submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/EmptyAccNoApiTest.kt
+++ b/submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/EmptyAccNoApiTest.kt
@@ -1,0 +1,68 @@
+package ac.uk.ebi.biostd.itest.test.submission.submit
+
+import ac.uk.ebi.biostd.client.integration.commons.SubmissionFormat
+import ac.uk.ebi.biostd.client.integration.web.BioWebClient
+import ac.uk.ebi.biostd.common.config.PersistenceConfig
+import ac.uk.ebi.biostd.itest.common.BaseIntegrationTest
+import ac.uk.ebi.biostd.itest.entities.SuperUser
+import ac.uk.ebi.biostd.persistence.model.Counter
+import ac.uk.ebi.biostd.persistence.model.Sequence
+import ac.uk.ebi.biostd.persistence.repositories.SequenceDataRepository
+import ac.uk.ebi.biostd.persistence.service.SubmissionRepository
+import ebi.ac.uk.asserts.assertThat
+import ebi.ac.uk.dsl.line
+import ebi.ac.uk.dsl.submission
+import ebi.ac.uk.dsl.tsv
+import ebi.ac.uk.model.extensions.title
+import io.github.glytching.junit.extension.folder.TemporaryFolder
+import io.github.glytching.junit.extension.folder.TemporaryFolderExtension
+import org.assertj.core.api.Assertions.assertThat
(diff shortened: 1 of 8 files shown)


submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/EmptyAccNoApiTest.kt
```diff
@@ -0,0 +1,68 @@
...
+    ) {
+        @LocalServerPort
+        private var serverPort: Int = 0
+
+        private lateinit var webClient: BioWebClient
+
+        @BeforeAll
+        fun init() {
+            webClient = getWebClient(serverPort, SuperUser)
+            sequenceRepository.save(Sequence("S-BSST", "").apply { counter = Counter("S-BSST") })
```
Juan-EBI: it looks handy to create a sequence constructor that creates the counter too.

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/EmptyAccNoApiTest.kt
```diff
@@ -0,0 +1,68 @@
...
+import org.springframework.test.context.junit.jupiter.SpringExtension
+
+@ExtendWith(TemporaryFolderExtension::class)
+internal class EmptyAccNoApiTest(tempFolder: TemporaryFolder) : BaseIntegrationTest(tempFolder) {
+    @Nested
+    @Import(PersistenceConfig::class)
+    @ExtendWith(SpringExtension::class)
+    @SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+    @DirtiesContext
+    inner class EmptyAccNoApiTest(
```
Juan-EBI: a full integration test only for this single scenario looks like too much. I think we can put this one in `SubmissionApiTest` with other submission specific cases as `rootpath`  and `release date`

jhoanmanuelms: I put it here because we'll have more of these in the future, for submission with parent accession with a pattern or negative cases for example

Juan-EBI: so we are having more test of this type name test `EmptyAccNoApiTest.kt` does not look correct for me, also all this test that you mention looks more suitable in `SubmissionApiTest`

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/EmptyAccNoApiTest.kt
jhoanmanuelms: Done

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/EmptyAccNoApiTest.kt
jhoanmanuelms: No, I mean more test cases. Right now there's only `empty accNo` but I'd include `empty accNo with parent`, `empty accNo without pattern`, `empty accNo with parent and no pattern` and so on.

PR #135, opened 2019-10-22
Pivotal ID #169289621: Null-Not present submission support

Juan-EBI:
https://www.pivotaltracker.com/story/show/169289621
- added null handling

### DIFF
--- a/commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/ext/JsonNode.kt
+++ b/commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/ext/JsonNode.kt
@@ -3,20 +3,20 @@ package ac.uk.ebi.biostd.ext
 import com.fasterxml.jackson.databind.JsonNode
 import com.fasterxml.jackson.databind.node.NullNode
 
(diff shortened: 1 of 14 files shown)


commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/ext/JsonNode.kt
```diff
@@ -1,20 +1,29 @@
...
 import com.fasterxml.jackson.databind.JsonNode
+import com.fasterxml.jackson.databind.node.NullNode
 
-internal inline fun <reified T : JsonNode?> JsonNode.findNode(property: String) =
-    get(property)?.also {
-        require(it is T) {
-            "Expecting node: '$this', property: '$property' to be of type '${T::class.java.simpleName}' " +
-                "but '${it::class.java.simpleName}' find instead"
-        }
+internal inline fun <reified T : JsonNode> JsonNode.findNode(property: String): T? {
```
jhoanmanuelms: 👍

jhoanmanuelms: You could use directly `=` instead of return

PR #136, opened 2019-10-23
Pivotal ID # 168982339: Project Pattern Sequence Generation

jhoanmanuelms:
- Change the projects endpoint to make the accNo template mandatory and create the corresponding sequence
- Create integration tests for project accNo template
- Move the empty accNo integration tests to submission test class

TODO In Next PR
- Simplify the sequence object by removing the suffix which is unused
- Include project sequences in the test database setup


submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/integration/PersistenceContextImpl.kt
```diff
@@ -28,6 +29,10 @@ open class PersistenceContextImpl(
     private val subDbMapper: SubmissionDbMapper,
     private val subMapper: SubmissionMapper
 ) : PersistenceContext {
+    override fun createAccNoPatternSequence(pattern: AccPattern) {
+        sequenceRepository.save(Sequence(pattern.prefix))
```
Juan-EBI: I believe we need to use both prefix and suffix here

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/project/submit/ProjectSubmitTest.kt
```diff
@@ -67,6 +70,7 @@ internal class ProjectSubmitTest(private val tempFolder: TemporaryFolder) : Base
             assertThat(submittedProject.accessTags.first()).isEqualTo("AProject")
 
             assertThat(tagsDataRepository.existsByName("AProject")).isTrue()
+            assertThat(sequenceRepository.existsByPrefixAndSuffix("S-APR", "")).isTrue()
```
Juan-EBI: dont we need to validate both?

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/SubmissionApiTest.kt
```diff
@@ -73,6 +81,48 @@ internal class SubmissionApiTest(private val tempFolder: TemporaryFolder) : Base
...
+            )
+        }
+
+        @Test
+        fun `empty accNo with parent`() {
+            val project = tsv {
+                line("Submission", "A-Project")
+                line("AccNoTemplate", "!{S-APR,}")
+                line()
+
```
Juan-EBI: any particular reason for this line break

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/ProjectSubmitter.kt
```diff
@@ -1,18 +1,25 @@
...
+    private val accNoPatternUtil: AccNoPatternUtil,
     private val validators: List<IProjectValidator>,
     private val processors: List<IProjectProcessor>
 ) {
     fun submit(project: ExtendedSubmission, context: PersistenceContext): Submission {
         validators.forEach { validator -> validator.validate(project, context) }
         processors.forEach { processor -> processor.process(project, context) }
+
+        val sequencePrefix = accNoPatternUtil.getPattern(project.accNoTemplate!!).prefix
+        context.createAccNoPatternSequence(AccPattern(sequencePrefix))
```
Juan-EBI: should not we use both?

submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/integration/PersistenceContextImpl.kt
jhoanmanuelms: This won't make sense after I remove the suffix as discussed before.

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/ProjectSubmitter.kt
jhoanmanuelms: Same

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/SubmissionApiTest.kt
Juan-EBI: I mean the single `line()` already represents a line break so not need to add another line break. Do I am missing something?

jhoanmanuelms: That's just to make the pagetab more readable. We're doing the same in all the other cases and I think it looks good and makes clear at a single glance, where the line breaks are:

https://github.com/EBIBioStudies/biostudies-backend-services/blob/master/submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/factory/TsvSubmissionFactory.kt#L13

Juan-EBI: it looks for me as a double line but is a matter of test. Completely subjective.

jhoanmanuelms: Ok, maybe we should change these and all the others in a separated PR, just to be consistent.

PR #137, opened 2019-10-24
Pivotal ID # 169315679: Delete draft when submission is submitted

awaisathar:


submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/integration/PersistenceContextImpl.kt
```diff
@@ -59,6 +62,10 @@ open class PersistenceContextImpl(
             subRepository.expireActiveVersions(submission.accNo)
             submission.version = nextVersion
             subRepository.save(subMapper.toSubmissionDb(submission))
+            // TODO: Refactor draft deletion by making DraftService listen to a successful submission event
+            userDataRepository.findByUserIdAndKeyIgnoreCaseContaining(submission.user.id, submission.accNo).ifNotEmpty {
```
jhoanmanuelms: I don't think we should place this directly here. There's already a method for that in `SubmissionDraftService`. The proper way of doing this should be placing it as a new step in the submission process in `SubmissionSubmitter`

submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/repositories/Repositories.kt
```diff
@@ -79,8 +79,7 @@ interface AccessPermissionRepository : JpaRepository<AccessPermission, Long> {
 }
 
 interface UserDataDataRepository : JpaRepository<UserData, UserDataId> {
-    @Query("FROM UserData ud WHERE ud.userId = :userId AND LOWER(ud.key) LIKE %:dataKey%")
-    fun findByUserAndKey(userId: Long, dataKey: String): List<UserData>
+    fun findByUserIdAndKeyIgnoreCaseContaining(userId: Long, dataKey: String): List<UserData>
```
jhoanmanuelms: Are we sure about ignoring case here?

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/SubmissionDraftApiTest.kt
```diff
@@ -76,5 +76,13 @@ internal class SubmissionDraftApiTest(tempFolder: TemporaryFolder) : BaseIntegra
             webClient.deleteSubmissionDraft(draftSubmission.key)
             assertThat(webClient.getAllSubmissionDrafts()).isEmpty()
         }
+
+        @Test
+        fun `delete submission draft after submission`() {
+            webClient.submitSingle(pageTab, SubmissionFormat.JSON)
+            val draftSubmission = webClient.getSubmissionDraft("ABC-123")
+            webClient.submitSingle(draftSubmission.content.toString(), SubmissionFormat.JSON)
```
jhoanmanuelms: I think it would be nice to modify this test to perform the same submission defined in the `pageTab` variable but in TSV format, which would also cover the case where a separated "direct submission" that should delete any existing drafts.

submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/integration/PersistenceContextImpl.kt
awaisathar: We want to carry out this particular operation in the same transaction to avoid multiple submissions by accident (due to lag or slow response of backend/proxy). As pointed out by @Juan-EBI , we can't guarantee consistency with distributed transactions so in this particular case, I think we should fix this here temporarily until we can discuss a better approach in detail. This will be a minor addition to the technical debt but a major issue would be closed.

submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/repositories/Repositories.kt
awaisathar: I wasn't sure about it too but the current code was relying on the service to convert the key to lowercase and this is a native Spring Data method.  In general, accessions are case insensitive so I don't think we should make them case-sensitive for drafts.

submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/integration/PersistenceContextImpl.kt
Juan-EBI: but having this here is a hidden effect. Basically a function that is call `getSequenceNextValue` also delete drafts. What I suggest we do here is to solve this,  is moving `@Transactional` to `SubmissionSubmitter` and follow @jhoanmanuelms advice.

awaisathar: I'm sold at the business logic point. We can make `SubmissionSubmitter.submit()` transactional but `SubmissionDraftService` is in the `submission-webapp` module. How about moving the logic to `SubmissionService.submit()` instead and making that `@Transactional?`

awaisathar: Scratch that. It creates a circular dependency. Let me try to refactor the logic in `PersistenceContext` and see if that makes more sense.

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/SubmissionSubmitter.kt
```diff
@@ -11,13 +11,16 @@ import ebi.ac.uk.model.ExtendedSubmission
 import ebi.ac.uk.model.Submission
 import ebi.ac.uk.persistence.PersistenceContext
 import ebi.ac.uk.util.collections.ifNotEmpty
+import org.springframework.transaction.annotation.Isolation
+import org.springframework.transaction.annotation.Transactional
 
-class SubmissionSubmitter(
+open class SubmissionSubmitter(
```
Juan-EBI: why this class is open now?

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/SubmissionSubmitter.kt
```diff
@@ -11,13 +11,16 @@ import ebi.ac.uk.model.ExtendedSubmission
...
     private val validators: List<SubmissionValidator>,
     private val processors: List<SubmissionProcessor>,
     private val filesHandler: FilesHandler
 ) {
-    fun submit(
+    @Transactional(isolation = Isolation.READ_UNCOMMITTED)
+    open fun submit(
```
Juan-EBI: why this one is open now?

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/SubmissionSubmitter.kt
awaisathar: "The default advice mode for processing `@Transactional` annotations is `proxy`" which requires overriding

https://docs.spring.io/spring/docs/5.2.x/spring-framework-reference/data-access.html#transaction-declarative-annotations

Juan-EBI: yes :) but we use `allopen` kotlin compiler plugin that will do that for us. So adding plugin in subproject will do the trick with not coding changes. (check persistence project for reference )

awaisathar: allopen not behaving as documented.

PR #142, opened 2019-10-30
pivotal-169315694: Release date retrieve as ISO (bug)

Juan-EBI:
https://www.pivotaltracker.com/story/show/169315694
- added iso date handling


commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/json/serialization/SubmissionJsonSerializer.kt
```diff
@@ -21,11 +20,7 @@ internal class SubmissionJsonSerializer : StdSerializer<Submission>(Submission::
             writeJsonObject(SubFields.SECTION, subm.section)
 
             if (subm is ExtendedSubmission) {
-                writeJsonArray(SubFields.ACCESS_TAGS, merge(subm.accessTags), gen::writeString)
```
jhoanmanuelms: So, we no longer need these other fields?

commons/commons-serialization/src/test/kotlin/ac/uk/ebi/biostd/json/serialization/SubmissionJsonSerializerTest.kt
```diff
@@ -0,0 +1,68 @@
...
+    @Test
+    fun `serialize extended submission`(@MockK user: User) {
+        val dateTime = OffsetDateTime.of(2018, 12, 31, 0, 0, 0, 0, ZoneOffset.UTC)
+        val extendedSubmission = ExtendedSubmission(submission, user).apply { releaseTime = dateTime }
+
+        val json = testInstance.writeValueAsString(extendedSubmission)
+        val expected = jsonObj {
+            "accno" to "abc123"
+            "section" to jsonObj { }
+            "rtime" to "2018-12-31Z"
```
jhoanmanuelms: Is the ISO date format expected to have that `Z` at the end?

commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/json/serialization/SubmissionJsonSerializer.kt
Juan-EBI: they were initially added to support exporter calling this system but better simplify until we got to that point and have a proper design.

PR #143, opened 2019-10-31
Pivotal ID # 169314354: Improve Error Handling

jhoanmanuelms:
- Add proper exception for the case when the project doesn't exist and no acc no is not provided
- Fix exception handling in the BioStudies CLI


Juan-EBI:
I think we should not include http logic and spring classes usage in the terminal client, this all are concerns of the web client and as we do this is we will not be able to change web client. (for example for using another http client). In my opinion, we should implement a ` ResponseErrorHandler` at web client level wrapping exception into `ClientException` at catch that one in the terminal to have clear separations of concerns and dependencies.

jhoanmanuelms:
Very good observation. I made the required changes.

PR #145, opened 2019-11-07
Pivotal ID # 169314431: Project Resource Docs

jhoanmanuelms:
- Use swagger annotations to describe the API
- Configure the swagger to show the documentation


submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/common/config/SecurityConfig.kt
```diff
@@ -44,6 +44,10 @@ class SecurityConfig(
             .and()
             .authorizeRequests()
             .antMatchers("/auth/**").permitAll()
+            .antMatchers("/v2/**").permitAll()
```
Juan-EBI: what are this resources starting with `v2` ?

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/web/resources/ProjectResource.kt
```diff
@@ -18,21 +22,33 @@ import org.springframework.web.bind.annotation.RequestParam
...
 
     @PostMapping(headers = ["${HttpHeaders.CONTENT_TYPE}=$MULTIPART_FORM_DATA"])
     @ResponseBody
+    @ApiOperation("Register a new project")
+    @ApiImplicitParam(name = "X-Session-Token", value = "The authentication token", required = true)
     fun submit(
-        @AuthenticationPrincipal user: SecurityUser,
+        @ApiIgnore
+        @AuthenticationPrincipal
+        user: SecurityUser,
```
Juan-EBI: I am wondering if ` .ignoredParameterTypes(AuthenticationPrincipal.class)` does not work?

jhoanmanuelms: You're right, nice catch!

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/common/config/SecurityConfig.kt
jhoanmanuelms: This is the path for the swagger resource generated by the annotations, which later is used by the swagger-ui plugin to show the documentation

PR #147, opened 2019-11-13
Pivotal ID # 169590703: Drafts Pagination

jhoanmanuelms:
- Changes to make the drafts resource pageable
- Remove unnecessary endpoints
- Rename the package containing filter related logic
- Improve variable and endpoint naming


submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/filter/PaginationFilter.kt
```diff
@@ -0,0 +1,6 @@
+package ac.uk.ebi.biostd.persistence.filter
+
+open class PaginationFilter(val limit: Int = 15, private val offset: Int = 0) {
```
Juan-EBI: I think we need to separate filter from paging for me both concepts are different and should not be coupled. `SubmissionDraftClient#getAllSubmissionDrafts` should receive in my opinion and offset and a limit as a parameter not a map that includes this two and user need to remember paging parameter names

client/bio-webclient/src/main/kotlin/ac/uk/ebi/biostd/client/api/SubmissionDraftClient.kt
```diff
@@ -5,22 +5,23 @@ import ebi.ac.uk.model.SubmissionDraft
...
 
-    override fun searchSubmissionDraft(searchText: String): List<SubmissionDraft> =
-        template.getForObject<Array<SubmissionDraft>>("$SUBMISSION_DRAFT_URL?searchText=$searchText").orEmpty().toList()
+    override fun getAllSubmissionDrafts(filter: Map<String, Any>): List<SubmissionDraft> {
+        val builder = UriComponentsBuilder.fromUriString(SUBMISSION_DRAFT_URL)
+        filter.entries.forEach { builder.queryParam(it.key, it.value) }
 
-    override fun getAllSubmissionDrafts(): List<SubmissionDraft> =
-        template.getForObject<Array<SubmissionDraft>>(SUBMISSION_DRAFT_URL).orEmpty().toList()
+        return template.getForObject<Array<SubmissionDraft>>(builder.toUriString()).orEmpty().toList()
```
Juan-EBI: what about functional version

```
    private fun asUrl(filter: Map<String, Any>): String = filter.entries
        .fold(fromUriString(SUBMISSION_DRAFT_URL), { builder, entry -> builder.queryParam(entry.key, entry.value) })
        .toUriString()
```

jhoanmanuelms: Done

submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/filter/PaginationFilter.kt
jhoanmanuelms: Done

PR #154, opened 2019-12-05
Pivotal ID #169314393: Submission Processing Status

jhoanmanuelms:
- Add the processing status to the submission persistence and business models
- Set the proper status once the submission has been processed


jcamiloradamesa:
we can not merge this one until old backend is 100% off.

submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/mapping/SubmissionDbMapper.kt
```diff
@@ -55,6 +56,7 @@ class SubmissionDbMapper {
             modificationTime = toInstant(submissionDb.releaseTime)
             releaseTime = toInstant(submissionDb.releaseTime)
             section = sectionMapper.toSection(submissionDb.rootSection)
+            processingStatus = ProcessingStatus.valueOf(submissionDb.status)
```
jcamiloradamesa: database records saved and retrieve by backend is not like a web API input that user can change so we can assume that value is a proper one. Not need the validation of unexpected value IMO.

PR #155, opened 2019-12-06
Pivotal ID # 169314393: Submission Processing Status

jhoanmanuelms:
- Add the processing status to the submission persistence and business models
- Set the proper status once the submission has been processed


commons/commons-bio/src/main/kotlin/ebi/ac/uk/model/constants/ProcessingStatus.kt
```diff
@@ -0,0 +1,19 @@
...
+
+private const val PROCESSED_STATUS = "PROCESSED"
+private const val PROCESSING_STATUS = "PROCESSING"
+
+sealed class ProcessingStatus(val value: String) {
+    companion object {
+        fun valueOf(value: String): ProcessingStatus = when (value) {
+            PROCESSED_STATUS -> Processed
+            PROCESSING_STATUS -> Processing
+            else -> throw UnsupportedStatusException(value)
```
Juan-EBI: status comes from DB I think we should not handle invalid status as the system itself  is the only one that store value there is not that there is an api where user can set anything.

submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/model/Submission.kt
```diff
@@ -91,6 +91,9 @@ class Submission(
     @Column
     var secretKey: String = ""
 
+    @Column
+    var status: String = ""
```
Juan-EBI: I think we can use here a `@Converter` and an enum type to avoid setting the wrong value.

commons/commons-bio/src/main/kotlin/ebi/ac/uk/model/constants/ProcessingStatus.kt
jhoanmanuelms: Done

submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/model/Submission.kt
jhoanmanuelms: Done

PR #164, opened 2020-01-27
Pivotal ID #170872493: Duplicated Projects

jhoanmanuelms:
Fix projects query in order to retrieve only the active ones


submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/repositories/Repositories.kt
```diff
@@ -45,7 +45,12 @@ interface SubmissionDataRepository :
 
     fun existsByAccNo(accNo: String): Boolean
 
-    fun findByRootSectionTypeAndAccNoIn(type: String, tags: List<String>, graph: GraphSpecification): List<Submission>
+    fun findByRootSectionTypeAndAccNoInAndVersionGreaterThan(
+        type: String,
+        tags: List<String>,
+        graph: GraphSpecification,
+        long: Int = 0
```
Juan-EBI: should be version

PR #166, opened 2020-01-28
Pivotal #ID 169314671: Direct upload submission flag

Juan-EBI:
https://www.pivotaltracker.com/story/show/169314671
- added submission source support


commons/commons-bio/src/main/kotlin/ebi/ac/uk/model/SubmissionSource.kt
```diff
@@ -0,0 +1,5 @@
+package ebi.ac.uk.model
+
+enum class SubmissionSource {
```
jhoanmanuelms: I would suggest: `SubmissionMethod` and for the methods, I would suggest `FILE`, `PAGE_TAB`

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/ProjectSubmitter.kt
```diff
@@ -17,18 +17,21 @@ open class ProjectSubmitter(
...
-        context.createAccNoPatternSequence(sequencePrefix)
-        context.saveAccessTag(project.accNo)
-
-        project.processingStatus = PROCESSED
-        context.saveSubmission(project)
+        throw InvalidSubmissionException("Submission validation errors", processingErrors)
+    }
 
-        return project
+    private fun process(submission: ExtendedSubmission, context: PersistenceContext): List<Throwable> {
```
jhoanmanuelms: Couldn't this be a one-liner?

PR #168, opened 2020-01-30
Pivotal ID # 170975897: Attach To Endpoint

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/170975897

Create an endpoint to attach submissions to a project instead of using the direct submission endpoint with an additional parameter.

### DIFF
--- a/client/bio-webclient/src/main/kotlin/ac/uk/ebi/biostd/client/api/ProjectClient.kt
+++ b/client/bio-webclient/src/main/kotlin/ac/uk/ebi/biostd/client/api/ProjectClient.kt
@@ -5,7 +5,9 @@ import ac.uk.ebi.biostd.client.integration.web.ProjectOperations
 import ac.uk.ebi.biostd.integration.SerializationService
 import ac.uk.ebi.biostd.integration.SubFormat
 import ebi.ac.uk.model.Submission
+import ebi.ac.uk.model.constants.FILES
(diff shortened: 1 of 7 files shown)


client/bio-webclient/src/main/kotlin/ac/uk/ebi/biostd/client/api/ProjectClient.kt
```diff
@@ -22,14 +24,30 @@ class ProjectClient(
...
+    ): ResponseEntity<Submission> {
+        val headers = HttpHeaders().apply { MediaType.MULTIPART_FORM_DATA }
+        val body = getMultipartBody(files, FileSystemResource(submission))
+
+        return template
+            .postForEntity<String>("/projects/$projectAccNo/submissions", HttpEntity(body, headers))
+            .map { serializationService.deserializeSubmission(it, SubFormat.JSON_PRETTY) }
+    }
+
+    private fun getMultipartBody(files: List<File>, submission: Any) = LinkedMultiValueMap<String, Any>().apply {
```
Juan-EBI: IMO always favor construction over mutation. 

`files.map { FILES to FileSystemResource(it) }.plus { SUBMISSION to FileSystemResource(submission) }.toMap()`

In this way you can use not a mutable object map. In general. In general we should never use mutable data structures,  is the exception because it use reflection so is mandatory.

PR #171, opened 2020-02-04
Pivotal ID # 170993540: Refactor Processors To Services

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/170993540

- Refactor project submitter and processors to use services
- Pass persistence context as a property instead of as a parameter
- Improve variable and classes naming
- Remove unnecessary code

### DIFF
--- a/submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/ProjectSubmitter.kt
+++ b/submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/ProjectSubmitter.kt
@@ -18,28 +18,24 @@ open class ProjectSubmitter(
     private val projectValidationService: ProjectValidationService
 ) {
     @Transactional(isolation = Isolation.READ_UNCOMMITTED)
-    open fun submit(project: ExtendedSubmission): Submission {
-        val processingErrors = process(project)
-        if (processingErrors.isEmpty()) {
-            val sequencePrefix = accNoPatternUtil.getPattern(project.accNoTemplate!!)
-            persistenceContext.createAccNoPatternSequence(sequencePrefix)
-            persistenceContext.saveAccessTag(project.accNo)
-            project.processingStatus = PROCESSED
-            return persistenceContext.saveSubmission(project)
-        }
+    open fun submit(project: ExtendedSubmission): Submission =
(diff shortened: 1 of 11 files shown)


submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/ProjectSubmitter.kt
```diff
@@ -1,36 +1,45 @@
...
         }
 
         throw InvalidSubmissionException("Submission validation errors", processingErrors)
     }
 
-    private fun process(submission: ExtendedSubmission, context: PersistenceContext) =
-        processors
-            .map { runCatching { it.process(submission, context) } }
-            .mapNotNull { it.exceptionOrNull() }
+    private fun process(project: ExtendedSubmission) =
```
Juan-EBI: no need to wrap into a list as single processing result. So simple instance and null check will do the job

jhoanmanuelms: `IvalidSubmissionException` expects a list of errors in case it fails. I think it's cleaner to leave it like this than having to declare a list later. Also, both submitters would be consistent.

Juan-EBI: do not agree, if the method retrieves a single error, the single error should be retrieved. We should not change it to make looks similar to the submission business case.  I consider consistent for example all variables starting with lower case, or use final keyword for parameter arguments, but not this case. 

About exception, I see from a business perspective that is totally possible that either multiple or single error is thrown, like users having a a single or multiple errors so InvalidSubmissionException should support single error too. 

Said that I totally understand your point, like from code perspective is cleaner retrieve a list and also is simpler to have similar solution for both submission and projects, for me is more important code to be similar to business (if single error is throw single element) but it hard to argument which quality attribute is better. 

In conclusion 1000 lines consider the argument and merge changes (adding more changes or not).There is not clear better solution IMO.

jhoanmanuelms: image
😆

PR #180, opened 2020-03-06
Fix Projects AccNo

jhoanmanuelms:
Fix the calculation of the AccNo for projects


jhoanmanuelms:
This is better fixed in #181

PR #182, opened 2020-03-11
Pivotal ID # 171647484: TSV Serialization Inconsistency

jhoanmanuelms:
**Pivotal Link:** https://www.pivotaltracker.com/story/show/171647484

- Set parentAccNo property during deserialization
- Validate duplicated section AccNo
- Fix parent accNo serialization
- Add validation for TSV sections table with parent

Fixes #171647484

### DIFF
--- a/commons/commons-bio/src/main/kotlin/ebi/ac/uk/model/TableModel.kt
+++ b/commons/commons-bio/src/main/kotlin/ebi/ac/uk/model/TableModel.kt
@@ -100,7 +100,7 @@ class FilesTable(files: List<File> = emptyList()) : Table<File>(files) {
 
 class SectionsTable(sections: List<Section> = emptyList()) : Table<Section>(sections) {
     private var sectionType = ""
-    private var parentAccNo: String? = null
+    var parentAccNo: String? = null
 
     override val header: String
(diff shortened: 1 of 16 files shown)


commons/commons-bio/src/main/kotlin/ebi/ac/uk/model/TableModel.kt
```diff
@@ -100,7 +100,7 @@ class FilesTable(files: List<File> = emptyList()) : Table<File>(files) {
 
 class SectionsTable(sections: List<Section> = emptyList()) : Table<Section>(sections) {
     private var sectionType = ""
-    private var parentAccNo: String? = null
+    var parentAccNo: String? = null
```
Juan-EBI: Here we were breaking immutability (ideally class should not be changed after created ) and now we are also breaking encapsulation as the property is public. It will be really hard to track changes in this and also will make the class no threadsafe. 

I understand that there is a need to know the parent section when serialization page tab so I think the proper solution will get the parent section at serialization time at writing here, rather than trust in this property.

jhoanmanuelms:
Fixed by #183

PR #185, opened 2020-03-16
Pivotal ID # 170971934: Use study section title for submission list

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/170971934

Use the section title for the submission list in case the  submission title is empty

### DIFF
--- a/submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/model/ext/SectionExt.kt
+++ b/submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/model/ext/SectionExt.kt
@@ -4,3 +4,6 @@ import ac.uk.ebi.biostd.persistence.common.NO_TABLE_INDEX
 import ac.uk.ebi.biostd.persistence.model.Section
 
 internal fun Section.isTableElement() = tableIndex != NO_TABLE_INDEX
+
+internal val Section.title: String?
+    get() = attributes.find { it.name == "Title" }?.value
--- a/submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/model/ext/SectionExt.kt
+++ b/submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/model/ext/SectionExt.kt
@@ -2,8 +2,9 @@ package ac.uk.ebi.biostd.persistence.model.ext
 
 import ac.uk.ebi.biostd.persistence.common.NO_TABLE_INDEX
 import ac.uk.ebi.biostd.persistence.model.Section
+import ebi.ac.uk.model.constants.SectionFields
 
 internal fun Section.isTableElement() = tableIndex != NO_TABLE_INDEX
 
 internal val Section.title: String?
-    get() = attributes.find { it.name == "Title" }?.value
+    get() = attributes.find { it.name == SectionFields.TITLE.value }?.value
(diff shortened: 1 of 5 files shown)


submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/query/SubmissionListApiTest.kt
```diff
@@ -114,6 +116,57 @@ internal class SubmissionListApiTest(private val tempFolder: TemporaryFolder) :
             assertThat(submissionList).hasSize(5)
         }
 
+        @Test
+        fun `submission with section title`() {
```
Juan-EBI: I am not sure Rick about this one, I like it that we have a strong API test but also worry that they become too many and hard are to track. 

The problem is that right now,  is that we do not have that good enough unit test so this full end to end is what is saving as. (Not changes are expected in code for this comment)

submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/model/ext/SectionExt.kt
```diff
@@ -4,3 +4,6 @@ import ac.uk.ebi.biostd.persistence.common.NO_TABLE_INDEX
 import ac.uk.ebi.biostd.persistence.model.Section
 
 internal fun Section.isTableElement() = tableIndex != NO_TABLE_INDEX
+
+internal val Section.title: String?
+    get() = attributes.find { it.name == "Title" }?.value
```
Juan-EBI: I think we have a constant for this one, havent we?

submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/projections/SimpleSubmission.kt
```diff
@@ -26,9 +25,11 @@ data class SimpleSubmission(
 ) {
 
     companion object {
-
         const val SIMPLE_GRAPH: String = SIMPLE_QUERY_GRAPH
+
         fun SubmissionDb.asSimpleSubmission(): SimpleSubmission {
+            val title = if (title.isNullOrBlank()) rootSection.title else title
```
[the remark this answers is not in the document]
jhoanmanuelms: I think this option only would work if the title is null, not if it's blank

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/query/SubmissionListApiTest.kt
jhoanmanuelms: I'm not sure if there's actually something like too much testing, maybe it's worth reading about it. AFAIK, is ok to have a new integration test for each feature.

submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/model/ext/SectionExt.kt
jhoanmanuelms: Done

submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/projections/SimpleSubmission.kt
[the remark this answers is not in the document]
jhoanmanuelms: Done

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/query/SubmissionListApiTest.kt
Juan-EBI: not saying too much testing, maybe unbalanced testing, like testing at the wrong level. Like for example testing unit by integration test. But about "is ok to have a new integration test for each feature." it may be yes, just wondering best way to organize them and keep track of them

PR #186, opened 2020-03-17
Pivotal ID # 138327043: Email User Successful Submission

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/138327043

- Changes to notify the user once the submission has been processed
- Create the corresponding events and templates
- Add a configuration parameter to decide whether or not to send the submission


submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/notifications/NotificationsSubscriber.kt
```diff
@@ -22,29 +26,52 @@ internal class NotificationsSubscriber(
...
         val template = PasswordResetTemplate(getTemplateContent("reset-password.html"))
         val subscription = NotificationType(EMAIL_FROM, "BioStudies password reset", template)
-        subscriptionService.create(subscription, passwordReset.map { asNotification(it) })
+        subscriptionService.create(subscription, passwordReset.map { asPasswordResetNotification(it) })
     }
 
-    private fun asNotification(source: PasswordReset) =
+    @PostConstruct
+    fun successfulSubmission() {
+        if (notificationProperties.submissionNotification) {
```
Juan-EBI: I believe there is no scenario when we do not want email notification, we may want to configure allow to configure this at the user level but not  system level

submission/submission-webapp/src/main/resources/emails/successful-submission.html
```diff
@@ -0,0 +1,43 @@
...
+  https://www.ebi.ac.uk/biostudies/studies/${ACC_NO}</a> in the next 24 hours.
+</p>
+<div style="display: ${IS_PRIVATE};">
+  <p>
+    The study release date is not set so it's not publicly available. You will be able to see it only by logging in or
+    by accessing it through this link: <a href="https://www.ebi.ac.uk/biostudies/studies/${ACC_NO}/${SECRET_KEY}">
+    https://www.ebi.ac.uk/biostudies/studies/${ACC_NO}/${SECRET_KEY}</a>.
+  </p>
+</div>
+<div style="display: ${IS_PRIVATE_RELEASE_DATE};">
```
Juan-EBI: I think we should not use display property to remove this, as HTML still will be available in user email. We need to use the template engine conditional.

I notice that we do not have any template engine yet in the project, so to solve this we can use freemaker or any other popular template engine for Java (is not that we have super complex requirements here) and change current HTML to be templates of that type. See for example  [Seding Email with Freemarker HTML](https://memorynotfound.com/spring-mail-sending-email-freemarker-html-template-example/)

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/SubmissionSubmitter.kt
```diff
@@ -58,9 +60,13 @@ open class SubmissionSubmitter(
         method: SubmissionMethod
     ): ExtSubmission {
         try {
-            return processSubmission(submission, user, source, method)
-        } catch (e: RuntimeException) {
-            throw InvalidSubmissionException("Submission validation errors", listOf(e))
+            val submitted = processSubmission(submission, user, source, method)
+            successfulSubmission.onNext(SuccessfulSubmission(
```
Juan-EBI: here the submission has not been submitted yet. I think this should be after the persistence layer process getscompleted.

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/notifications/NotificationsSubscriber.kt
jhoanmanuelms: It doesn't make much sense to me to have this enabled in dev and beta since these are used mostly for testing purposes.

awaisathar: Just wanted to add that from experience, sometimes we do need to turn of notifications globally before reloading part of the database automatically (usually in bulk).

submission/notifications/src/main/kotlin/ebi/ac/uk/notifications/api/RtClient.kt
```diff
@@ -0,0 +1,43 @@
...
+import ebi.ac.uk.util.collections.second
+import ebi.ac.uk.util.regex.match
+import ebi.ac.uk.util.regex.secondGroup
+import org.springframework.util.LinkedMultiValueMap
+import org.springframework.web.client.RestTemplate
+import org.springframework.web.client.postForEntity
+import org.springframework.web.util.UriComponentsBuilder
+
+class RtClient(private val rtConfig: RtConfig) {
+    fun createTicket(content: String): String {
```
Juan-EBI: I was expecting to a user email, how the notification will be sent to the end-user?

commons/commons-bio/src/main/kotlin/ebi/ac/uk/model/User.kt
```diff
@@ -1,3 +1,3 @@
 package ebi.ac.uk.model
 
-data class User(val id: Long, val email: String, val secretKey: String)
+data class User(val id: Long, val email: String, val secretKey: String, val fullName: String? = null)
```
Juan-EBI: so full name is declared as nullable but then latter in `RtNotificationSubscriber` we call ` source.user.fullName!!`

submission/notifications/src/main/kotlin/ebi/ac/uk/notifications/api/RtClient.kt
Juan-EBI: The idea of creating a client is to encapsulate rtConfig by chequing [documentation](https://rt-wiki.bestpractical.com/wiki/REST#Ticket_Create)  I see that values are passed in string with line break separated, making a client receiving this string and sending the same string does not simplify RT. What I  would expect is client receiving a  RtNotification class which contains subject email and textContent and like that so internally we used it.

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/notifications/templates/SuccessfulSubmissionTemplate.kt
```diff
@@ -0,0 +1,58 @@
+package ac.uk.ebi.biostd.notifications.templates
+
+import ebi.ac.uk.base.isNotBlank
+import ebi.ac.uk.notifications.integration.model.NotificationTemplate
+import ebi.ac.uk.notifications.integration.model.NotificationTemplateModel
+
+internal class SuccessfulSubmissionTemplate : NotificationTemplate<SuccessfulSubmissionModel>("") {
+    override fun getContent(model: SuccessfulSubmissionModel): String = """
```
Juan-EBI: not really a fan of template content in code, why we just not use the same approach as plain resource files (as with HTML email), it is are simpler/easier to control and edit.

submission/notifications/src/main/kotlin/ebi/ac/uk/notifications/integration/model/NotificationTemplate.kt
```diff
@@ -7,7 +7,7 @@ open class NotificationTemplate<T : NotificationTemplateModel>(private var templ
     /**
      * Obtaining the string representation by replacing model properties into template string.
      */
-    fun getContent(model: T): String {
+    open fun getContent(model: T): String {
```
Juan-EBI: why we now required to override this?  it is the only function in this base class so for me overriding means that we do not want to sue this one.

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/SubmissionSubmitter.kt
```diff
@@ -47,7 +49,12 @@ open class SubmissionSubmitter(
     open fun submit(request: SubmissionRequest): Submission {
         val user = request.user.asUser()
         val extSubmission = process(request.submission, request.user.asUser(), request.files, request.method)
-        return context.saveSubmission(extSubmission, user.email, user.id).toSimpleSubmission()
+        val submitted = context.saveSubmission(extSubmission, user.email, user.id).toSimpleSubmission()
+
+        successfulSubmission.onNext(SuccessfulSubmission(
```
Juan-EBI: I think SuccessfulSubmission can contain both SecurityUser and ExtSubmission so even is generic and we can use for other purposes in the future.

commons/commons-model-extended/src/main/kotlin/ebi/ac/uk/extended/model/ExtSubmissionExtensions.kt
```diff
@@ -11,3 +13,6 @@ val ExtSubmission.allFiles
 
 val ExtSubmission.allFileListSections
     get(): List<ExtFileList> = allSections.mapNotNull { it.fileList }
+
+val ExtSubmission.releaseDate
+    get(): String = releaseTime?.toStringDate() ?: ""
```
Juan-EBI, line 18: is this actually making us lost information, now we do not know when the submission was released and when not. Nullable type is perfect for this so we should continue using it.

submission/notifications/src/main/kotlin/ebi/ac/uk/notifications/api/RtClient.kt
```diff
@@ -0,0 +1,47 @@
...
+        val body = getRequestBody(subject, owner, content)
+        val response = restTemplate.postForEntity<String>(url, body).body ?: throw InvalidResponseException()
+
+        return getTicketId(response)
+    }
+
+    private fun getRequestBody(subject: String, owner: String, content: String) = LinkedMultiValueMap<String, String>(
+        mapOf("content" to listOf(
+            "Queue: ${rtConfig.queue}\nSubject: $subject\nRequestor: $owner\nText: ${content.replace("\n", "\n ")}")))
+
```
Juan-EBI, line 37: maybe it will help formating as
   "Queue: ${rtConfig.queue}\n" +
   "Subject: $subject\n" ....

PR #194, opened 2020-04-09
Pivotal ID 172080341: File List Independant Format

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/172080341

Improve error message for files with unsupported format

### DIFF
--- a/commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/exception/InvalidExtensionException.kt
+++ b/commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/exception/InvalidExtensionException.kt
@@ -2,5 +2,5 @@ package ac.uk.ebi.biostd.exception
 
 class InvalidExtensionException(private val ext: String) : RuntimeException() {
     override val message: String
-        get() = "Unsupported submission or file list format $ext"
+        get() = "Unsupported page tab format $ext"
 }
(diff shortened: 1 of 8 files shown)


commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/exception/InvalidExtensionException.kt
```diff
@@ -1,7 +1,6 @@
 package ac.uk.ebi.biostd.exception
 
-import java.io.File
-
-class InvalidExtensionException(override val message: String) : RuntimeException(message) {
-    constructor(file: File) : this("Unrecognized data format, ${file.name} can not be processed")
+class InvalidExtensionException(private val ext: String) : RuntimeException() {
+    override val message: String
+        get() = "Unsupported submission or file list format $ext"
```
Juan-EBI, line 5: maybe unsuported pagetab format $ext

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/MultipartFileSubmissionApiTest.kt
```diff
@@ -242,6 +244,15 @@ internal class MultipartFileSubmissionApiTest(
             assertThat(savedSubmission["Title"]).isEqualTo("Test Submission")
         }
 
+        @Test
+        fun `invalid format file`() {
+            val submission = tempFolder.createFile("submission.txt")
+
+            assertThatExceptionOfType(WebClientException::class.java)
```
Juan-EBI, line 251: I tend to favor here assertThrows which is more kotlin friendly as a parameter as supported as generics. Said that still not big difference

jhoanmanuelms: Yeah, I like that one better too but it doesn't support inline which we use in the web client, that's why we have to use these in the itests

PR #196, opened 2020-04-20
Pivotal #Id: 172415684: Store user direct submission into user folder

Juan-EBI:
https://www.pivotaltracker.com/story/show/172415684

- fixed attribute support
- separated submission from data controllers


submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/web/handlers/SubmitWebHandler.kt
```diff
@@ -0,0 +1,60 @@
...
+import ebi.ac.uk.model.Submission
+import ebi.ac.uk.model.SubmissionMethod.FILE
+import ebi.ac.uk.model.SubmissionMethod.PAGE_TAB
+import ebi.ac.uk.model.extensions.rootPath
+import ebi.ac.uk.security.integration.model.api.SecurityUser
+import java.io.File
+
+private const val DIRECT_UPLOAD_PATH = "direct-uploads"
+
+class SubmitWebHandler(
```
jhoanmanuelms, line 22: Why did you change the name? I think `SubmissionwebHandler` is better

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/web/handlers/SubmitWebHandler.kt
```diff
@@ -0,0 +1,60 @@
...
+        return submissionService.submit(SubmissionRequest(submission, request.user, source, FILE, request.fileMode))
+    }
+
+    private fun sources(user: SecurityUser, submission: Submission, files: List<File> = emptyList()): FilesSource {
+        val request = RequestSources(user, files, submission.rootPath, subFolder(submission.accNo))
+        return sourceGenerator.submissionSources(request)
+    }
+
+    private fun withAttributes(submission: Submission, attrs: Map<String, String>): Submission {
+        attrs.forEach { submission[it.key] = it.value }
```
jhoanmanuelms, line 49: I think this one-liner would work? `submission.also { attrs.forEach { submission[it.key] = it.value } }`

Juan-EBI: do not like in this case as we still need to use consturctor parameter submission as []`notation does not work here

PR #207, opened 2020-05-06
Pivotal ID #  170598906:  refresh full support

Juan-EBI:
https://www.pivotaltracker.com/story/show/170598906


submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/service/filesystem/RefFilesService.kt
```diff
@@ -18,44 +18,45 @@ import ebi.ac.uk.io.FileUtils.reCreateDirectory
 import ebi.ac.uk.paths.FILES_PATH
 import ebi.ac.uk.paths.SubmissionFolderResolver
 import java.io.File
+import java.nio.file.Path
 
-class FilePersistenceService(
+class RefFilesService(
```
jhoanmanuelms, line 23: I think the previous name was clearer. Also, `RefFile` implies a direct correlation with the database table with the same name which is only used for regular files, not file list referenced ones

submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/service/SubmissionRepository.kt
```diff
@@ -22,8 +23,9 @@ class SubmissionRepository(
     private val submissionDbMapper: SubmissionDbMapper,
     private var toExtSubmissionMapper: ToExtSubmissionMapper
 ) {
-    fun getByAccNo(accNo: String): Submission = submissionDbMapper.toSubmission(getSubmission(accNo))
+    fun getByAccNo(accNo: String): Submission = getExtByAccNo(accNo).toSimpleSubmission()
 
+    @Deprecated("Extended model has been deprecated")
```
jhoanmanuelms, line 28: I understand the importance of deprecating but, since we're the only ones using this model right now, I think we could just remove it altogether

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/web/resources/SubmitResource.kt
```diff
@@ -99,4 +101,20 @@ class SubmitResource(private val submitWebHandler: SubmitWebHandler) {
         val request = ContentSubmitWebRequest(submission, user, JSON_PRETTY, mode, attributes.orEmpty(), emptyList())
         return submitWebHandler.submit(request)
     }
+
+    @PostMapping("refresh/{accNo}")
```
jhoanmanuelms, line 105: :+1:

submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/service/SubmissionRepository.kt
Juan-EBI: there are still some parts using it :(

PR #210, opened 2020-05-12
Pivotal #Id: 172798038: Add support between the submitter and submission owner

Juan-EBI:
https://www.pivotaltracker.com/story/show/172798038


commons/commons-bio/src/main/kotlin/ebi/ac/uk/api/security/GetOrRegisterUserRequest.kt
```diff
@@ -0,0 +1,8 @@
+package ebi.ac.uk.api.security
+
+data class GetOrRegisterUserRequest(
+    val register: Boolean,
+    val userEmail: String,
+    val userName: String?
+
```
jhoanmanuelms, line 7: Unnecessary blank

submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/integration/PersistenceContextImpl.kt
```diff
@@ -43,29 +43,30 @@ open class PersistenceContextImpl(
...
-        saveSubmission(SaveRequest(submission.copy(version = submission.version + 1), submitter, FileMode.MOVE))
+        saveSubmission(SaveRequest(submission.copy(version = submission.version + 1), FileMode.MOVE))
     }
 
-    private fun saveSubmission(submission: ExtSubmission, submitter: User, mode: FileMode): ExtSubmission {
+    private fun saveSubmission(submission: ExtSubmission, mode: FileMode): ExtSubmission {
         subRepository.expireActiveVersions(submission.accNo)
-        deleteSubmissionDrafts(submitter.id, submission.accNo)
-        return saveNewVersion(submission, submitter.email, mode)
+        deleteSubmissionDrafts(submission.submitter, submission.accNo)
```
jhoanmanuelms, line 57: I understand in some cases the owner and the submitter are the same but I'm not sure about deleting the drafts for both. I think we only need to delete it only for the owner which is the one that should have the draft in the first place. If they're the same, then deleting for both won't have any effect.

submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/integration/SubmissionSqlQueryService.kt
```diff
@@ -35,6 +35,8 @@ class SubmissionSqlQueryService(
     override fun getExistingFolder(accNo: String): File? =
         findSubmission(accNo)?.let { folderResolver.getSubmissionFolder(it.relPath).toFile() }
 
+    override fun getOwner(accNo: String): String? = findSubmission(accNo)?.owner?.email
```
jhoanmanuelms, line 38: Why is the owner nullable? I don't think we can have a submission without owner

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/web/handlers/SubmitWebHandler.kt
```diff
@@ -10,48 +10,80 @@ import ac.uk.ebi.biostd.submission.domain.service.SubmissionService
...
+        return submissionService.submit(Request(
+            submission = submission,
+            submitter = request.user,
+            sources = source,
+            method = PAGE_TAB,
+            mode = FileMode.MOVE
+        ))
+    }
+
+    private fun getOnBehalfUser(request: OnBehalfRequest): SecurityUser =
```
jhoanmanuelms, line 77: Are we sure about creating a user if it doesn't exist? I would expect `onBehalf` functionality for only work with existing users. We're probably mixing two things here.

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/web/handlers/SubmitWebHandler.kt
```diff
@@ -10,48 +10,80 @@ import ac.uk.ebi.biostd.submission.domain.service.SubmissionService
...
 import ebi.ac.uk.io.sources.FilesSource
 import ebi.ac.uk.model.Submission
 import ebi.ac.uk.model.SubmissionMethod.FILE
 import ebi.ac.uk.model.SubmissionMethod.PAGE_TAB
 import ebi.ac.uk.model.extensions.rootPath
+import ebi.ac.uk.security.integration.components.ISecurityService
 import ebi.ac.uk.security.integration.model.api.SecurityUser
 import java.io.File
 
+typealias Request = SubmissionRequest
```
jhoanmanuelms, line 25: Why? I think it looks clearer to just call it SubmissionRequest.

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/web/resources/MultipartSubmitResource.kt
```diff
@@ -101,9 +122,12 @@ class MultipartSubmitResource(
     @ResponseBody
     @ApiOperation("Make a submission using a TSV file. The given files will override the ones in the user folder")
     @ApiImplicitParam(name = "X-Session-Token", value = "User authentication token", required = true)
+    @Suppress("LongParameterList")
     fun submitMultipartTsv(
         @BioUser user: SecurityUser,
 
+        onBehalfRequest: OnBehalfRequest?,
```
jhoanmanuelms, line 129: I think we'll need docs for this one

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/converters/OnBehalfUserRequestResolver.kt
```diff
@@ -0,0 +1,31 @@
...
+import ebi.ac.uk.base.orFalse
+import org.springframework.core.MethodParameter
+import org.springframework.web.bind.support.WebDataBinderFactory
+import org.springframework.web.context.request.NativeWebRequest
+import org.springframework.web.method.support.HandlerMethodArgumentResolver
+import org.springframework.web.method.support.ModelAndViewContainer
+import java.net.URLDecoder
+import kotlin.text.Charsets.UTF_8
+
+class OnBehalfUserRequestResolver : HandlerMethodArgumentResolver {
```
jhoanmanuelms, line 16: :metal:

submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/integration/SubmissionSqlQueryService.kt
Juan-EBI: `in case submission does not exist `

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/web/handlers/SubmitWebHandler.kt
Juan-EBI: that is needed and that is current system behavior,  unified guys need user creation as they submit with the user on behalf of another and owner need to be new user.

submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/integration/PersistenceContextImpl.kt
Juan-EBI: mmm the problem for me is imagining we login in ST as manager or (future reviewer) role, you get a draft created, then original submission owner edit, then manager user check again and because it is a draft it will see not see author changes.

PR #211, opened 2020-05-18
Pivotal ID # 169759546: Add Number Of Views Metadata For Each Accession

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/169759546

- Paginate the stats response 
-  Validate that the submission exists before registering the stat

### DIFF
--- a/submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/stats/web/StatsResource.kt
+++ b/submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/stats/web/StatsResource.kt
@@ -1,16 +1,19 @@
 package ac.uk.ebi.biostd.stats.web
 
+import ac.uk.ebi.biostd.persistence.filter.PaginationFilter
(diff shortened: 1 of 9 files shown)


submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/stats/web/StatsResource.kt
```diff
@@ -35,17 +38,33 @@ class StatsResource(
     @GetMapping("/{type}", produces = [APPLICATION_JSON_VALUE])
     @ResponseBody
     @ApiOperation("List all the registered submission stats of the given type")
-    @ApiImplicitParam(name = "X-SESSION-TOKEN", value = "Authentication token", required = true, paramType = "header")
+    @ApiImplicitParams(value = [
+        ApiImplicitParam(
+            name = "X-SESSION-TOKEN",
```
Juan-EBI, line 43: this one is not needed as it added to every endpoint as part of the configuration

PR #212, opened 2020-05-18
Pivotal ID # 172476764: Refresh Interdatabase Support

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/172435555

Endpoint to get the extended submission


commons/commons-model-extended/build.gradle.kts
```diff
@@ -8,7 +10,8 @@ dependencies {
 
     implementation(Arrow)
     implementation(Guava)
-    implementation(kotlin("stdlib"))
+    implementation(KotlinStdLib)
+    implementation(JacksonKotlin)
```
Juan-EBI, line 14: I think this one is not needed

Juan-EBI, line 14: I think this one is not needed

submission/submission-webapp/src/main/resources/application.yml
```diff
@@ -23,10 +23,10 @@ spring:
       max-request-size: 20GB
       location: ${app.tempDirPath}
 
-logging.file: logs/application_logs.txt
 logging:
   file:
     max-history: 7
+    name: logs/application_logs.txt
```
Juan-EBI, line 29: did you test this?

jhoanmanuelms: Yes, it was deployed in dev and the logs were properly generated.

PR #224, opened 2020-06-09
Pivotal ID # 172889138: Stats Increment Endpoint

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/172889138

Endpoint to increment the stats value

### DIFF
--- a/submission/stats/src/main/kotlin/uk/ac/ebi/stats/service/SubmissionStatsService.kt
+++ b/submission/stats/src/main/kotlin/uk/ac/ebi/stats/service/SubmissionStatsService.kt
@@ -42,7 +42,7 @@ open class SubmissionStatsService(
     @Transactional
     open fun incrementAll(
         stats: List<SubmissionStat>
-    ): List<SubmissionStat> = stats.map(::increment).map(::save)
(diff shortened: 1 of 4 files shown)


submission/stats/src/main/kotlin/uk/ac/ebi/stats/service/SubmissionStatsService.kt
```diff
@@ -37,7 +37,12 @@ open class SubmissionStatsService(
     }
 
     @Transactional
-    open fun saveAll(stats: List<SubmissionStat>): List<SubmissionStat> = stats.map { save(it) }
+    open fun saveAll(stats: List<SubmissionStat>): List<SubmissionStat> = stats.map(::save)
+
+    @Transactional
+    open fun incrementAll(
+        stats: List<SubmissionStat>
+    ): List<SubmissionStat> = stats.map(::increment).map(::save)
```
Juan-EBI, line 45: I think rather here we want to map and then call `saveAll`

jhoanmanuelms: Done

PR #229, opened 2020-06-17
Pivotal ID # 173361277: Excel File List Empty Attributes

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/173361277

- Process empty excel cells
- Allow empty values in table attributes

### DIFF
--- a/commons/commons-util/src/main/kotlin/ebi/ac/uk/util/file/PoiExt.kt
+++ b/commons/commons-util/src/main/kotlin/ebi/ac/uk/util/file/PoiExt.kt
@@ -21,7 +21,14 @@ fun Sheet.asTsvList(): List<String> {
     return elements
 }
 
-private fun Row.asString(): String = asIterable().joinToString("\t") { it.valueAsString }
+private fun Row.asString(): String {
+    val cells = mutableListOf<String>()
+    for (idx in 0 until lastCellNum) {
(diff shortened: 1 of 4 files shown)


commons/commons-util/src/main/kotlin/ebi/ac/uk/util/file/PoiExt.kt
```diff
@@ -21,7 +21,14 @@ fun Sheet.asTsvList(): List<String> {
...
 }
 
-private fun Row.asString(): String = asIterable().joinToString("\t") { it.valueAsString }
+private fun Row.asString(): String {
+    val cells = mutableListOf<String>()
+    for (idx in 0 until lastCellNum) {
+        cells.add(getCell(idx)?.valueAsString ?: EMPTY)
+    }
+
+    return cells.joinToString("\t")
```
Juan-EBI, line 30: 99.9% we do not need use classic loops (0 ... x) and 80% we do not need mutable data structures.  What about

` return cellIterator().asSequence().joinToString(separator = "\t") { it.valueAsString }`

jhoanmanuelms: Unfortunately for this case we need to use an explicit loop since `cellIterator` method doesn't return empty cells which were causing the issue int he first place.

PR #232, opened 2020-06-25
Pivotal ID # 173457329: Project Access Tag Validation

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/173457329

Validate project access tag at submission time

### DIFF
--- a/submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/service/ParentInfoService.kt
+++ b/submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/service/ParentInfoService.kt
@@ -1,6 +1,8 @@
 package ac.uk.ebi.biostd.submission.service
 
+import ac.uk.ebi.biostd.persistence.exception.ProjectNotFoundException
 import ac.uk.ebi.biostd.persistence.integration.SubmissionQueryService
+import ac.uk.ebi.biostd.submission.exceptions.ProjectInvalidAccessTagException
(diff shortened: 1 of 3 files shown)


submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/service/ParentInfoService.kt
```diff
@@ -11,9 +13,13 @@ class ParentInfoService(private val queryService: SubmissionQueryService) {
     }
 
     private fun parentInfo(parentAccNo: String): ParentInfo {
-        require(queryService.existByAccNo(parentAccNo)) { "The project '$parentAccNo' doesn't exist" }
+        require(queryService.existByAccNo(parentAccNo)) { throw ProjectNotFoundException(parentAccNo) }
+
+        val accessTags = queryService.getAccessTags(parentAccNo).filterNot { it == SubFields.PUBLIC_ACCESS_TAG.value }
```
Juan-EBI, line 18: I am wondering if we should query-specific project access tag which match accNo instead of getting all of them by relation.

jhoanmanuelms: I get your point but we would still need to get all the other tags for the parent submission so it would be two database calls instead of one

Juan-EBI: agree

PR #236, opened 2020-07-02
Pivotal ID #173574802: submissions files permissions

Juan-EBI:
https://www.pivotaltracker.com/story/show/173574802


commons/commons-util/src/main/kotlin/ebi/ac/uk/io/FileUtils.kt
```diff
@@ -4,18 +4,20 @@ import ebi.ac.uk.io.FileUtilsHelper.createFileHardLink
...
 import java.nio.file.StandardCopyOption
-import java.nio.file.attribute.FileAttribute
 import java.nio.file.attribute.PosixFilePermission
 import java.nio.file.attribute.PosixFilePermissions
 import java.nio.file.attribute.PosixFilePermissions.asFileAttribute
 import kotlin.streams.toList
 
 val ONLY_USER: Set<PosixFilePermission> = PosixFilePermissions.fromString("rwx------")
+val READ_ONLY_GROUP: Set<PosixFilePermission> = PosixFilePermissions.fromString("rwxr-x---")
+val ALL_CAN_READ: Set<PosixFilePermission> = PosixFilePermissions.fromString("rwxr-xr-x")
```
jhoanmanuelms, line 20: I thought what we needed here was to have the permissions as in `ALL_GROUP` constant like "rwxrwxr---"

Juan-EBI: that one is in the security service as it is only used by this class. But yes agree is worth it to have them all here while we move then into an enum of something like that

PR #239, opened 2020-07-14
pivotal-173792650: Replace base path for multiple absolute paths 

Juan-EBI:
https://www.pivotaltracker.com/story/show/173792650

### DIFF
--- a/commons/commons-bio/src/main/kotlin/ebi/ac/uk/paths/SubmissionFolderResolver.kt
+++ b/commons/commons-bio/src/main/kotlin/ebi/ac/uk/paths/SubmissionFolderResolver.kt
@@ -1,25 +1,22 @@
 package ebi.ac.uk.paths
 
-import ebi.ac.uk.extended.model.ExtSubmission
 import java.nio.file.Path
 
 const val FILES_PATH = "Files"
-const val SUBMISSION_PATH = "submission"
-const val FTP_FOLDER = "submission/ftp"
 
-class SubmissionFolderResolver(private val basePath: Path) {
+class SubmissionFolderResolver(
+    private val submissionFolder: Path,
+    private val ftpFolder: Path
+) {
 
-    fun getSubmissionFtpFolder(submission: ExtSubmission): Path =
-        basePath.resolve(FTP_FOLDER).resolve(submission.relPath)
-
-    fun getSubmissionFolder(submission: ExtSubmission): Path =
-        basePath.resolve(SUBMISSION_PATH).resolve(submission.relPath)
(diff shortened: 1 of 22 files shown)


commons/commons-bio/src/main/kotlin/ebi/ac/uk/paths/SubmissionFolderResolver.kt
```diff
@@ -1,25 +1,22 @@
...
+    private val submissionFolder: Path,
+    private val ftpFolder: Path
+) {
 
-    fun getSubmissionFtpFolder(submission: ExtSubmission): Path =
-        basePath.resolve(FTP_FOLDER).resolve(submission.relPath)
-
-    fun getSubmissionFolder(submission: ExtSubmission): Path =
-        basePath.resolve(SUBMISSION_PATH).resolve(submission.relPath)
+    fun getSubmissionFtpFolder(submissionRelPath: String): Path =
```
jhoanmanuelms, line 12: Silly detail but, don't these two methods fit in a single line?

submission/submission-webapp/src/itest/resources/application.yml
```diff
@@ -18,7 +18,9 @@ spring:
     port: 5672
 
 app:
-  basepath:
+  submissionPath:
+  ftpPath:
+  magicPath:
```
jhoanmanuelms, line 23: Don't we need this property in the application's main config file?

Also, I didn't see it's being used in the code, only in the `ApplicationProperties` class but I assume it's the user magic folder? In any case, I think the property name can be improved or at least add a comment indicating what it's used for.

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/common/config/GeneralConfig.kt
```diff
@@ -10,6 +10,7 @@ import org.springframework.context.annotation.Configuration
 @Configuration
 @EnableConfigurationProperties(ApplicationProperties::class)
 internal class GeneralConfig {
+
```
jhoanmanuelms, line 13: image

submission/submission-webapp/src/itest/resources/application.yml
Juan-EBI: ohh yes let me remove this one, is for future changes

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/common/config/GeneralConfig.kt
Juan-EBI: hahaha

PR #240, opened 2020-07-14
Pivotal ID # 173569202: Empty Section Parent Cell

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/173569202

Improve excel reader:
- When all the cells in a row are empty, this will be interpreted as an empty line
- Remove empty elements from the end of a row
- Remove all empty rows starting at the last populated
- Add unit tests for these edge cases

Fixes # 173569202

### DIFF
--- a/commons/commons-util/src/main/kotlin/ebi/ac/uk/dsl/excel/ExcelSheet.kt
+++ b/commons/commons-util/src/main/kotlin/ebi/ac/uk/dsl/excel/ExcelSheet.kt
@@ -1,5 +1,6 @@
 package ebi.ac.uk.dsl.excel
 
+import ebi.ac.uk.base.EMPTY
 import org.apache.poi.xssf.streaming.SXSSFSheet
 
 class ExcelSheet(private val sheet: SXSSFSheet) {
@@ -7,7 +8,10 @@ class ExcelSheet(private val sheet: SXSSFSheet) {
 
     fun row(cellsBuilder: ExcelRow.() -> Unit) = ExcelRow(newRow()).apply { cellsBuilder() }
 
-    fun emptyRow() = row { }
+    fun emptyRow() = row {
+        cell(EMPTY)
+        cell(EMPTY)
+    }
 
     private fun newRow() = sheet.createRow(rowIterator.next())
 }
--- a/commons/commons-util/src/main/kotlin/ebi/ac/uk/util/file/PoiExt.kt
+++ b/commons/commons-util/src/main/kotlin/ebi/ac/uk/util/file/PoiExt.kt
@@ -5,20 +5,19 @@ import org.apache.poi.ss.usermodel.Cell
 import org.apache.poi.ss.usermodel.CellType.NUMERIC
 import org.apache.poi.ss.usermodel.Row
 import org.apache.poi.ss.usermodel.Sheet
-import java.util.LinkedList
 
 /**
  * Retrieves the sheet representation as a list of tsv separated string List. Note that as stream reader ignored empty
  * rows they are completed with emtpy values.
  */
 fun Sheet.asTsvList(): List<String> {
-    val elements = LinkedList<String>()
+    val elements = mutableListOf<String>()
     for (row in this) {
         while (elements.size < row.rowNum) elements.add(EMPTY)
         elements.add(row.asString())
     }
 
-    return elements
+    return elements.trim()
 }
 
 private fun Row.asString(): String {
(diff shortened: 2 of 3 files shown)


commons/commons-util/src/main/kotlin/ebi/ac/uk/dsl/excel/ExcelSheet.kt
```diff
@@ -1,13 +1,17 @@
...
+import ebi.ac.uk.base.EMPTY
 import org.apache.poi.xssf.streaming.SXSSFSheet
 
 class ExcelSheet(private val sheet: SXSSFSheet) {
     private val rowIterator = buildIterator()
 
     fun row(cellsBuilder: ExcelRow.() -> Unit) = ExcelRow(newRow()).apply { cellsBuilder() }
 
-    fun emptyRow() = row { }
+    fun emptyRow() = row {
```
Juan-EBI, line 11: this is only used as part of test/itest , IMO it should be in  test class package or function

commons/commons-util/src/main/kotlin/ebi/ac/uk/util/file/PoiExt.kt
```diff
@@ -5,20 +5,19 @@ import org.apache.poi.ss.usermodel.Cell
...
 import org.apache.poi.ss.usermodel.Sheet
-import java.util.LinkedList
 
 /**
  * Retrieves the sheet representation as a list of tsv separated string List. Note that as stream reader ignored empty
  * rows they are completed with emtpy values.
  */
 fun Sheet.asTsvList(): List<String> {
-    val elements = LinkedList<String>()
+    val elements = mutableListOf<String>()
```
Juan-EBI, line 14: mutable list in kotlin is by default a ArrayList which performance for remove operations is terrible. (as backing array need to be re created)

commons/commons-util/src/main/kotlin/ebi/ac/uk/util/file/PoiExt.kt
```diff
@@ -27,7 +26,21 @@ private fun Row.asString(): String {
...
+    }
+}
+
+private fun MutableList<String>.trim(): MutableList<String> {
+    if (last().isBlank()) {
+        for (idx in indices.reversed()) {
+            if (elementAt(idx).isNotBlank()) break
+            removeAt(idx)
+        }
+    }
```
Juan-EBI, line 41: this for looks overcomplex, definitely kotlin have a cleaner declarative approach. What about

```
this.asReversed().dropWhile { it.isBlank() }
```

commons/commons-util/src/test/kotlin/ebi/ac/uk/util/file/ExcelReaderTest.kt
```diff
@@ -30,6 +30,8 @@ class ExcelReaderTest(private val temporaryFolder: TemporaryFolder) {
                 row {
                     cell("Study")
                     cell("SECT-001")
+                    cell("")
```
Juan-EBI, line 33: here we are complicating main test case to include corner case scenarios testing, I think instead, we should add an extra simple test case `when extra empty cells` and keep this one only for happy path.

commons/commons-util/src/main/kotlin/ebi/ac/uk/dsl/excel/ExcelSheet.kt
jhoanmanuelms: All DSLs we have are used for testing, perhaps we could create a separate module in another PR and move them all? Right now they're in `commons-util` but we could have something like `commons-dsl` ?

commons/commons-util/src/main/kotlin/ebi/ac/uk/util/file/PoiExt.kt
jhoanmanuelms: Reverted

commons/commons-util/src/main/kotlin/ebi/ac/uk/util/file/PoiExt.kt
jhoanmanuelms: Thanks! I knew there was a better way. You're like my Kotlin linter :laughing: 

Only one detail, I ended up using: `dropLastWhile { it.isBlank() }`

commons/commons-util/src/test/kotlin/ebi/ac/uk/util/file/ExcelReaderTest.kt
jhoanmanuelms: Done

PR #243, opened 2020-07-15
Pivotal ID # 172817960: Resubmit Notification

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/172817960

- Changes in the architecture to use queues instead of events for notifications
- Get the extended submission from the service using the URL in the queue message
- Add logic to load different templates based on the submission version
- Create/update the RT ticket depending on its existence

Delivers # 172817960


submission/notifications/src/main/kotlin/ebi/ac/uk/notifications/service/RtNotificationService.kt
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    private fun createTicket(accNo: String, subject: String, owner: String, content: String) {
+        val ticketId = rtClient.createTicket(accNo, subject, owner, content)
+        notificationPersistenceService.saveRtNotification(accNo, ticketId)
+    }
+
+    private fun templateContent(submission: ExtSubmission): String {
+        val template = when (submission.version) {
+            1 -> SUCCESSFUL_SUBMISSION_TEMPLATE
```
Juan-EBI, line 41: this looks like let escape some of the business knowledge of ext submission, maybe we should add extended property isNew or is aReSubmission.

Still not sure this is a 100% good recomendation so ignore if you do not agree.

jhoanmanuelms: This would require some other changes in several other places making this PR even bigger. I think we could do this in a separate PR.

PR #244, opened 2020-07-17
Pivotal ID # 173457302: RelPath Calculation

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/173457302

In an accession with several number groups only the last will be used

Fixes # 173457302


submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/util/AccNoPatternUtil.kt
```diff
@@ -13,7 +13,7 @@ private const val EXPECTED_PATTERN = "([A-Z,-]*)"
 
 class AccNoPatternUtil {
     private val prefix = ACC_PATTERN.format("([A-Z,-]*)").toPattern()
-    private val extractionPattern = "(\\D*)([0-9]+)(\\D*)".toPattern()
+    private val extractionPattern = "^(.*?)(\\d+)(\\D*)\$".toPattern()
```
Juan-EBI, line 16: what about `^(.*?)(\d+)$`

jhoanmanuelms: Done

PR #245, opened 2020-07-20
Pivotal ID # 173750220: Author TSV Serialization

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/173750220

Fix serialization for subsections without accession

### DIFF
--- a/commons/commons-serialization/src/test/kotlin/ac/uk/ebi/biostd/tsv/TsvToStringSerializerTest.kt
+++ b/commons/commons-serialization/src/test/kotlin/ac/uk/ebi/biostd/tsv/TsvToStringSerializerTest.kt
@@ -98,7 +98,7 @@ class TsvToStringSerializerTest {
     }
 
     @Test
-    fun `serialize subsection without accession`() {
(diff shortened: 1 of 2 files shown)


commons/commons-serialization/src/test/kotlin/ac/uk/ebi/biostd/tsv/TsvToStringSerializerTest.kt
```diff
@@ -96,6 +97,30 @@ class TsvToStringSerializerTest {
         assertThat(testInstance.serialize(section)).isEqualToIgnoringNewLines(expectedTsv)
     }
 
+    @Test
+    fun `serialize subsection without accession`() {
```
Juan-EBI, line 101: what about "serialize subsection without accession but the parent accession"  and not sure if we need test for ""serialize subsection with accession and parent accession"

jhoanmanuelms: Done

PR #254, opened 2020-08-11
Pivotal #ID 174275642: Submission loading strategy

Juan-EBI:
https://www.pivotaltracker.com/story/show/174275642

### DIFF
--- a/submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/model/DbSection.kt
+++ b/submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/model/DbSection.kt
@@ -21,38 +21,25 @@ import javax.persistence.OneToOne
 import javax.persistence.OrderBy
 import javax.persistence.Table
 
-internal const val SECTION_FULL_GRAPH = "Section.fullGraph"
 internal const val SECTION_SIMPLE_GRAPH = "Section.simpleGraph"
+private const val ATTRIBUTES_GRAPH = "Object.attributesGraph"
 
 @Entity
 @Table(name = "Section")
 @NamedEntityGraphs(value = [
     NamedEntityGraph(
-        name = SECTION_FULL_GRAPH,
-        attributeNodes = [
-            Node(LINKS, subgraph = "attrs"),
-            Node(ATTRS),
-            Node(FILES, subgraph = "attrs"),
-            Node(SECTS, subgraph = SECTION_FULL_GRAPH)
-        ]),
-    NamedEntityGraph(
         name = SECTION_SIMPLE_GRAPH,
         attributeNodes = [
             Node(ATTRS),
             Node(SECTS),
-            Node(LINKS, subgraph = "attrs-sub-graph"),
-            Node(FILES, subgraph = "attrs-sub-graph")],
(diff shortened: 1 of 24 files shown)


submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/model/DbSection.kt
```diff
@@ -15,24 +15,44 @@ import javax.persistence.Id
...
+            Node(ATTRS),
+            Node(FILES, subgraph = "attrs"),
+            Node(SECTS, subgraph = SECTION_FULL_GRAPH)
+        ]),
+    NamedEntityGraph(
+        name = SECTION_SIMPLE_GRAPH,
+        attributeNodes = [
+            Node(ATTRS),
+            Node(SECTS),
+            Node(LINKS, subgraph = "attrs-sub-graph"),
```
jhoanmanuelms, line 43: Couldn't you just use the same "attrs" subgraph here?

submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/service/SubmissionRepository.kt
```diff
@@ -42,9 +52,30 @@ class SubmissionRepository(
             .map { it.asSimpleSubmission() }
     }
 
-    private fun getSubmission(accNo: String) =
-        submissionRepository.getByAccNo(accNo) ?: throw SubmissionNotFoundException(accNo)
+    /**
+     * Load submission information strategy used is basically first load submission and then load each section and it
```
jhoanmanuelms, line 56: typo: its subsections

submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/service/SubmissionRepository.kt
```diff
@@ -42,9 +52,30 @@ class SubmissionRepository(
...
+     * subsections recursively starting from the root section.
+     */
+    private fun submission(accNo: String, version: Int? = null): DbSubmission {
```
jhoanmanuelms, line 59: I think `loadSubmission` would be clearer

submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/service/SubmissionRepository.kt
```diff
@@ -42,9 +52,30 @@ class SubmissionRepository(
...
+        val submission = findSubmission(accNo, version)
+        submission ?: throw SubmissionNotFoundException(accNo)
+        loadSection(submission.rootSectionId)
+
+        logger.info { "Loaded submission $accNo" }
+        return submission
+    }
+
+    private fun findSubmission(accNo: String, version: Int?): DbSubmission? {
+        return when (version) {
```
jhoanmanuelms, line 71: Are we sure we need this logic? why not just leave the version with a default value?

Juan-EBI: I notice that was kind of misleading. With default parameter, we do not know if we need a specific version or we need an active one

PR #262, opened 2020-08-25
Pivotal ID # 174275207: Test Using The Persistence Layer

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/174275207

- Remove persistence layer assertions
- Remove deprecated methods
- Remove unnecessary configuration properties

Delivers # 174275207

### DIFF
--- a/submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/refresh/SubmissionRefreshApiTest.kt
+++ b/submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/refresh/SubmissionRefreshApiTest.kt
@@ -6,7 +6,6 @@ import ac.uk.ebi.biostd.common.config.PersistenceConfig
 import ac.uk.ebi.biostd.itest.common.BaseIntegrationTest
 import ac.uk.ebi.biostd.itest.common.SecurityTestService
 import ac.uk.ebi.biostd.itest.entities.SuperUser
-import ac.uk.ebi.biostd.persistence.model.DbFile
 import ac.uk.ebi.biostd.persistence.model.DbSubmission
 import ac.uk.ebi.biostd.persistence.model.DbSubmissionAttribute
 import ac.uk.ebi.biostd.persistence.model.DbTag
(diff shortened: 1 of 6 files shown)


submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/refresh/SubmissionRefreshApiTest.kt
```diff
@@ -99,48 +105,47 @@ internal class SubmissionApiTest(private val tempFolder: TemporaryFolder) : Base
...
+            assertThat(extSubmission.attributes.map { it.name to it.value }).containsExactlyElementsOf(attributes)
 
-            assertThat(submission.rootSection.type).isEqualTo("Study")
-            assertThat(submission.rootSection.files).hasSize(2)
+            assertThat(extSubmission.section.type).isEqualTo("Study")
+            assertThat(extSubmission.section.files).hasSize(2)
 
-            assertFile(submission.rootSection.files.first(), "regular")
-            assertFile(submission.rootSection.files.last(), "duplicated")
+            extSubmission.section.files.first().ifLeft { assertFile(it, "regular") }
```
Juan-EBI, line 143: I think this is incorrect as if it is a right test will not fail. We need to ensure is a left AND  is the expected file (same for the next file)

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/refresh/SubmissionRefreshApiTest.kt
```diff
@@ -150,9 +155,11 @@ internal class SubmissionApiTest(private val tempFolder: TemporaryFolder) : Base
...
             submission.title = NEW_SUBTITLE
             submission.attributes = sortedSetOf(DbSubmissionAttribute(ATTR_NAME, NEW_ATTR_VALUE, 1))
-            submissionRepository.save(submission)
+            submissionDataRepository.save(submission)
         }
 
-        private fun getSubmissionDb(): DbSubmission = submissionRepository.readByAccNoAndVersionGreaterThan(ACC_NO, 0)!!
+        private fun getExtSubmission(): ExtSubmission = submissionRepository.getExtByAccNo(ACC_NO)
+
+        private fun getSubmissionDb(): DbSubmission = submissionDataRepository.getBasicWithAttributes(ACC_NO)
```
Juan-EBI, line 163: we should not use DB submission at integration level

jhoanmanuelms: This case is special because its purpose is testing the refresh feature which was actually created to reflect changes in the database. I understand your point but I think here we should make an exception.

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/refresh/SubmissionRefreshApiTest.kt
jhoanmanuelms: Got it

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/refresh/SubmissionRefreshApiTest.kt
Juan-EBI: yes make sense

PR #263, opened 2020-08-26
Pivotal ID # 173919045: Restrict Execute Permissions On HTTP Uploaded…

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/173919045

Set execute permission only for folders


commons/commons-util/src/main/kotlin/ebi/ac/uk/io/FileUtils.kt
```diff
@@ -19,9 +19,11 @@ import kotlin.streams.toList
 
 val ONLY_USER: Set<PosixFilePermission> = PosixFilePermissions.fromString("rwx------")
 val READ_ONLY_GROUP: Set<PosixFilePermission> = PosixFilePermissions.fromString("rwxr-x---")
+val READ_ONLY_GROUP_NO_EXE: Set<PosixFilePermission> = PosixFilePermissions.fromString("rw-r-----")
```
Juan-EBI, line 22: this is kind of getting out of control I am wondering if we should name permission contatns like `RWX_____`

jhoanmanuelms: Yeah it might be better so is easily understandable anywhere in the code

PR #265, opened 2020-09-01
Pivotal ID # 173919045: Restrict Execute Permissions On HTTP Uploaded Files

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/173919045

- Implement a mechanism to provide separated permissions for folders and files
- Set execute permission only for folders
- Improve permissions constants names

Fixes # 173919045


commons/commons-util/src/main/kotlin/ebi/ac/uk/io/FileUtils.kt
```diff
@@ -152,17 +146,60 @@ internal object FileUtilsHelper {
         Files.createSymbolicLink(createParentDirectories(link, permissions), target)
     }
 
-    fun copyFolder(source: Path, target: Path, permissions: Set<PosixFilePermission>) {
+    fun copyFolder(
+        source: Path,
+        target: Path,
+        filePermissions: Set<PosixFilePermission>,
+        folderPermissions: Set<PosixFilePermission>
+    ) = applyFileVisitor(source, target, CopyFileVisitor(source, target, filePermissions, folderPermissions))
```
Juan-EBI, line 154: I think `applyFileVisitor` is not that good as it save as like two lines of code but is completely unknow what "apply" means it is way clear IMO just call visitor.

submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/service/filesystem/SubmissionFilesPermissions.kt
```diff
@@ -0,0 +1,16 @@
+package ac.uk.ebi.biostd.persistence.service.filesystem
+
+import java.nio.file.attribute.PosixFilePermission
+import java.nio.file.attribute.PosixFilePermissions
+
+enum class SubmissionFilesPermissions(val stringValue: String) {
+    ALL_CAN_READ("rwxr-xr-x"),
```
Juan-EBI, line 7: This looks like a duplication of `RW_RW____` constants efforts, I mean the idea of those constants I understand is have al permissions string there, but now we also have then here. 

I am wondering if instead of constants maybe we need a sealed class with object (in case we need more features in them)

submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/service/filesystem/SubmissionFilesPermissions.kt
```diff
@@ -0,0 +1,16 @@
+package ac.uk.ebi.biostd.persistence.service.filesystem
+
+import java.nio.file.attribute.PosixFilePermission
+import java.nio.file.attribute.PosixFilePermissions
+
+enum class SubmissionFilesPermissions(val stringValue: String) {
```
Juan-EBI, line 6: it looks to me that this class will be easier to model as a simple data class with the two `RW_RW____` constants values (files and folder).

jhoanmanuelms: You're right, this seems like an overhead. I just went ahead and removed the specific enum class and passed down the corresponding file/folder permissions. I think it's much clearer.

PR #268, opened 2020-09-09
Pivotal ID # 172671360: Warn Users About Upcoming Releases

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/172671360

Upcoming releases RT notifications

### DIFF
--- a/events/events-publisher/src/main/kotlin/uk/ac/ebi/events/service/EventsPublisherService.kt
+++ b/events/events-publisher/src/main/kotlin/uk/ac/ebi/events/service/EventsPublisherService.kt
@@ -17,18 +17,18 @@ class EventsPublisherService(
     fun securityNotification(notification: SecurityNotification) =
         rabbitTemplate.convertAndSend(BIOSTUDIES_EXCHANGE, SECURITY_NOTIFICATIONS_ROUTING_KEY, notification)
 
-    fun submissionSubmitted(submission: ExtSubmission, ownerEmail: String) =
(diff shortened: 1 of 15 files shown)


events/events-publisher/src/main/kotlin/uk/ac/ebi/events/service/EventsPublisherService.kt
```diff
@@ -1,30 +1,34 @@
...
-        val submissionNotification = SubmissionSubmitted(
-            accNo = submission.accNo,
-            pagetabUrl = "$instanceBaseUrl/submissions/${submission.accNo}.json",
-            extTabUrl = "$instanceBaseUrl/submissions/extended/${submission.accNo}",
-            extUserUrl = "$instanceBaseUrl/security/users/extended/$ownerEmail")
+    fun submissionSubmitted(submission: ExtSubmission, ownerEmail: String) =
+        rabbitTemplate.convertAndSend(
+            BIOSTUDIES_EXCHANGE, SUBMISSIONS_ROUTING_KEY, submissionMessage(submission, ownerEmail))
+
+    fun submissionReleased(submission: ExtSubmission, ownerEmail: String) =
```
Juan-EBI, line 24: I think the owner is not needed here. (exte already have field)

jhoanmanuelms: Right, I think it comes from when it didn't have it. Fixed.

Juan-EBI (approved):
LGTM

PR #269, opened 2020-09-11
Pivotal ID # 173783625: Partial Updates Timestamp

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/173783625

Add the current timestamp to the submission submitted event

Delivers #173783625


Juan-EBI (approved):
LGTM

PR #274, opened 2020-10-05
Pivotal ID #171026062: Submission files MD5 Generation

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/171026062

- Create a method for MD5 generation
- Verify MD5 to decide whether or not a file should be copied


submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/service/filesystem/FilesService.kt
```diff
@@ -122,25 +124,32 @@ class FilesService(
...
     ) {
         val source = extFile.file
-        val target = folder.resolve(extFile.fileName)
-
-        logger.info { "copying file ${source.absolutePath} into ${target.absolutePath}" }
-        copyOrReplaceFile(source, target, filePermissions, folderPermissions)
+        val target = tempFolder.resolve(extFile.fileName)
+        val current = submissionFilesFolder.resolve(extFile.fileName)
+
+        if (current.notExist() || source.md5() != current.md5()) {
```
Juan-EBI, line 136: not the following the logic here, why if md5 has not been changed we copy the file while if md5 is different we move it?

Juan-EBI (commented):
Not clear for me here changes, I think expectation here is to calculate and store files md5 in db rather than take decision if copy or move files.

Juan-EBI (approved):
LGTM

PR #281, opened 2020-11-02
Pivotal ID # 175359447: Async Concurrent Processing Submission

jhoanmanuelms:
Validate a submission doesn't have a current processing version before accepting a new submission request


Juan-EBI (approved):
LGTM

PR #283, opened 2020-11-08
Pivotal #ID 175632448: Extra persistence layer interfaces to avoid dependencies on SQL classes

Juan-EBI:
https://www.pivotaltracker.com/story/show/175632448


submission/persistence/src/main/kotlin/ac/uk/ebi/biostd/persistence/common/SubmissionTypes.kt
```diff
@@ -1,8 +1,7 @@
 package ac.uk.ebi.biostd.persistence.common
 
-enum class SubmissionTypes(val value: String) {
+internal enum class SubmissionTypes(val value: String) {
     Study("Study"),
     Project("Project"),
     Array("Array"),
-    Compound("Compound");
```
awaisathar, line 7: For consistency, we should probably keep Compound in or get rid of all unused submission types (everything but Project?)

jhoanmanuelms (approved):
LGTM. Minor changes.

PR #285, opened 2020-11-10
Pivotal ID # 175359447: Async Concurrent Processing Submission

jhoanmanuelms:
- Validate all other possible submission status
- Avoid calculating the MD5 every time an ExtFile is instantiated


Juan-EBI (approved):
LGTM

PR #289, opened 2020-11-13
Pivotal ID: 175632448: Extra persistence layer interfaces to avoid dependencies on SQL classes 

Juan-EBI:


jhoanmanuelms (approved):
LGTM

PR #290, opened 2020-11-14
Pivotal ID: 175632448: Extra persistence layer interfaces to avoid dependencies on SQL classes

Juan-EBI:


submission/persistence-common-api/src/main/kotlin/ac/uk/ebi/biostd/persistence/common/service/PersistenceService.kt
```diff
@@ -0,0 +1,10 @@
+package ac.uk.ebi.biostd.persistence.common.service
+
+@Suppress("TooManyFunctions")
+interface PersistenceService {
```
jhoanmanuelms, line 4: Just 5 triggers too many functions??

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/ext/getSimpleByAccNo.kt
```diff
@@ -0,0 +1,6 @@
+package ac.uk.ebi.biostd.submission.ext
+
+import ac.uk.ebi.biostd.persistence.common.service.SubmissionQueryService
+import ebi.ac.uk.extended.mapping.to.toSimpleSubmission
+
+fun SubmissionQueryService.getSimpleByAccNo(accNo: String) = getExtByAccNo(accNo).toSimpleSubmission()
```
jhoanmanuelms, line 6: Why is this an extension?

submission/persistence-common-api/src/main/kotlin/ac/uk/ebi/biostd/persistence/common/service/SubmissionOperations.kt
```diff
@@ -0,0 +1,44 @@
...
+import ac.uk.ebi.biostd.persistence.common.request.SaveSubmissionRequest
+import ac.uk.ebi.biostd.persistence.common.request.SubmissionFilter
+import ebi.ac.uk.extended.model.ExtSubmission
+import ebi.ac.uk.extended.model.FileMode
+import org.springframework.data.domain.Page
+import java.io.File
+import java.time.OffsetDateTime
+
+interface SubmissionRequestService {
+    fun saveAndProcessSubmissionRequest(saveRequest: SaveSubmissionRequest): ExtSubmission
```
jhoanmanuelms, line 13: As discussed in my last PR this should be separated?

jhoanmanuelms (approved):
LGTM. Please consider my comments before merging.

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/ext/getSimpleByAccNo.kt
Juan-EBI: This service comes from the persistence module so if we won't allow it in class we need to add the serialization module as a dependency on persistence. As that is not necessary at all. I Declared this an extension of main product (that already depends on the serialization model)

PR #293, opened 2020-11-19
Pivotal #Id: Rename fileName to path

Juan-EBI:
https://www.pivotaltracker.com/story/show/175101522


jhoanmanuelms:
Moving changes to a separated branch

PR #294, opened 2020-11-20
Pivotal ID # 175101522: File Path Field

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/175101522

Add file path to the extended model


Juan-EBI (approved):
LGTM

PR #296, opened 2020-11-22
Pivotal: ID: 175839554 : include user tag as part of extended submission tags

Juan-EBI:
https://www.pivotaltracker.com/story/show/175839554

### DIFF
--- a/commons/commons-model-extended-serialization/src/test/kotlin/uk/ac/ebi/extended/serialization/serializers/ExtFileSerializerTest.kt
+++ b/commons/commons-model-extended-serialization/src/test/kotlin/uk/ac/ebi/extended/serialization/serializers/ExtFileSerializerTest.kt
@@ -22,11 +22,12 @@ class ExtFileSerializerTest(private val tempFolder: TemporaryFolder) {
         val file = tempFolder.createFile("test-file.txt", "content")
         val extFile = ExtFile(
             file = file,
-            fileName = "test-file.txt",
+            fileName = "my_path/test-file.txt",
             attributes = listOf(ExtAttribute("Type", "Data", false))
         )
         val expectedJson = jsonObj {
             "fileName" to "test-file.txt"
+            "path" to "my_path/test-file.txt"
             "file" to file.absolutePath
             "attributes" to jsonArray(jsonObj {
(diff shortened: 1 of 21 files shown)


commons/commons-model-extended-serialization/src/test/kotlin/uk/ac/ebi/extended/serialization/serializers/ExtFileSerializerTest.kt
```diff
@@ -22,12 +22,12 @@ class ExtFileSerializerTest(private val tempFolder: TemporaryFolder) {
         val file = tempFolder.createFile("test-file.txt", "content")
         val extFile = ExtFile(
             file = file,
-            fileName = "test/path/test-file.txt",
```
jhoanmanuelms, line 25: I included this to cover the scenario of a file with an inner folder, I think it's worth having it.

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/security/web/SecurityMapper.kt
```diff
@@ -28,9 +26,8 @@ class SecurityMapper {
 
     private fun getAllow(user: SecurityUser): List<String> {
         val accessTags = ArrayList<String>()
-        accessTags.add("~" + user.email)
-        accessTags.add("#" + user.id)
```
jhoanmanuelms, line 32: Are you sure we don't need this anymore?

jhoanmanuelms (approved):

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/security/web/SecurityMapper.kt
Juan-EBI: no we are not

awaisathar: Feel free to delete the "~" and the user id (including #). The front end only needs to match the accessTags in the submission with the allow list and already ignores # and ~.

PR #297, opened 2020-11-25
Pivotal # 175163839: Submission Operations Refactor

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/175163839

- Simplify SubmissionOperations interface and implementation
- Adjust services that used these queries in order to avoid multiple DB requests
- Adjust tests
- Rename SimpleSubmission to BasicSubmission in order to preserve the same terms all over the code


Juan-EBI (approved):
LGTM

PR #302, opened 2020-12-02
Pivotal #Id : 175632448: Extra persistence layer interfaces to avoid dependencies on SQL classes

Juan-EBI:
https://www.pivotaltracker.com/story/show/175632448

### DIFF
--- a/submission/persistence-sql/src/main/kotlin/ac/uk/ebi/biostd/persistence/integration/services/SqlSubmissionRequestService.kt
+++ b/submission/persistence-sql/src/main/kotlin/ac/uk/ebi/biostd/persistence/integration/services/SqlSubmissionRequestService.kt
@@ -1,0 +1,43 @@
+package ac.uk.ebi.biostd.persistence.integration.services
+
+import ac.uk.ebi.biostd.persistence.common.request.SaveSubmissionRequest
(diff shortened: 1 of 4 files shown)


submission/persistence-sql/src/main/kotlin/ac/uk/ebi/biostd/persistence/integration/services/SqlSubmissionRequestService.kt
```diff
@@ -0,0 +1,43 @@
+package ac.uk.ebi.biostd.persistence.integration.services
+
+import ac.uk.ebi.biostd.persistence.common.request.SaveSubmissionRequest
+import ac.uk.ebi.biostd.persistence.common.service.SubmissionRequestService
+import ac.uk.ebi.biostd.persistence.repositories.LockExecutor
+import ebi.ac.uk.extended.model.ExtSubmission
+import ebi.ac.uk.extended.model.FileMode
+import org.springframework.transaction.annotation.Transactional
+
+@Suppress("TooManyFunctions")
```
jhoanmanuelms, line 10: I don't think this suppress is necessary for just four methods.

jhoanmanuelms (approved):

PR #314, opened 2020-12-11
Pivotal ID # 172671360: Ignore Empty Value Attributes

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/172671360

- Create extension to filter out attributes with no value
- Update schema to avoid null attribute values on the database
- Remove unused code


Juan-EBI (approved):
LGTM

PR #315, opened 2020-12-15
Pivotal-176168841: Allow to check inactive user registration

Juan-EBI:
https://www.pivotaltracker.com/story/show/176168841

### DIFF
--- a/submission/submission-security/src/main/kotlin/ebi/ac/uk/security/service/SecurityService.kt
+++ b/submission/submission-security/src/main/kotlin/ebi/ac/uk/security/service/SecurityService.kt
@@ -6,6 +6,7 @@ import ac.uk.ebi.biostd.persistence.model.ext.activated
 import ac.uk.ebi.biostd.persistence.repositories.UserDataRepository
 import arrow.core.getOrElse
 import ebi.ac.uk.api.security.ChangePasswordRequest
+import ebi.ac.uk.api.security.CheckUserRequest
 import ebi.ac.uk.api.security.GetOrRegisterUserRequest
 import ebi.ac.uk.api.security.LoginRequest
 import ebi.ac.uk.api.security.RegisterRequest
@@ -75,6 +76,10 @@ class SecurityService(
                 getOrCreateInactive(request.userEmail, userName)
             }
         }
+    }
+
+    override fun checkUserRegistration(register: CheckUserRequest): SecurityUser {
+        return getOrCreateInactive(register.userEmail, register.userEmail)
     }
 
     private fun createUserInactive(email: String, username: String): DbUser {
--- a/submission/submission-security/src/main/kotlin/ebi/ac/uk/security/service/SecurityService.kt
+++ b/submission/submission-security/src/main/kotlin/ebi/ac/uk/security/service/SecurityService.kt
@@ -79,7 +79,7 @@ class SecurityService(
     }
 
     override fun checkUserRegistration(register: CheckUserRequest): SecurityUser {
-        return getOrCreateInactive(register.userEmail, register.userEmail)
+        return getOrCreateInactive(register.userEmail, register.userName)
     }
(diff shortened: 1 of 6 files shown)


submission/submission-security/src/main/kotlin/ebi/ac/uk/security/service/SecurityService.kt
```diff
@@ -77,6 +78,10 @@ class SecurityService(
         }
     }
 
+    override fun checkUserRegistration(register: CheckUserRequest): SecurityUser {
+        return getOrCreateInactive(register.userEmail, register.userEmail)
```
jhoanmanuelms, line 82: The second parameter should be the name

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/security/web/SecurityResource.kt
```diff
@@ -45,6 +46,15 @@ class SecurityResource(
         securityService.registerUser(register)
     }
 
+    @PostMapping(value = ["/check-user"])
+    @ApiOperation("Check user registration")
```
jhoanmanuelms, line 50: I think that if we're going with this approach, the documentation should be better something like:

```
Checks if a user with the given email is registered in the system. In case it's not, a new user will be created using such email and the given user name. The new user will be inactive and should be activated via activation link.
```

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/security/web/SecurityResource.kt
```diff
@@ -45,6 +46,15 @@ class SecurityResource(
         securityService.registerUser(register)
     }
 
+    @PostMapping(value = ["/check-user"])
```
jhoanmanuelms, line 49: I think this name for this particular endpoint is very misleading. We should go for something like `check-register` or `search-register`; something that indicates that this endpoint actually performs two operations.

Said this, I'm not sure about how REST compliant this is, maybe we need just to have a check endpoint and let them decide whether or not to use the already existing register endpoint.

Besides, the actual registration endpoint has a lot of info that this one doesn't like the password, whether or not the notifications are enabled, etc. How will this be handled later for the user? I'm just worried this could become a workaround for our actual registration process and we should be careful now that our API is having external users.

image

jhoanmanuelms (commented):
I left some comments.

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/security/web/SecurityResource.kt
Juan-EBI: hahaha , the issue with the registration one is that is used for humans this is only for machine integration that need to register users. As array express or PMC so as we do not want to for example allow then to provide i.e password or other information that they do not have this is the endpoint for that :).


I agree that `check-register`is a much better name

jhoanmanuelms: Got it

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/security/web/SecurityResource.kt
```diff
@@ -45,6 +46,17 @@ class SecurityResource(
         securityService.registerUser(register)
     }
 
+    @PostMapping(value = ["/check-registration"])
+    @ApiOperation("Checks if a user with the given email is registered in the system. " +
```
jhoanmanuelms, line 50: You could use """ in order to avoid concatenation

jhoanmanuelms (approved):
LGTM. Please consider my comment.

PR #323, opened 2021-01-20
Pivotal #ID 176479473:  Migrate biostudies test to TestContainers

Juan-EBI:
https://www.pivotaltracker.com/story/show/176479473

### DIFF
--- a/submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/integration/MongoDbReposConfig.kt
+++ b/submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/integration/MongoDbReposConfig.kt
@@ -14,7 +14,7 @@ import org.springframework.data.mongodb.core.MongoTemplate
 
 @Configuration
 @Import(MongoDbConfig::class)
-@ConditionalOnProperty(prefix = "app.persistence", name = ["enabledMongo"])
+@ConditionalOnProperty(prefix = "app.persistence", name = ["enabledMongo"], havingValue = "true")
 class MongoDbReposConfig {
 
     @Bean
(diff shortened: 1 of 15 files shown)


submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/integration/MongoDbReposConfig.kt
```diff
@@ -0,0 +1,33 @@
...
+import ac.uk.ebi.biostd.persistence.doc.service.SubmissionMongoMetaQueryService
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+import org.springframework.context.annotation.Import
+import org.springframework.data.mongodb.core.MongoTemplate
+
+@Configuration
+@Import(MongoDbConfig::class)
+@ConditionalOnProperty(prefix = "app.persistence", name = ["enabledMongo"])
```
jhoanmanuelms, line 17: As I suggested in the previous PR. The property name should be either `enableMongo` or `mongoEnabled`.

Also, we need to update the file `application-local.yml` with the new properties.

submission/persistence-mongo/src/test/kotlin/ac/uk/ebi/biostd/persistence/doc/db/data/SubmissionDocDataRepositoryTest.kt
```diff
@@ -0,0 +1,117 @@
...
+    @Autowired
+    lateinit var submissionMongoRepository: SubmissionMongoRepository
+
+    @Nested
+    inner class UpdateStatus {
+        @Test
+        fun `successful status update`() {
+            submissionMongoRepository.save(testDocSubmission("accNo1", 1, PROCESSING))
+
+            testInstance.updateStatus(PROCESSED, "accNo1", 1)
```
jhoanmanuelms, line 43: I think using the named `version` parameter would add more clarity to the tests

jhoanmanuelms (commented):
I left some comments

submission/persistence-mongo/src/test/kotlin/ac/uk/ebi/biostd/persistence/doc/db/data/SubmissionDocDataRepositoryTest.kt
Juan-EBI: this is really sugetive one may same the same about status and processes status not sure what convention of logic conclusion we can arrive to add parameters names or not

jhoanmanuelms (approved):
LGTM

PR #325, opened 2021-01-26
Pivotal ID # 176479473: Migrate biostudies test to TestContainers

Juan-EBI:
https://www.pivotaltracker.com/story/show/176479473


submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/common/BaseIntegrationTest.kt
```diff
@@ -4,28 +4,40 @@ import ac.uk.ebi.biostd.client.integration.web.BioWebClient
 import ac.uk.ebi.biostd.client.integration.web.SecurityWebClient
 import ac.uk.ebi.biostd.itest.entities.TestUser
 import io.github.glytching.junit.extension.folder.TemporaryFolder
+import org.junit.jupiter.api.AfterAll
 import org.junit.jupiter.api.BeforeAll
+import org.testcontainers.containers.MySQLContainer
+
+private const val CHARACTER_SET = "utf8mb4"
+private const val COLLATION = "utf8mb4_unicode_ci"
+private const val MYSQL_IMAGE = "mysql:5.7.33"
```
jhoanmanuelms, line 13: Maybe we should have these versions (mongo, mysql) in a constants file. I wonder if we can reference the versions defined [here](https://github.com/EBIBioStudies/biostudies-backend-services/blob/master/buildSrc/src/main/kotlin/Dependencies.kt#L59) from the code?

I made this same suggestion in another PR

jhoanmanuelms (approved):
LGTM. It's gonna be great to have these tests running over an actual MySQL instance! Please take a look at my question.

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/common/BaseIntegrationTest.kt
Juan-EBI: no we can not, that is only grade I guess we can put then in test constants module. I will test this in next PR but do not want to change dependencies between the projects now.

PR #326, opened 2021-01-26
Pivotal #ID: 176670057 Modification date needs to be changed when submission is deleted

Juan-EBI:
https://www.pivotaltracker.com/story/show/176670057
- added logic to update modification date on deletion

### DIFF
--- a/submission/persistence-sql/src/main/kotlin/ac/uk/ebi/biostd/persistence/repositories/data/SubmissionRepository.kt
+++ b/submission/persistence-sql/src/main/kotlin/ac/uk/ebi/biostd/persistence/repositories/data/SubmissionRepository.kt
@@ -25,6 +25,7 @@ import org.springframework.data.domain.Sort
 import org.springframework.data.domain.Sort.Order
 import org.springframework.transaction.annotation.Transactional
 import uk.ac.ebi.extended.serialization.service.ExtSerializationService
+import java.time.OffsetDateTime
 
 private val logger = KotlinLogging.logger {}
 private val defaultOrder = Order.asc("id")
@@ -52,6 +53,7 @@ internal open class SubmissionRepository(
         val submission = submissionRepository.findByAccNoAndVersionGreaterThan(accNo)
         if (submission != null) {
             submission.version = -submission.version
+            submission.modificationTime = OffsetDateTime.now()
             submissionRepository.save(submission)
         }
     }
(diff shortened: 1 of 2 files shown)


submission/persistence-sql/src/main/kotlin/ac/uk/ebi/biostd/persistence/repositories/data/SubmissionRepository.kt
```diff
@@ -52,6 +53,7 @@ internal open class SubmissionRepository(
         val submission = submissionRepository.findByAccNoAndVersionGreaterThan(accNo)
         if (submission != null) {
             submission.version = -submission.version
+            submission.modificationTime = OffsetDateTime.now()
```
jhoanmanuelms, line 56: 👍

submission/persistence-sql/src/test/kotlin/ac/uk/ebi/biostd/persistence/repositories/data/SubmissionRepositoryTest.kt
```diff
@@ -0,0 +1,65 @@
...
+import io.mockk.verify
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.ExtendWith
+import uk.ac.ebi.extended.serialization.service.ExtSerializationService
+import java.time.OffsetDateTime
+import java.time.ZoneOffset
+
+@ExtendWith(MockKExtension::class)
+class SubmissionRepositoryTest(
```
jhoanmanuelms, line 23: Should we also check in the integrations tests that the date is properly modified?

https://github.com/EBIBioStudies/biostudies-backend-services/blob/master/submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/security/DeletePermissionTest.kt#L66

jhoanmanuelms (approved):
LGTM. Please check my question before merging.

submission/persistence-sql/src/test/kotlin/ac/uk/ebi/biostd/persistence/repositories/data/SubmissionRepositoryTest.kt
Juan-EBI: not sure to be honest, more test is usually never bad but actually such going such low level IT can increase its complexity and make it harder to identify code business conditions.  Modification date is not even publised to users.

PR #331, opened 2021-02-01
Pivotal ID # 175163839: Mongo Submission Stats

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/176479720

- Implementation of the service to persist and update submission stats in Mongo
- Scripts to generate a Mongo docker container for BioStudies
- Add missing properties to the Submission Web App configuration
- Add missing properties to the local instance configuration
- Remove unnecessary step in the builder configuration


ci/submission-handlers.yml
```diff
@@ -17,13 +17,6 @@ deploy-dev-handlers:
     - build-test
   script: gradle deployHandlers -Penv=dev -PdeployPath=/nfs/biostudies/.adm/apps/handlers/dev
 
-deploy-dev-handlers:
```
Juan-EBI, line 20: why we are removing this module deployment?

submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/service/StatsMongoDataService.kt
```diff
@@ -0,0 +1,66 @@
...
+            .map { toSubmissionStat(it, submissionStatType) }
+
+    override fun findByAccNoAndType(accNo: String, submissionStatType: SubmissionStatType): SubmissionStat =
+        subDataRepository
+            .findByAccNoAndStatType(accNo, submissionStatType)
+            ?.let { toSubmissionStat(it, submissionStatType) }
+            ?: throw StatNotFoundException(accNo, submissionStatType)
+
+    override fun saveAll(stats: List<SubmissionStat>): List<SubmissionStat> =
+        stats.filterValid()
```
Juan-EBI, line 30: not following why we need to filter valid this looks to me like and accept a request and ignore errors, I think the correct thing to do is that an API fail if input is invalid.

submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/service/StatsMongoDataService.kt
```diff
@@ -0,0 +1,66 @@
...
+        subDataRepository.updateStats(sub.accNo, sub.version, updatedStats(update, sub.stats))
+        return update
+    }
+
+    private fun updatedStats(update: SubmissionStat, currentStats: List<DocStat>): List<DocStat> =
+        currentStats
+            .filterNot { it.name == update.type.name }
+            .plus(DocStat(update.type.name, update.value))
+
+    private fun summarizeIncrements(currentStats: List<DocStat>, increments: List<SubmissionStat>): SubmissionStat {
```
Juan-EBI, line 54: not following what is the purpose of this function. Name/code is not self-explanatory I suspect is because is not single purposed. May you explain the logic, please.

submission/submission-webapp/src/main/resources/application-local.yml
```diff
@@ -36,3 +41,6 @@ app:
     magicDirPath: # Absolute path to the folder to be used for the user/groups magic folders links
     environment: LOCAL
     requireActivation: false
+  persistence:
+    enableMongo: false
+    connection: mongodb://root:admin@localhost:27017
```
Juan-EBI, line 46: not following why we need this updates

submission/submission-webapp/src/main/resources/application-local.yml
```diff
@@ -5,6 +5,11 @@ spring:
     password: admin
     driver-class-name: com.mysql.jdbc.Driver
 
+  data:
+    mongodb:
+      database: biostudies-dev
+      uri: mongodb://root:admin@localhost:27017
```
Juan-EBI, line 11: not following why we need this updates

ci/submission-handlers.yml
jhoanmanuelms: These lines are duplicated

submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/service/StatsMongoDataService.kt
jhoanmanuelms: We already discussed this with the frontend guys when this was developed and we agreed the API shouldn't fail in this case. Should we really change this behavior?

submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/service/StatsMongoDataService.kt
jhoanmanuelms: A file may contain several increments for the same submission, for example:

```
S-BSST1    8
S-BSST2    4
S-BSST1    2
```

The method should summarize all the increments for the same submission so at the end it's incremented the corresponding amount, 10 for S-BSST1, 4 for S-BSST2. The purpose of this method is to summarize this before and then make a single update.

submission/submission-webapp/src/main/resources/application-local.yml
jhoanmanuelms: The purpose of this file is to have a ready to go application.yml file for local environments. We already have the [connection string for MySQL there](https://github.com/EBIBioStudies/biostudies-backend-services/pull/331/files/396f85a421cb49b5b0d2219b3b6437d8f5839407#diff-3fdf6382a2e250b6405e23bc785966755aaca13cbe545636bd0d56da307704eaL3), I'm just adding the missing mongo part.

submission/submission-webapp/src/main/resources/application-local.yml
jhoanmanuelms: Same as above

submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/db/data/SubmissionDocDataRepository.kt
```diff
@@ -54,6 +58,12 @@ class SubmissionDocDataRepository(
     }
 
     fun getProjects(accNo: String): List<DocProject> = submissionRepository.getSubmissionProjects(accNo).projects
+
+    private fun updateSubmissionField(accNo: String, version: Int, field: String, value: Any) {
```
Juan-EBI, line 62: no quite sure about this, we save 5% of code that we use but at the same time there is an extra call so method is also 5% more complex. Decisions decisions...

```
        mongoTemplate.updateFirst<DocSubmission>(
            Query(where(SUB_ACC_NO).`is`(accNo).andOperator(where(SUB_VERSION).`is`(version))),
            update(SUB_STATUS, status))
 ```

File: submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/service/StatsMongoDataService.kt (hunk 2 above)
Comment (Juan-EBI) [reply submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/service/StatsMongoDataService.kt:30]: lets just reference TODO ticket to remove this hack and allow stats for expired versions

File: submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/service/StatsMongoDataService.kt (hunk 3 above)
Comment (Juan-EBI) [reply submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/service/StatsMongoDataService.kt:54]: I see I think we should perform this in the database so it thread-safe, simpler and efficient.

```
BulkOperations ops = template.bulkOps(BulkMode.UNORDERED, Match.class);
for (val stat : newStats) {
    ops.updateOne(query(where("acc").is(stat.acc).and("type").is(stats.type), createUpdateSet(stat));
}
ops.execute();
```

submission/submission-webapp/src/main/resources/application-local.yml
Juan-EBI: got it you are right

submission/submission-webapp/src/main/resources/application-local.yml
Juan-EBI: but here why we need this one, we use spring config so AFAIK this is not needed

jhoanmanuelms: Dunno, you added this property in the application.yml. I'm just adding the local value here

submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/db/data/SubmissionDocDataRepository.kt
jhoanmanuelms: Got it, I'll roll these back

submission/submission-webapp/src/main/resources/application-local.yml
jhoanmanuelms: UPDATE: Apparently, this property isn't necessary so I'll just remove it

submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/service/StatsMongoDataService.kt
jhoanmanuelms: Done

Juan-EBI (approved):
LGTM

PR #332, opened 2021-02-05
Pivotal ID # 176779952: Mongo iTests CI

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/176779952

- Use a Gradle property to decide whether to run the tests in Mongo or MySQL
- Run both types of integration tests in the CI
- Load the schema for the MySQL database container from a proper SQL script
- Remove unused database column

### DIFF
--- a/gradle/itest.gradle.kts
+++ b/gradle/itest.gradle.kts
@@ -9,7 +9,13 @@ sourceSets {
     }
 }
 
+val copySqlSchema = tasks.create<Copy>("copySqlSchema") {
+    from("$rootDir/infrastructure/src/main/resources/setup/database/Schema.sql")
+    into("$buildDir/resources/itest")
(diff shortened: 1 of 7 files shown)


Juan-EBI (approved):
LGTM

gradle/itest.gradle.kts
```diff
@@ -9,10 +9,20 @@ sourceSets {
     }
 }
 
+val copySqlSchema = tasks.create<Copy>("copySqlSchema") {
+    from("$rootDir/infrastructure/src/main/resources/setup/database/Schema.sql")
```
Juan-EBI, line 13: one thing here is that we need slightly different files for MySQL in mongo mode and MySQL in normal mode. We may fix this here or in another ticket.

PR #339, opened 2021-02-24
Pivotal ID #175894089: Check if queues in RabbitMQ are persistent

Juan-EBI:
https://www.pivotaltracker.com/story/show/175894089


jhoanmanuelms (approved):
LGTM

PR #350, opened 2021-03-13
Pivotal #ID 174642598: TSV Double Quotes

Juan-EBI:
https://www.pivotaltracker.com/story/show/174642598


commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/service/PageTabSerializationService.kt
```diff
@@ -33,7 +32,7 @@ internal class PageTabSerializationService(
         fileListSerializer.deserializeFileList(serializer.deserializeSubmission(content, format), source)
 
     override fun deserializeSubmission(file: File): Submission =
-        when (fromFile(file)) {
+        when (SubFormat.fromFile(file)) {
```
jhoanmanuelms, line 35: but-why

commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/tsv/deserialization/chunks/TsvChunkGenerator.kt
```diff
@@ -47,26 +50,23 @@ internal class TsvChunkGenerator(private val parser: CSVFormat = createParser())
...
 
     companion object {
-        private fun createParser(): CSVFormat {
-            return CSVFormat.DEFAULT
-                .withDelimiter(TAB)
-                .withIgnoreSurroundingSpaces()
-                .withIgnoreEmptyLines(false)
-                .withCommentMarker(TSV_COMMENT)
+        private fun createParser(): CsvParser {
+            val settings = CsvParserSettings()
```
jhoanmanuelms, line 64: This could be a one-liner if `apply` is used

commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/tsv/deserialization/common/Functions.kt
```diff
@@ -20,7 +19,6 @@ internal inline fun validate(value: Boolean, lazyMessage: () -> String) {
 internal fun toAttributes(chunkLines: List<TsvChunkLine>): MutableList<Attribute> {
     val attributes: MutableList<Attribute> = mutableListOf()
     chunkLines.forEach { line ->
-        line.value.ifBlank { throw InvalidElementException(REQUIRED_ATTR_VALUE) }
```
jhoanmanuelms, line 23: Why are we removing this validation?

commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/tsv/deserialization/chunks/TsvChunkGenerator.kt
```diff
@@ -47,26 +50,23 @@ internal class TsvChunkGenerator(private val parser: CSVFormat = createParser())
...
+            settings.format.setLineSeparator("\n")
```
jhoanmanuelms, line 65: Is this the line that fixes the bug? I didn't see any other difference compared to the previous config. If son, couldn't this be added to the existing config instead of changing the whole block? I think the builder approach is clearer

commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/service/PageTabSerializationService.kt
Juan-EBI: I founded  hard while coding to identify what was function doing it without the class name like figured out was that fromFile from so I added back, another option I think will be name function `SubFormatFromFile`

commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/tsv/deserialization/chunks/TsvChunkGenerator.kt
Juan-EBI: added documentation

jhoanmanuelms (approved):
LGTM

PR #351, opened 2021-03-14
Pivotal ID# 174948695: PTSubmit Delete Operation

Juan-EBI:
https://www.pivotaltracker.com/story/show/174948695


client/bio-commandline/src/main/kotlin/uk/ac/ebi/biostd/client/cli/BioStudiesCommandLine.kt
```diff
@@ -1,83 +1,17 @@
...
-        transformer.transform(xmlInput, xmlOutput)
-
-        return xmlOutput.writer.toString()
+        echo("That's BioStudies command line")
     }
 }
 
-fun main(args: Array<String>) = BioStudiesCommandLine().main(args)
+fun main(args: Array<String>) = BioStudiesCommandLine()
+    .subcommands(SubmitCommand(SubmissionService()), DeleteCommand(SubmissionService()))
```
jhoanmanuelms, line 16: Just out of curiosity, couldn't we define a single instance of `SubmissionService` and pass it down to both sub commands? Would this give us any advantage?

client/bio-commandline/src/main/kotlin/uk/ac/ebi/biostd/client/cli/services/SubmissionService.kt
```diff
@@ -0,0 +1,49 @@
...
+
+data class SubmissionDeleteRequest(
+    val server: String,
+    val user: String,
+    val password: String,
+    val onBehalf: String?,
+    val accNo: String
+)
+
+class SubmissionService {
```
jhoanmanuelms, line 27: Really nice refactor for decoupling bro 👏

jhoanmanuelms (commented):
All the changes look really nice. I think the refactor for decoupling you did is awesome. I only have one question: how would the user specify which operation is performed? I didn't see it in the tests.

I'm assuming it would be by specifying the operation name somewhere? Or is it based on the parameter list? If so, maybe we should add a new parameter, something like `-sb --submit`/`-dl --delete` because leaving it to the amount of parameters look very implicit.

Other than that I think the changes are great! 🤘

PR #352, opened 2021-03-16
Pivotal ID # 177267343: Validate EU-ToxRisk Submissions

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/177267343

- Implement collection validators
- Update the submission submitter to use them

Delivers # 177267343


commons/commons-model-extended-mapping/src/main/kotlin/ebi/ac/uk/extended/mapping/to/ToSubmission.kt
```diff
@@ -18,11 +22,14 @@ fun ExtSubmission.toSimpleSubmission(): Submission = Submission(
 )
 
 private fun ExtSubmission.getSubmissionAttributes(): List<Attribute> {
-    val subAttrs = attributes.map { it.toAttribute() }.toMutableSet()
+    val subAttrs =
```
Juan-EBI, line 25: why are we filtering it here?

jhoanmanuelms: https://github.com/EBIBioStudies/biostudies-backend-services/pull/352/files/7b1cd4da93596bcedb0b382c16ffa1886d527ecf#diff-9f69c64127e65171eccdde20f28b77b1545af6262592db23b1512fade38f528cL16

commons/commons-model-extended-mapping/src/main/kotlin/ebi/ac/uk/extended/mapping/to/ToSubmission.kt
```diff
@@ -18,11 +21,14 @@ fun ExtSubmission.toSimpleSubmission(): Submission = Submission(
...
+        attributes
```
Juan-EBI, line 25: I think we should create a private method `getpageTabAttributes`  to make clear why we remove the validator.

commons/commons-test/src/main/kotlin/ebi/ac/uk/test/ExtSubmissionFactory.kt
```diff
@@ -0,0 +1,29 @@
+package ebi.ac.uk.test
+
+import ebi.ac.uk.extended.model.ExtProcessingStatus.PROCESSED
+import ebi.ac.uk.extended.model.ExtSection
+import ebi.ac.uk.extended.model.ExtSubmission
+import ebi.ac.uk.extended.model.ExtSubmissionMethod.PAGE_TAB
+import java.time.OffsetDateTime
+import java.time.ZoneOffset.UTC
+
+val basicExtSubmission = ExtSubmission(
```
Juan-EBI, line 10: we have create many of this, eventually we will need to put in common and use the same one

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/service/ParentInfoService.kt
```diff
@@ -2,26 +2,49 @@ package ac.uk.ebi.biostd.submission.service
...
+    fun executeCollectionValidators(submission: ExtSubmission) = submission.isCollection.ifFalse {
+        submission
+            .collections
+            .mapNotNull { collectionInfo(it.accNo).validator }
+            .forEach { validate(it, submission) }
+    }
+
+    private fun collectionInfo(accNo: String) = queryService.getBasicCollection(accNo)
+
+    private fun validate(validator: String, submission: ExtSubmission) =
```
Juan-EBI, line 44: this method is not doing what is said that is does making logic harder to follow, it does not validate but it gets the validation bean and validate (it is also doing to two operations and not a single one)

submission/submitter/src/test/kotlin/ac/uk/ebi/biostd/submission/validator/collection/EuToxRiskValidatorTest.kt
```diff
@@ -0,0 +1,99 @@
...
+        val error = assertThrows<CollectionValidationException> { testInstance.validate(submission) }
+        assertThat(error.message).isEqualTo(
+            "The submission doesn't comply with the collection requirements. Errors: [$EXCEL_FILE_REQUIRED]")
+    }
+
+    private fun setUpRestTemplate(response: EuToxRiskValidatorResponse): CapturingSlot<HttpEntity<FileSystemResource>> {
+        val requestSlot = slot<HttpEntity<FileSystemResource>>()
+
+        every {
+            restTemplate.postForObject(testUrl, capture(requestSlot), EuToxRiskValidatorResponse::class.java)
```
Juan-EBI, line 94: I think we can use `.postForObject<EuToxRiskValidatorResponse` can not we? as in the normal code.

submission/submitter/src/test/kotlin/ac/uk/ebi/biostd/submission/validator/collection/EuToxRiskValidatorTest.kt
```diff
@@ -0,0 +1,99 @@
...
+            files = listOf(left(ExtFile("test.txt", textFile)))
+        ))
+
+        val error = assertThrows<CollectionValidationException> { testInstance.validate(submission) }
+        assertThat(error.message).isEqualTo(
+            "The submission doesn't comply with the collection requirements. Errors: [$EXCEL_FILE_REQUIRED]")
+    }
+
+    private fun setUpRestTemplate(response: EuToxRiskValidatorResponse): CapturingSlot<HttpEntity<FileSystemResource>> {
+        val requestSlot = slot<HttpEntity<FileSystemResource>>()
```
Juan-EBI, line 91: IMO we should declare this as a field and in each test:
```
  every {  restTemplate.postForObject<>(testUrl, capture(requestSlot)) } } returns response
```
I think way when one is reviewing the test does not need to check what `setUpRestTemplate` does as the setup is part of each test. Also I think we do not need type  I think compiler with infer it.

Juan-EBI (approved):
LGTM please check comments before merge

commons/commons-test/src/main/kotlin/ebi/ac/uk/test/ExtSubmissionFactory.kt
jhoanmanuelms: That's actually what I did with this change. I placed this one in commons-test and used it from everywhere else

PR #357, opened 2021-03-21
Pivotal ID #177429626: Bio WebClient errors should not be parsed as XML

Juan-EBI:
https://www.pivotaltracker.com/story/show/177429626


jhoanmanuelms (approved):
LGTM

PR #362, opened 2021-03-29
Pivotal ID # 177304040: PT Submit Migration Operation

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/177304040

Create an operation in PT submit to migrate a submission from one environment to another


Juan-EBI (approved):
LGTM

PR #363, opened 2021-03-30
Pivotal ID # 177304040: Fix Build

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/177304040

- Increase submitter RAM
- Update deployment artifact name for BioStudies CLI


Juan-EBI (approved):
LGTM

PR #366, opened 2021-04-07
Pivotal # 177647985: MD5 Calculation For Big Files

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/177647985

Switch from full read file for MD5 calculation to a buffered reader approach

### DIFF
--- a/commons/commons-util/src/main/kotlin/ebi/ac/uk/io/FileUtils.kt
+++ b/commons/commons-util/src/main/kotlin/ebi/ac/uk/io/FileUtils.kt
@@ -7,17 +7,22 @@ import ebi.ac.uk.io.FileUtilsHelper.createFolderIfNotExist
 import ebi.ac.uk.io.FileUtilsHelper.createParentDirectories
 import ebi.ac.uk.io.FileUtilsHelper.createSymLink
 import ebi.ac.uk.io.ext.notExist
-import org.apache.commons.codec.digest.DigestUtils
 import java.io.File
 import java.io.InputStream
+import java.math.BigInteger
 import java.nio.file.Files
 import java.nio.file.Files.exists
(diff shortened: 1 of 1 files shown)


commons/commons-util/src/main/kotlin/ebi/ac/uk/io/FileUtils.kt
```diff
@@ -123,14 +128,27 @@ object FileUtils {
...
+        val buffer = ByteArray(BUFFER_SIZE)
+        val digest = MessageDigest.getInstance(MD5_ALGORITHM)
+
+        while (inputStream.read(buffer).also { read = it } > 0) {
+            digest.update(buffer, 0, read)
+        }
+
+        val md5sum = BigInteger(CHECKSUM_SIGNUM, digest.digest()).toString(HEXADECIMAL_BASE)
+        return String.format("%32s", md5sum).replace(" ", "0")
+    }
```
Juan-EBI, line 151:
```
  fun md5(file: File): String {
        val digest = MessageDigest.getInstance(MD5_ALGORITHM)
        file.inputStream().buffered(BUFFER_SIZE).iterator().forEach { digest.update(it) }
        return digest.digest().joinToString("") { "%02x".format(it) }
    }
```

Maybe?

jhoanmanuelms: Works like a charm! I knew there was a better way to write this. Nice comment 😁👍

jhoanmanuelms: Fixed

Juan-EBI (approved):

PR #373, opened 2021-04-20
Pivotal ID # 176806007: Fix indexing of `gen` links 

Juan-EBI:
https://www.pivotaltracker.com/story/show/176806007

### DIFF
--- a/scheduler/pmc-processor-task/src/main/kotlin/ac/uk/ebi/pmc/load/PmcSubmissionTabProcessor.kt
+++ b/scheduler/pmc-processor-task/src/main/kotlin/ac/uk/ebi/pmc/load/PmcSubmissionTabProcessor.kt
@@ -1,0 +1,34 @@
+package ac.uk.ebi.pmc.load
+
+import ac.uk.ebi.biostd.integration.SerializationService
+import ac.uk.ebi.biostd.integration.SubFormat
+import arrow.core.Either
+import ebi.ac.uk.model.Link
+import ebi.ac.uk.model.LinksTable
+import ebi.ac.uk.model.Submission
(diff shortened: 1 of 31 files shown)


scheduler/pmc-processor-task/src/main/kotlin/ac/uk/ebi/pmc/load/PmcSubmissionTabProcessor.kt
```diff
@@ -0,0 +1,34 @@
...
+        val submission = serializationService.deserializeSubmission(originalPagetab, SubFormat.TSV)
+        submission.allSections().forEach { it.links = it.links.map { either -> updateLinks(either) }.toMutableList() }
+        return submission
+    }
+
+    private fun updateLinks(either: Either<Link, LinksTable>) =
+        either.bimap(this::transformLink, this::transformLinksTable)
+
+    private fun transformLink(links: Link): Link {
+        links.attributes.filter { it.value == "gen" }.forEach { it.value = "ENA" }
```
jhoanmanuelms, line 26: I'm wondering why do we have to make this change here since EuropePMC is actually producing the page tab, shouldn't they send the correct value? 🤔

jhoanmanuelms (approved):
LGTM. Just one thought

scheduler/pmc-processor-task/src/main/kotlin/ac/uk/ebi/pmc/load/PmcSubmissionTabProcessor.kt
awaisathar: For legacy reasons, they need to keep `gen` at their end. They are looking into if they can change it while exporting but it would take time.

PR #374, opened 2021-04-27
Pivotal ID # 172336299: Remove Access Tags From Public Export

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/172336299

- Create the exporter service to export the public studies
- Add utilities to trigger it daily or by demand


Juan-EBI (approved):
LGTM but I think the deployment script is missing, are you adding it as part of other ticket?

jhoanmanuelms:
Nice catch. I updated the deployment

PR #376, opened 2021-04-30
Pivotal ID#  177979104:Issues submitting S-BIAD116

Juan-EBI:
https://www.pivotaltracker.com/story/show/177979104

- added custom serializers
- added support to reloading after request processing to guarantee the last version of the information available is stored.

### DIFF
--- a/submission/persistence-sql/src/main/kotlin/ac/uk/ebi/biostd/persistence/mapping/extended/from/ToDbSubmissionMapper.kt
+++ b/submission/persistence-sql/src/main/kotlin/ac/uk/ebi/biostd/persistence/mapping/extended/from/ToDbSubmissionMapper.kt
@@ -23,7 +23,7 @@ internal class ToDbSubmissionMapper(
     private var userRepository: UserDataRepository
 ) {
 
-    fun toSubmissionDb(submission: ExtSubmission) = toSubmissionDb(submission, DbSubmission())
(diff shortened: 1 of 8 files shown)


submission/persistence-sql/src/main/kotlin/ac/uk/ebi/biostd/persistence/mapping/extended/from/ToDbSubmissionMapper.kt
```diff
@@ -22,25 +22,30 @@ internal class ToDbSubmissionMapper(
...
-        submitter = getUser(submission.submitter)
-        accessTags = toAccessTag(submission.collections)
-        tags = toTags(submission.tags)
-        released = submission.released
-        attributes = submission.attributes.mapIndexedTo(sortedSetOf(), ::toDbSubmissionAttribute)
-        rootSection = submission.section.toDbSection(this, ROOT_SECTION_ORDER)
+
+    fun toSubmissionDb(submission: ExtSubmission) = toSubmissionDb(submission, DbSubmission())
+
+    fun toSubmissionDb(submission: ExtSubmission, dbSubmission: DbSubmission): DbSubmission {
```
jhoanmanuelms, line 28: I think it would be clearer is this fucntion is an extension like  `DbSubmission.toUpdatedSubmission(submission: ExtSubmission)` because, from the code, it seems weird to pass the current to a mapping method

jhoanmanuelms (approved):

PR #379, opened 2021-05-05
Pivotal #ID: 176069911: PTSubmit Async Support

Juan-EBI:
https://www.pivotaltracker.com/story/show/176069911


client/bio-commandline/src/main/kotlin/uk/ac/ebi/biostd/client/cli/common/Parameters.kt
```diff
@@ -19,3 +23,6 @@ internal object MigrationParameters {
     const val TARGET_USER = "BioStudies user in the target environment"
     const val TARGET_PASSWORD = "Password for the BioStudies user in the target environment"
 }
+
+fun getFiles(file: File): List<File> =
```
jhoanmanuelms, line 27: I don't think this file is the right place for this function to exist. I think it should be on a utility object

jhoanmanuelms (approved):

PR #384, opened 2021-05-15
Pivotal ID # 178166470: Upgrade Kotlin and Kotlin corotuines

Juan-EBI:
https://www.pivotaltracker.com/story/show/178166470

### DIFF
--- a/commons/commons-bio/src/main/kotlin/ebi/ac/uk/api/security/RegisterRequest.kt
+++ b/commons/commons-bio/src/main/kotlin/ebi/ac/uk/api/security/RegisterRequest.kt
@@ -3,6 +3,7 @@ package ebi.ac.uk.api.security
 import com.fasterxml.jackson.annotation.JsonProperty
 import javax.validation.constraints.Email
 
+@Suppress("LongParameterList")
 class RegisterRequest(
     val name: String,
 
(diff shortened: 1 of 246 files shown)


commons/commons-bio/src/main/kotlin/ebi/ac/uk/api/security/RegisterRequest.kt
```diff
@@ -3,6 +3,7 @@ package ebi.ac.uk.api.security
 import com.fasterxml.jackson.annotation.JsonProperty
 import javax.validation.constraints.Email
 
+@Suppress("LongParameterList")
```
jhoanmanuelms, line 6: I see a lot of these were added during this refactor. Shouldn't we just adjust the parameter to something that makes more sense to us?

gradle.properties
```diff
@@ -1 +1,2 @@
 org.gradle.parallel=false
+org.gradle.jvmargs=-Xmx4096m
```
jhoanmanuelms, line 2: Missing new end of file

jhoanmanuelms (approved):
LGTM. It looks like most of the changes were just for formatting however, I left a couple of comments, please take a look.

commons/commons-bio/src/main/kotlin/ebi/ac/uk/api/security/RegisterRequest.kt
Juan-EBI: lets discuss this and address in individual version

PR #390, opened 2021-05-27
Pivotal ID # 178315201: Migrate Collections

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/178315201

A collection shouldn't be required to exist before it's created


Juan-EBI (approved):
LGTM

PR #394, opened 2021-06-04
Pivotal #ID 177042710: Support for deleting multiple submissions at once

Juan-EBI:
https://www.pivotaltracker.com/story/show/177042710


submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/db/data/SubmissionDocDataRepository.kt
```diff
@@ -86,13 +88,17 @@ class SubmissionDocDataRepository(
...
+            DocSubmission::class.java
         )
+    }
 
+    fun expireVersions(submissions: List<String>) {
         mongoTemplate.updateMulti(
-            Query(criteria),
-            ExtendedUpdate().multiply(SUB_VERSION, -1),
+            Query(where(SUB_ACC_NO).`in`(submissions).andOperator(where(SUB_VERSION).gt(0))),
+            ExtendedUpdate().multiply(SUB_VERSION, -1).set(SUB_MODIFICATION_TIME, Instant.now()),
```
jhoanmanuelms, line 101: Why are you changing the modification time here but not in the single expire?

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/common/BaseIntegrationTest.kt
```diff
@@ -26,7 +26,7 @@ internal open class BaseIntegrationTest(private val tempFolder: TemporaryFolder)
 
     @BeforeAll
     fun beforeAll() {
-        if (System.getProperty("itest.mode") == "mongo") {
+        if (System.getProperty("testing.mode") == "mongo") {
```
jhoanmanuelms, line 29: This property shouldn't be changed, the name is defined here:

https://github.com/EBIBioStudies/biostudies-backend-services/blob/master/submission/submission-webapp/gradle.properties#L2

And it's referenced through several scripts and commands. Please revert this change.

jhoanmanuelms (changes requested):

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/web/resources/SubmissionResource.kt
```diff
@@ -33,6 +34,12 @@ class SubmissionResource(
         @PathVariable accNo: String
     ): Unit = submissionsWebHandler.deleteSubmission(accNo, user)
 
+    @DeleteMapping
+    fun deleteSubmissions(
```
jhoanmanuelms, line 38: Please add the corresponding documentation for the new endpoint, i.e.:

https://github.com/EBIBioStudies/biostudies-backend-services/blob/master/docs/api/BioStudiesPublicAPI.yaml#L61

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/common/BaseIntegrationTest.kt
Juan-EBI: gradle propery is call `itest.mode` but gradle property and system properties are not the same if you check gradle tasks which transform this property into a system property you can notice is call `testing.mode`, do I am missing something?
```
val itest = tasks.create<Test>("itest") {
    dependsOn(copySqlSchema)
    testClassesDirs = sourceSets["itest"].output.classesDirs
    classpath = sourceSets["itest"].runtimeClasspath

    val testingMode = project.property("itest.mode")
    println("##### Running integration tests in $testingMode mode #######")
    systemProperty("testing.mode", testingMode)

    useJUnitPlatform()
    extensions.configure(JacocoTaskExtension::class) {
        setDestinationFile(file("$buildDir/jacoco/jacocoITest.exec"))
        classDumpDir = file("$buildDir/jacoco/classpathdumps")
    }
}
```

jhoanmanuelms: Good catch man, then I think we should change to `itest.mode` [here](https://github.com/EBIBioStudies/biostudies-backend-services/blame/master/gradle/itest.gradle.kts#L24) instead of changing the `BaseIntegrationTest` class.

This typo was here from the very beginning, I'm afraid the tests were never running on mysql then 🤦‍♂️

jhoanmanuelms (approved):
LGTM

PR #395, opened 2021-06-04
Pivotal #ID 178411427: Keyword searchs need to check root section title

Juan-EBI:
https://www.pivotaltracker.com/story/show/178411427


submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/db/data/SubmissionDocDataRepository.kt
```diff
@@ -141,9 +144,16 @@ class SubmissionDocDataRepository(
                 filter.type?.let { add(where("$SUB_SECTION.$SEC_TYPE").`is`(it)) }
                 filter.rTimeFrom?.let { add(where(SUB_RELEASE_TIME).gte(it.toInstant())) }
                 filter.rTimeTo?.let { add(where(SUB_RELEASE_TIME).lte(it.toInstant())) }
-                filter.keywords?.let { add(where(SUB_TITLE).regex("(?i).*$it.*")) }
+                filter.keywords?.let { add(keywordsCriteria(it)) }
                 filter.released?.let { add(where(SUB_RELEASED).`is`(it)) }
             }.build().toTypedArray()
+
+        private fun keywordsCriteria(keywords: String) = Criteria().orOperator(
```
jhoanmanuelms, line 151: I was pretty sure we had an integration test for this and I was wondering how was it passing in mongo so I took a look:

https://github.com/EBIBioStudies/biostudies-backend-services/blob/master/submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/query/SubmissionListApiTest.kt#L153

We have the right case but we're making the wrong test, searching by accNo instead of keyword 🤦‍♂️. Could you please take a look?

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/query/SubmissionListApiTest.kt
```diff
@@ -151,58 +151,82 @@ internal class SubmissionListApiTest(private val tempFolder: TemporaryFolder) :
...
+                line("Title", "Submission With Section")
                 line()
             }.toString()
 
             assertThat(webClient.submitSingle(submission, TSV)).isSuccessful()
 
             val submissionList = webClient.getSubmissions(
                 mapOf(
-                    "accNo" to "SECT-123"
+                    "keywords" to "subTitle"
```
jhoanmanuelms, line 169: I think we should include spaces, i.e. "Section Title". It seems like a case the users might need

Juan-EBI: https://github.com/EBIBioStudies/biostudies-backend-services/pull/395/files#diff-57ceb60d5cf13b0bf3e3bd1a58158ba778820910103525101b7cf15ff92630f0R206

PR #396, opened 2021-06-07
Pivotal ID # 178213145: Activate ArrayExpress inactive users

jhoanmanuelms:
- Create an endpoint to activate users by email
- Improve response given by the "check-registration" endpoint


Juan-EBI (approved):
LGTM

PR #401, opened 2021-06-25
Pivotal ID # 177839764: FIRE FileProcessingService

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/177839764

- Refactor file processing services into interfaces
- Make implementations for nfs/fire files
- Rename the persistence-filesystem module in order to be more aligned with the actual module functionality
- Improve some packages and classes names


submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/api/FtpService.kt
```diff
@@ -0,0 +1,9 @@
+package ac.uk.ebi.biostd.persistence.filesystem.api
+
+import ebi.ac.uk.extended.model.ExtSubmission
+
+interface FtpService {
+    fun processSubmissionFiles(submission: ExtSubmission)
+
+    fun createFtpFolder(relPath: String)
```
Juan-EBI, line 8: I think this operation should receive `submission: ExtSubmission` to keep abstractions at the same level.

submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/service/FileProcessingService.kt
```diff
@@ -0,0 +1,60 @@
...
+import ebi.ac.uk.extended.model.ExtSubmission
+import ebi.ac.uk.extended.model.FileMode.COPY
+import ebi.ac.uk.extended.model.FireFile
+import ebi.ac.uk.extended.model.NfsFile
+
+class FileProcessingService {
+    fun processFiles(processRequest: FileProcessingRequest): ExtSubmission {
+        val (mode, submission, config) = processRequest
+
+        fun processFile(file: ExtFile): ExtFile = when (file) {
```
Juan-EBI, line 23: not clear for me why we are handling both scenarios here, this class is used by `NfsFilesService`  which only receive NFs request. I think we should not share `FileProcessingService`  for both `FilesService` implementation

Juan-EBI (approved):
LGTM please check comment before merge

submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/api/FtpService.kt
jhoanmanuelms: There's an endpoint that uses the submission's relPath in order to force the ftp link creation. This was kept like that to keep compatibility with that endpoint.

https://github.com/EBIBioStudies/biostudies-backend-services/blob/master/submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/web/resources/FtpResource.kt#L17

Otherwise, we would need to change the signature.

submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/service/FileProcessingService.kt
jhoanmanuelms: Nice catch. Fixed

PR #405, opened 2021-06-29
Pivotal ID # 177839764: FIRE FileProcessingService

jhoanmanuelms:
Fix build


jhoanmanuelms:
This change was also included in another PR

PR #416, opened 2021-07-23
Pivotal ID # 178974842: Section Title Index

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/178974842

Add the section title to the text search index for both, submissions and submission requests


submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/migrations/DatabaseChangeLog.kt
```diff
@@ -34,20 +35,32 @@ class DatabaseChangeLog {
...
+            ensureIndex(Index().on("$SUB.$SUB_ACC_NO", ASC))
+            ensureIndex(Index().on("$SUB.$SUB_OWNER", ASC))
+            ensureIndex(Index().on("$SUB.$SUB_SUBMITTER", ASC))
+            ensureIndex(Index().on("$SUB.$SUB_RELEASE_TIME", ASC))
+            ensureIndex(submissionRequestTitleIndex())
+            ensureIndex(Index().on("$SUB.$SUB_RELEASED", ASC))
         }
     }
+
+    private fun submissionTitleIndex() =
```
Juan-EBI, line 55: 1. why we are adding a textIndex on both submission title and attributes values?
2. We need to add new migration not update existing one. as they will not be executed

Juan-EBI (approved):
LGTM

PR #417, opened 2021-07-27
Pivotal ID # 179024679: Fix RabbitMQ version to 3.8

liviuba:
https://www.pivotaltracker.com/story/show/179024679

* Pin RabbitMQ container to 3.8-management (already used)


jhoanmanuelms (approved):
LGTM

PR #430, opened 2021-08-12
Pivotal ID # 179055347: Update AE Activation Email

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/179055347

Move ST instance keys from the code base to security properties

### DIFF
--- a/submission/submission-security/src/main/kotlin/ebi/ac/uk/security/util/SecurityUtil.kt
+++ b/submission/submission-security/src/main/kotlin/ebi/ac/uk/security/util/SecurityUtil.kt
@@ -1,5 +1,6 @@
 package ebi.ac.uk.security.util
 
+import ac.uk.ebi.biostd.common.properties.InstanceKeys
 import ac.uk.ebi.biostd.persistence.model.DbUser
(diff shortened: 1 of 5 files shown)


submission/submission-security/src/main/kotlin/ebi/ac/uk/security/util/SecurityUtil.kt
```diff
@@ -17,10 +18,6 @@ import java.time.Clock
 import java.time.OffsetDateTime
 import java.util.UUID
 
-internal const val DEV_KEY = "975dd2ca-58eb-407b-ba0f-858f15f7304d"
-internal const val BETA_KEY = "9c584ae3-678a-4462-b685-54c37a1bc047"
-internal const val PROD_KEY = "01ecc118-dbec-4df8-8fe8-f5cd7364b2b7"
-
 internal const val DEV_INSTANCE = "http://ves-hx-f2.ebi.ac.uk:8120"
```
awaisathar, line 21: We should move this to the config (in this ticket or a separate one). Will need that for migration to k8s.

Juan-EBI (approved):

PR #439, opened 2021-09-06
Pivotal ID # 179389139: FIRE Directories Persistence

jhoanmanuelms:
- Create new type for the FIRE directories
- Skip them during file processing on FIRE mode and persist them just in the database instead
- Adjust the extended serializers


Juan-EBI (commented):
LGTM

Juan-EBI (approved):

PR #446, opened 2021-09-27
Pivotal ID #179728378: On behalf register user improvements

Juan-EBI:
https://www.pivotaltracker.com/story/show/179728378


commons/commons-bio/src/main/kotlin/ebi/ac/uk/api/Parameters.kt
```diff
@@ -1,5 +1,5 @@
 package ebi.ac.uk.api
 
 const val ON_BEHALF_PARAM = "onBehalf"
-const val REGISTER_PARAM = "sse"
+const val REGISTER_PARAM = "register"
```
jhoanmanuelms, line 4: Good one. I always thought this name was kind of confusing. Could you please update the parameter description in the docs?

https://github.com/EBIBioStudies/biostudies-backend-services/tree/master/docs/api/submissions/operations

jhoanmanuelms (approved):
LGTM. Please check my comment before merging

PR #448, opened 2021-09-30
Pivotal ID # 179513930: Add Warning For Empty File Lists

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/179513930

- Change the excel reader from class to object
- Use a single method to read page tab from a file
- Add exception for empty page tab files


commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/extension/PageTabFileExt.kt
```diff
@@ -0,0 +1,14 @@
...
+
+import ac.uk.ebi.biostd.exception.EmptyPageTabFileException
+import ac.uk.ebi.biostd.integration.SubFormat
+import ac.uk.ebi.biostd.integration.SubFormat.TsvFormat.XlsxTsv
+import ebi.ac.uk.io.ext.size
+import ebi.ac.uk.util.file.ExcelReader.readContentAsTsv
+import java.io.File
+
+internal fun File.readAsPageTab(): String {
+    require(size() > 0) { throw EmptyPageTabFileException(name) }
```
Juan-EBI, line 11: I am not sure this is the right place to put this validation IMO serializers should fail if empty content. Meaning xml, json and tsv need to fail if you try to get a submission or file list from empty file. I think that is the bug. Here  it does work but our serialization layer still allow this so if someone call it directly (like the clients) it will still serialize.)

Juan-EBI (approved):
Please consider comments before merge

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/web/exception/ValidationExceptionHandler.kt
```diff
@@ -28,4 +29,12 @@ class ValidationExceptionHandler {
 
         return ValidationTree(FAIL, node)
     }
+
+    @ResponseBody
+    @ResponseStatus(HttpStatus.BAD_REQUEST)
+    @ExceptionHandler(EmptyPageTabFileException::class)
+    fun handleRuntime(exception: EmptyPageTabFileException): ValidationTree {
+        exception.printStackTrace()
```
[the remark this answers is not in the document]
jhoanmanuelms: Yes, the change is for the response status. This should be a bad request not an internal server error.

commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/extension/PageTabFileExt.kt
jhoanmanuelms: Serializers and readers both will fail now with empty content, I'm just trying to make a centralized error for when the user sends an empty file.

PR #449, opened 2021-10-06
Pivotal ID #179834295: Excel dsl library

Juan-EBI:
https://www.pivotaltracker.com/story/show/179834295


commons/commons-util/src/test/kotlin/ebi/ac/uk/util/file/ExcelReaderTest.kt
```diff
@@ -42,7 +42,7 @@ class ExcelReaderTest(private val temporaryFolder: TemporaryFolder) {
                     cell("123")
                 }
 
-                emptyRow()
```
Juan-EBI, line 45: @jhoanmanuelms I remove emptyRow as this is a page tab concept and not related to dsl. I mean an empty two row is not defiened as as row with two other cells in other context that is not pagetab.

PR #450, opened 2021-10-06
Pivotal ID # 179784398: Improve all in one itest to include file list

Juan-EBI:
https://www.pivotaltracker.com/story/show/179784398


commons/commons-test/src/main/kotlin/ebi/ac/uk/test/TemporaryFolderExt.kt
```diff
@@ -23,6 +23,15 @@ fun TemporaryFolder.createOrReplaceFile(fileName: String): File {
     return createFile(fileName)
 }
 
+/**
+ * Creates a file with the given name and content or replaces it if already exist.
+ */
+fun TemporaryFolder.createOrReplaceFile(fileName: String, content: String): File {
```
jhoanmanuelms, line 29: Couldn't this lead to failures? Maybe it'd be better just cleaning the temp folder after each test if required.

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/assertions/AllInOneSubmissionHelper.kt
```diff
@@ -6,33 +6,85 @@ import ac.uk.ebi.biostd.itest.factory.assertAllInOneSubmissionTsv
...
         val extendedSubmission = submissionRepository.getExtByAccNo(accNo)
         assertThat(extendedSubmission.status).isEqualTo(ExtProcessingStatus.PROCESSED)
         assertThat(extendedSubmission.method).isEqualTo(method)
         assertThat(extendedSubmission.toSimpleSubmission()).isEqualTo(allInOneSubmission(accNo))
-        assertSubmissionFiles(extendedSubmission)
+        assertSubmissionFiles(extendedSubmission, checkTabFiles)
     }
 
-    private fun assertSubmissionFiles(submission: ExtSubmission) {
+    private fun assertSubmissionFiles(submission: ExtSubmission, checkTabFiles: Boolean) {
```
jhoanmanuelms, line 37: As a wise friend once told me, boolean flags in methods aren't usually a good idea. Maybe it's better to have separated methods or just a method for specifically checking the page tab files and call it whenever necessary.

Also, I'd suggest `checkPageTabFiles` to keep consistency.

jhoanmanuelms (approved):
LGTM. Just a couple of thoughts.

PR #453, opened 2021-10-11
Pivotal ID # 179436227: Improve Logs Messages 

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/179436227

- Adjust the submitter log messages format in order to be identified by Kibana
- Include the AccNo and Submitter in each message so they can be properly indexed
- Adjust tests


commons/commons-model-extended/src/main/kotlin/ebi/ac/uk/extended/events/SubmissionRequestMessage.kt
```diff
@@ -13,6 +13,9 @@ data class SubmissionRequestMessage(
     @JsonProperty("fileMode")
     val fileMode: FileMode,
 
+    @JsonProperty("owner")
+    val submitter: String,
```
Juan-EBI, line 17: I think this is consufusing, in submission system we have the concept of submitter and owner and here you are basically inverting it.

jhoanmanuelms: You're right, it should be the owner. Nice catch.

PR #456, opened 2021-10-13
 Pivotal ID # 179810581: Files files storage

Juan-EBI:
https://www.pivotaltracker.com/story/show/179810581

### DIFF
--- a/client/bio-commandline/src/test/kotlin/uk/ac/ebi/biostd/client/cli/services/SubmissionServiceMigrationTest.kt
+++ b/client/bio-commandline/src/test/kotlin/uk/ac/ebi/biostd/client/cli/services/SubmissionServiceMigrationTest.kt
@@ -97,7 +97,7 @@ class SubmissionServiceMigrationTest(
             type = "Study",
             fileList = ExtFileList(
                 "test-file-list",
-                filesUrl = "/submissions/extended/S-BSST1/fileList/test-file-list/files",
+                filesUrl = "/submissions/extended/S-BSST1/referencedFiles/test-file-list",
                 pageTabFiles = listOf(NfsFile("filePath", "relPath", referencedFile, emptyList()))
             )
         )
@@ -114,7 +114,7 @@ class SubmissionServiceMigrationTest(
     private fun setUpExtSubmissionClient() {
         every { sourceClient.getExtByAccNo("S-BSST1") } returns extSubmission
         every {
-            sourceClient.getReferencedFiles("/submissions/extended/S-BSST1/fileList/test-file-list/files")
+            sourceClient.getReferencedFiles("/submissions/extended/S-BSST1/referencedFiles/test-file-list")
         } returns ExtFileTable(
             NfsFile(
                 filePath = "folder/referenced.pdf",
--- a/submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/nfs/NfsFileProcessingConfig.kt
+++ b/submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/nfs/NfsFileProcessingConfig.kt
@@ -39,8 +39,12 @@ fun NfsFileProcessingConfig.nfsCopy(extFile: NfsFile): NfsFile {
 }
 
 fun NfsFileProcessingConfig.nfsMove(extFile: NfsFile): NfsFile {
+    val file = extFile.file
     val target = targetFolder.resolve(extFile.filePath)
     val subFile = subFolder.resolve(extFile.filePath)
+
+    logger.info { "$accNo $owner Moving file $file with size ${file.size()} into ${target.absolutePath}" }
+
     if (target.exists().not()) moveFile(extFile.file, target, permissions)
     return extFile.copy(fullPath = subFile.absolutePath, file = subFile)
 }


client/bio-commandline/src/test/kotlin/uk/ac/ebi/biostd/client/cli/services/SubmissionServiceMigrationTest.kt
```diff
@@ -109,6 +109,15 @@ class SubmissionServiceMigrationTest(
         every { sourceClient.getExtByAccNo("S-BSST1") } returns extSubmission
         every {
             sourceClient.getReferencedFiles("/submissions/extended/S-BSST1/fileList/test-file-list/files")
-        } returns ExtFileTable(NfsFile(referencedFile.name, referencedFile, emptyList()))
+        } returns ExtFileTable(
+            NfsFile(
+                referencedFile.name,
+                "filePath",
```
jhoanmanuelms, line 115: I think it'd be more realistic if you used these fields for their actual purpose like
"file1.txt"
"my/folder/file1.txt"
"Files/my/folder/file1.txt"

submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/fire/FireFilesService.kt
```diff
@@ -46,21 +46,24 @@ fun FireFileProcessingConfig.processNfsFile(relPath: String, nfsFile: NfsFile):
...
-        store.objectSize.toLong(),
+        nfsFile.filePath,
+        nfsFile.relPath,
+        fireFile.fireId,
+        fireFile.md5,
+        fireFile.size,
         nfsFile.attributes
     )
+
+private fun FireFileProcessingConfig.saveFile(subRelPath: String, nfsFile: NfsFile): ExtFile {
```
jhoanmanuelms, line 59: 👍

submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/pagetab/FirePageTabService.kt
```diff
@@ -21,53 +20,54 @@ class FirePageTabService(
...
 
         return when {
-            section.changed -> sub.copy(pageTabFiles = extFiles(sub.filesPath, subFiles), section = section.section)
-            else -> sub.copy(pageTabFiles = extFiles(sub.filesPath, subFiles))
+            section.changed -> sub.copy(pageTabFiles = subExtFiles(subFiles, sub.relPath), section = section.section)
+            else -> sub.copy(pageTabFiles = subExtFiles(subFiles, sub.relPath))
         }
     }
 
     private fun updateFileList(sec: ExtSection, path: String, tab: Map<String, TabFiles>): Section {
```
jhoanmanuelms, line 31: Sorry I didn't catch this in previews reviews but I think the `TabFiles` class should be called `PageTabFiles`

submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/pagetab/FirePageTabService.kt
```diff
@@ -21,53 +20,54 @@ class FirePageTabService(
...
         return when (val lst = sec.fileList) {
             null -> Section(false, sec)
-            else -> Section(
-                true, sec.copy(fileList = lst.copy(pageTabFiles = extFiles(path, tab.getValue(lst.fileName))))
-            )
+            else -> {
+                val name = lst.fileName
+                val tabFiles = tab.getValue(name)
```
jhoanmanuelms, line 36: Same with this variable

submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/pagetab/FirePageTabService.kt
```diff
@@ -21,53 +20,54 @@ class FirePageTabService(
...
-            pageTab.tsv.name,
-            pageTab.tsv.path,
-            tsv.fireOid,
-            tsv.objectMd5,
-            tsv.objectSize.toLong(),
-            listOf()
+    private fun saveFileListFile(file: File, subFolder: String, filePath: String): FireFile {
+        val relPath = "Files/$filePath"
+        val db = fireWebClient.save(file, file.md5(), "$subFolder/$relPath")
+        return FireFile(
```
jhoanmanuelms, line 51: If we don't use named params, could this fit in a single line like [here](https://github.com/EBIBioStudies/biostudies-backend-services/pull/456/files#diff-6de59b48d495dece61110c80d0ead258363f3dfe275132ace53f5d49d464a903R66)?

submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/pagetab/FirePageTabService.kt
```diff
@@ -21,53 +20,54 @@ class FirePageTabService(
...
+    private fun subExtFiles(pageTab: TabFiles, subFolder: String): List<ExtFile> = listOf(
+        saveSubFile(pageTab.json, subFolder),
+        saveSubFile(pageTab.xml, subFolder),
+        saveSubFile(pageTab.tsv, subFolder)
+    )
 
-        return listOf(extJson, extXml, extTsv)
+    private fun saveSubFile(file: File, subFolder: String): FireFile {
+        val name = file.name
+        val db = fireWebClient.save(file, file.md5(), "$subFolder/${file.name}")
```
jhoanmanuelms, line 70: We actually have the same block of code in several places. Should this be moved to a helper method? 🤔

submission/submission-webapp/gradle.properties
```diff
@@ -1,3 +1,3 @@
 coverage=0.7
-itest.mode=mysql
-enableFire=false
\ No newline at end of file
+itest.mode=mongo
+enableFire=true
```
jhoanmanuelms, line 3: Missing EOF

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/FileListSubmissionTest.kt
```diff
@@ -179,34 +181,86 @@ internal class FileListSubmissionTest(private val tempFolder: TemporaryFolder) :
...
-                if (enableFire) listOf() else listOf(
-                    NfsFile("$accNo.json", File(subFolder).resolve("$accNo.json")),
-                    NfsFile("$accNo.xml", File(subFolder).resolve("$accNo.xml")),
-                    NfsFile("$accNo.pagetab.tsv", File(subFolder).resolve("$accNo.pagetab.tsv"))
-                )
-            } else listOf()
-
-            assertThat(createdSub.section.fileList).isEqualTo(ExtFileList("FileList", pageTabFiles = expectedTabFiles))
-            assertThat((createdSub.pageTabFiles)).isEqualTo(submissionPageTabFiles)
+            if (mongoMode)
```
jhoanmanuelms, line 184: Why are the brackets ( {} ) missing on this `if` statement? This ain't no python kid!

In any case, this shouldn't be necessary. Why should the files be affected by the database if we're consuming extended? The extended submission object should be the same regardless if it was persisted on mongo or mysql

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/FileListSubmissionTest.kt
```diff
@@ -179,34 +181,86 @@ internal class FileListSubmissionTest(private val tempFolder: TemporaryFolder) :
...
+            assertThat(Paths.get("$subFolder/Files/FileList.xml")).exists()
+            assertThat(Paths.get("$subFolder/Files/FileList.json")).exists()
+            assertThat(Paths.get("$subFolder/Files/FileList.pagetab.tsv")).exists()
 
             assertThat(Paths.get("$subFolder/${createdSub.accNo}.xml")).exists()
             assertThat(Paths.get("$subFolder/${createdSub.accNo}.json")).exists()
             assertThat(Paths.get("$subFolder/${createdSub.accNo}.pagetab.tsv")).exists()
         }
+
+        private fun submissionFireTabFiles(accNo: String, subFolder: String): List<FireFile> {
```
jhoanmanuelms, line 213: For consistency, I think we should always use PageTabFiles rather than TabFiles i.e. `submissionFirePageTabFiles`. The same apply to the other methods.

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/FileListSubmissionTest.kt
```diff
@@ -179,34 +181,86 @@ internal class FileListSubmissionTest(private val tempFolder: TemporaryFolder) :
...
+                    val fileListTabFiles = createdSub.section.fileList!!.pageTabFiles as List<NfsFile>
+                    assertThat(fileListTabFiles).hasSize(3)
+                    assertThat(fileListTabFiles).isEqualTo(fileListNfsTabFiles(subFolder))
+                }
 
             assertThat(Paths.get("$subFolder/Files/$testFile")).exists()
-            assertThat(Paths.get("$subFolder/Files/$fileListName.xml")).exists()
-            assertThat(Paths.get("$subFolder/Files/$fileListName.json")).exists()
-            assertThat(Paths.get("$subFolder/Files/$fileListName.pagetab.tsv")).exists()
+            assertThat(Paths.get("$subFolder/Files/FileList.xml")).exists()
```
jhoanmanuelms, line 204: Why are the page tab files on NFS always checked? Shouldn't this be checked only if the system is not running on fire mode?

I think we should have something like:

`if (enableFire) assertFireSubmissionFiles() else assertNfsSubmissionFiles()`

And then, move the assertion logic to this methods accordingly

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/MultipartFileSubmissionApiTest.kt
```diff
@@ -266,53 +268,90 @@ internal class MultipartFileSubmissionApiTest(
...
-                if (enableFire) listOf() else listOf(
-                    NfsFile("$accNo.json", File(subFolder).resolve("$accNo.json")),
-                    NfsFile("$accNo.xml", File(subFolder).resolve("$accNo.xml")),
-                    NfsFile("$accNo.pagetab.tsv", File(subFolder).resolve("$accNo.pagetab.tsv"))
-                )
-            } else listOf()
-
-            assertThat(createdSub.section.fileList).isEqualTo(ExtFileList("FileList", pageTabFiles = expectedTabFiles))
-            assertThat((createdSub.pageTabFiles)).isEqualTo(submissionPageTabFiles)
+            if (mongoMode)
```
jhoanmanuelms, line 275: Same previous comment. I'm even wondering now if we should have different test classes that are enabled/disabled based on the properties 🤔. But this would be a "nice to have" thing that we could tackle in a separate ticket.

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/FileListSubmissionTest.kt
Juan-EBI: Both submission and file list page tab files are not stored in mysql just in mongo.  We did not update schema to support this as is not worth it given we are migrating now.

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/FileListSubmissionTest.kt
Juan-EBI: TestWireMockTransformer class store the files in the same path in that way we do not need specific logic for this part for each system (Fire and NFS).

submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/pagetab/FirePageTabService.kt
Juan-EBI: yes totally agree,  I think we need to decorate client  into fireservice and add there this api and probably also retry mechanism.

PR #465, opened 2021-11-03
Pivotal ID # 179922324: File Mode CLI Option

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/179922324

-  Add an option to move files instead of copying in the CLI submit operations
- Adjust the involved endpoints and web client's methods
- Add a converter for the "attributes" parameter

### DIFF
--- a/client/bio-commandline/src/main/kotlin/uk/ac/ebi/biostd/client/cli/commands/SubmitCommand.kt
+++ b/client/bio-commandline/src/main/kotlin/uk/ac/ebi/biostd/client/cli/commands/SubmitCommand.kt
@@ -1,16 +1,20 @@
 package uk.ac.ebi.biostd.client.cli.commands
 
 import com.github.ajalt.clikt.core.CliktCommand
+import com.github.ajalt.clikt.parameters.options.flag
 import com.github.ajalt.clikt.parameters.options.option
 import com.github.ajalt.clikt.parameters.options.required
 import com.github.ajalt.clikt.parameters.types.file
-import uk.ac.ebi.biostd.client.cli.common.CommonParameters.ATTACHED_HELP
-import uk.ac.ebi.biostd.client.cli.common.CommonParameters.INPUT_HELP
+import ebi.ac.uk.extended.model.FileMode.COPY
+import ebi.ac.uk.extended.model.FileMode.MOVE
 import uk.ac.ebi.biostd.client.cli.common.CommonParameters.ON_BEHALF_HELP
 import uk.ac.ebi.biostd.client.cli.common.CommonParameters.PASSWORD_HELP
 import uk.ac.ebi.biostd.client.cli.common.CommonParameters.SERVER_HELP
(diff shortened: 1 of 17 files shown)


client/bio-commandline/src/main/kotlin/uk/ac/ebi/biostd/client/cli/commands/SubmitCommand.kt
```diff
@@ -23,6 +27,7 @@ internal class SubmitCommand(private val submissionService: SubmissionService) :
     private val onBehalf by option("-b", "--onBehalf", help = ON_BEHALF_HELP)
     private val input by option("-i", "--input", help = INPUT_HELP).file(exists = true).required()
     private val attached by option("-a", "--attached", help = ATTACHED_HELP)
+    private val moveFiles by option("-m", "--moveFiles", help = MOVE_FILES).flag(default = false)
```
Juan-EBI, line 30: I think here we should use same convention as in rest api which is fileMode with default value.  In  that way is easier if someone is transition from comeline to direct api call or when discussing about the functionality.

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/web/resources/MultipartSubmitResource.kt
```diff
@@ -131,17 +132,18 @@ class MultipartSubmitResource(
...
         val subFile = tempFileGenerator.asFile(file)
+
         val contentWebRequest = FileSubmitWebRequest(
             submission = subFile,
             onBehalfRequest = onBehalfRequest,
             user = user,
             format = TSV,
             fileMode = mode,
-            attrs = attributes,
+            attrs = attributes?.associate { it.name to it.value } ?: emptyMap(),
```
Juan-EBI, line 146: I think here you can do `attributes.orEmpty().associate { it.name to it.value }` not big change but lik 1% simpler.

client/bio-commandline/src/main/kotlin/uk/ac/ebi/biostd/client/cli/commands/SubmitCommand.kt
```diff
@@ -31,7 +36,8 @@ internal class SubmitCommand(private val submissionService: SubmissionService) :
             password = password,
             onBehalf = onBehalf,
             file = input,
-            attached = attached?.split(FILES_SEPARATOR)?.flatMap { getFiles(File(it)) }.orEmpty()
+            attached = attached?.split(FILES_SEPARATOR)?.flatMap { getFiles(File(it)) }.orEmpty(),
```
Juan-EBI, line 39: I think we can move orEmpty to the begining to avoid the null chcecks 
`attached.orEmpty().split(FILES_SEPARATOR).flatMap { getFiles(File(it)) }`

Juan-EBI (approved):
LGTM please consider comments before merge

PR #472, opened 2021-11-19
Pivotal #ID 180345449: handle empty title submission email notification

Juan-EBI:
https://www.pivotaltracker.com/story/show/180345449

### DIFF
--- a/commons/commons-model-extended/src/main/kotlin/ebi/ac/uk/extended/model/ExtSubmissionExtensions.kt
+++ b/commons/commons-model-extended/src/main/kotlin/ebi/ac/uk/extended/model/ExtSubmissionExtensions.kt
@@ -17,5 +17,5 @@ val ExtSubmission.isCollection
 val ExtSubmission.filesPath
     get(): String = "$relPath/Files"
 
-val ExtSubmission.description
+val ExtSubmission.computedTitle
     get(): String? = title ?: section.title
--- a/commons/commons-model-extended/src/main/kotlin/ebi/ac/uk/extended/model/ExtSubmissionExtensions.kt
+++ b/commons/commons-model-extended/src/main/kotlin/ebi/ac/uk/extended/model/ExtSubmissionExtensions.kt
@@ -16,3 +16,6 @@ val ExtSubmission.isCollection
 
 val ExtSubmission.filesPath
     get(): String = "$relPath/Files"
+
+val ExtSubmission.description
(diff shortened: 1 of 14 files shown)


commons/commons-model-extended/src/test/kotlin/ebi/ac/uk/extended/model/ExtSubmissionExtensionsTest.kt
```diff
@@ -0,0 +1,44 @@
...
+        val submissionTitle = testSubmission(subTitle = "submission title", secTitle = null)
+        val submissionNoTitleSecTitle = testSubmission(subTitle = null, secTitle = "section title")
+        val submissionNoTitleNoSecTitle = testSubmission(subTitle = null, secTitle = null)
+
+        assertThat(submissionTitle.description).isEqualTo("submission title")
+        assertThat(submissionNoTitleSecTitle.description).isEqualTo("section title")
+        assertThat(submissionNoTitleNoSecTitle.description).isNull()
+    }
+
+    private fun testSubmission(subTitle: String? = null, secTitle: String? = null): ExtSubmission {
```
jhoanmanuelms, line 20: One-liner?

commons/commons-model-extended/src/main/kotlin/ebi/ac/uk/extended/model/ExtSubmissionExtensions.kt
```diff
@@ -16,3 +16,6 @@ val ExtSubmission.isCollection
 
 val ExtSubmission.filesPath
     get(): String = "$relPath/Files"
+
+val ExtSubmission.description
```
jhoanmanuelms, line 20: I think it's confusing that the description actually returns the title. We should call it something like `submissionTitle` or something like that.

jhoanmanuelms (approved):
LGTM. Please consider my comment before merging.

commons/commons-model-extended/src/main/kotlin/ebi/ac/uk/extended/model/ExtSubmissionExtensions.kt
Juan-EBI: I had issues naming this as sometimes is the section title and submissionTitle is kind of redudant as it bellows to submission object and is confusing with the title attribute itself.

PR #473, opened 2021-11-22
Pivotal ID # 180152694: File List Validation

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/180152694

- Improve file list validation error messages
- Add handler to return the corresponding status code


Juan-EBI (approved):
LGTM

PR #492, opened 2022-01-16
Pivotal ID # 180928073: Xml deserialization/serialization null support

Juan-EBI:
https://www.pivotaltracker.com/story/show/180928073


jhoanmanuelms (approved):
LGTM

PR #501, opened 2022-02-01
Pivotal ID # 180481465: No attribute/empty attribute value handling

Juan-EBI:
https://www.pivotaltracker.com/story/show/180481465


commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/tsv/deserialization/common/Functions.kt
```diff
@@ -18,56 +17,51 @@ internal inline fun validate(value: Boolean, lazyMessage: () -> String) {
...
 }
 
-private fun getDetailName(attrName: String) = attrName.substring(1, attrName.lastIndex)
-
-private fun addNameAttributeDetail(name: String, value: String, attributes: MutableList<Attribute>) {
-    attributes.ifEmpty { throw InvalidElementException(MISPLACED_ATTR_NAME) }
-    attributes.last().nameAttrs.add(AttributeDetail(name, value))
+private fun getAttributes(values: List<Pair<String, String?>>): List<Attribute> = buildList {
+    var previous: Attribute? = null
+    for ((header, value) in values) {
```
jhoanmanuelms, line 43: Just to clarify: we are only accepting empty values for general attributes, right? Neither namequal nor valqual

jhoanmanuelms (approved):
LGTM. Please consider my question.

commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/tsv/deserialization/common/Functions.kt
Juan-EBI: good question let me ask question to Ugis as I do not know.

PR #503, opened 2022-02-03
Pivotal ID # 181142221: Emails formating issues

Juan-EBI:
https://www.pivotaltracker.com/story/show/181142221

### DIFF
--- a/submission/notifications/src/main/resources/templates/successful-resubmission.txt
+++ b/submission/notifications/src/main/resources/templates/successful-resubmission.txt
@@ -1,6 +1,6 @@
 Dear ${USERNAME},
 
-Your BioStudies submission ${TITLE} with accession number ${ACC_NO} has been updated. You will be able to see the updated version at ${UI_URL}/studies/${ACC_NO} in the next 24 hours. ${RELEASE_MESSAGE}
+Your BioStudies ${TITLE} with accession number ${ACC_NO} has been updated. You will be able to see the updated version at ${UI_URL}/studies/${ACC_NO} in the next 24 hours. ${RELEASE_MESSAGE}
 
 Should you have any further questions, please reply to this message making sure you keep the subject or contact us at ${MAIL_TO}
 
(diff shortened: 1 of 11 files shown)


submission/notifications/src/main/resources/templates/successful-resubmission.txt
```diff
@@ -1,9 +1,9 @@
 Dear ${USERNAME},
 
-Your BioStudies submission ${TITLE} with accession number ${ACC_NO} has been updated. You will be able to see the updated version at ${UI_URL}/studies/${ACC_NO} in the next 24 hours. ${RELEASE_MESSAGE}
+Your BioStudies ${TITLE} with accession number ${ACC_NO} has been updated. You will be able to see the updated version at ${UI_URL}/studies/${ACC_NO} in the next 24 hours. ${RELEASE_MESSAGE}
```
jhoanmanuelms, line 3: Why are we removing the word "submission"? Maybe we should remove "BioStudies" instead, I think it'd look better:

"Your submission My Super Study with accession..."

Rather than

"Your Biostudies My Super Study with accession..."

submission/notifications/src/main/resources/templates/successful-submission.txt
```diff
@@ -1,6 +1,6 @@
 Dear ${USERNAME},
 
-Thank you for submitting your data to BioStudies. Your submission ${TITLE} has been assigned the BioStudies accession number ${ACC_NO}.
+Thank you for submitting your data to BioStudies. Your ${TITLE} has been assigned the BioStudies accession number ${ACC_NO}.
```
jhoanmanuelms, line 3: Same here

jhoanmanuelms (approved):

submission/notifications/src/main/resources/templates/successful-resubmission.txt
Juan-EBI: I am not changing  email content as part of ticket  I am not removing submission word but including as part for keywork {TITLE} as when there is not title email need to be "Your BioStudies submission with" rather than "Your BioStudies submission  with" (extra space product of replace title with an space. )

submission/notifications/src/main/resources/templates/successful-submission.txt
Juan-EBI: same answer as before :)

PR #507, opened 2022-02-07
Pivotal ID # 181179211 : Unactive draft when request is accepted

Juan-EBI:
https://www.pivotaltracker.com/story/show/181179211


jhoanmanuelms:
We have this method on the Submission service class that logs an error and sends a message to the failed-submission-log queue in case a submission fails

https://github.com/EBIBioStudies/biostudies-backend-services/blob/master/submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/service/SubmissionService.kt#L68

I think here we should set the draft status back to "ACTIVE"

jhoanmanuelms (approved):
LGTM. Please consider my comment.

Juan-EBI:
@jhoanmanuelms given we already acept the request (and validate) I think is most likely that problem is in our side rather than customer one that is the reason that we are not re activating draft on error.

PR #511, opened 2022-02-09
Pivotal ID # 181206866 Remove application property enableMongo

Juan-EBI:
https://www.pivotaltracker.com/story/show/181206866


jhoanmanuelms (approved):
LGTM

PR #523, opened 2022-03-08
Pivotal ID # 181416527: Fixes for FIRE

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/181416527

- Set the path to the files of every submission even if these are private
- Clean submission folder before files processing
- Remove unnecessary parameter in the FIRE save operation
- Include referenced files in the releasing process

### DIFF
--- a/submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/fire/FireFilesService.kt
+++ b/submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/fire/FireFilesService.kt
@@ -26,7 +26,7 @@ class FireFilesService(
         val (sub, _, previousFiles) = request
         logger.info { "${sub.accNo} ${sub.owner} Persisting files of submission ${sub.accNo} on FIRE" }
 
-        cleanFtpFolder(sub)
+        cleanSubmissionFolder(sub)
         val config = FireFileProcessingConfig(sub.accNo, sub.owner, sub.relPath, fireWebClient, previousFiles)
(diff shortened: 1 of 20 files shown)


submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/fire/FireFilesService.kt
```diff
@@ -9,25 +10,43 @@ import ebi.ac.uk.extended.model.ExtSubmission
...
+        cleanFtpFolder(sub)
         val config = FireFileProcessingConfig(sub.accNo, sub.owner, sub.relPath, fireWebClient, previousFiles)
         val processed = processFiles(sub) { config.processFile(request.submission, it) }
 
         logger.info { "${sub.accNo} ${sub.owner} Finished persisting files of submission ${sub.accNo} on FIRE" }
 
         return processed
     }
+
+    private fun cleanFtpFolder(submission: ExtSubmission) {
```
Juan-EBI, line 38: cleaning the FTP folder here does not look correct for me. If you notice we have two interfaces FtpService and FilesService, the first one is in charge of ftp operation and the second one is in charge of Ftp operation. As this class is the implementation of the first one, no ftp concer should be included here. 

I think what we need is to change ftp service implementation to unpublish old files.

jhoanmanuelms: Yes, perhaps the naming was poor, it should be clean submission folder

PR #524, opened 2022-03-09
Pivotal ID # 180582617: Change Password Reset Email Text

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/180582617

Display link text in html based e-mails

### DIFF
--- a/submission/notifications/src/main/resources/templates/activation.html
+++ b/submission/notifications/src/main/resources/templates/activation.html
@@ -3,7 +3,7 @@
 <p>Dear ${USERNAME}</p>
 <p>You have received this email because you have signed up for a BioStudies database account. To activate your account
   please click on the following link:<br/>
-  <a href="${URL}">Activation link</a>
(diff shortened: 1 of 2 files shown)


submission/notifications/src/main/resources/templates/activation.html
```diff
@@ -3,7 +3,7 @@
 <p>Dear ${USERNAME}</p>
 <p>You have received this email because you have signed up for a BioStudies database account. To activate your account
   please click on the following link:<br/>
-  <a href="${URL}">Activation link</a>
+  <a href="${URL}">${URL}</a>
```
Juan-EBI, line 6: why do we want to change title and display an url directly?

jhoanmanuelms: As stated in the ticket, non HTML based e-mail clients won't render the link correctly

Juan-EBI (approved):
LGTM

PR #526, opened 2022-03-11
Pivotal ID # 179922370: Modify Submission Metadata

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/179922370

- Add a resource to update the metadata of a submission
- Add support for FIRE files deserialization
- Improve logging messages

### DIFF
--- a/submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/web/handlers/MetadataUpdateWebHandler.kt
+++ b/submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/web/handlers/MetadataUpdateWebHandler.kt
@@ -1,0 +1,62 @@
+package ac.uk.ebi.biostd.submission.web.handlers
+
+import ac.uk.ebi.biostd.integration.SerializationService
+import ac.uk.ebi.biostd.submission.domain.helpers.SourceGenerator
+import ac.uk.ebi.biostd.submission.domain.service.ExtSubmissionService
+import ac.uk.ebi.biostd.submission.domain.service.SubmissionService
+import ac.uk.ebi.biostd.submission.model.SubmitRequest
+import ac.uk.ebi.biostd.submission.web.model.FileSubmitWebRequest
+import ebi.ac.uk.extended.mapping.to.toFile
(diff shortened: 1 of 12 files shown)


submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/web/resources/SubmissionMetadataResource.kt
```diff
@@ -0,0 +1,59 @@
...
+    // TODO Add support for the rest of submission extensions and direct page tab content
+    // TODO Add documentation for the update metadata endpoints
+    // TODO Add support for the update metadata operation in the CLI
+    @PutMapping(
+        value = ["/metadata/update"],
+        headers = ["${HttpHeaders.CONTENT_TYPE}=$MULTIPART_FORM_DATA"],
+        produces = [MediaType.APPLICATION_JSON_VALUE]
+    )
+    @ResponseBody
+    fun updateMetadata(
```
Juan-EBI, line 41: I was thinking we were going to implement detect metada change ourself instead of making ST or use choose wich approach to do. This looks not good for me, it looks like if email client has to options, write email and write email with attachtment.

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/web/handlers/MetadataUpdateWebHandler.kt
```diff
@@ -0,0 +1,62 @@
...
+import ebi.ac.uk.model.extensions.fileListName
+
+class MetadataUpdateWebHandler(
+    private val sourceGenerator: SourceGenerator,
+    private val webHandlerHelper: WebHandlerHelper,
+    private val submissionService: SubmissionService,
+    private val serializationService: SerializationService,
+    private val extSubmissionService: ExtSubmissionService
+) {
+    fun updateMetadata(request: FileSubmitWebRequest) {
```
Juan-EBI, line 28: from what I see this operation will be as a slow as normal operation as all files will be re process (md5 calculated again)

PR #527, opened 2022-03-15
Pivotal ID # 181328889: File List Extension Is Overwritten

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/181328889

Include file lists in the resubmission request

### DIFF
--- a/commons/commons-model-extended/src/test/kotlin/ebi/ac/uk/extended/model/ExtSubmissionExtensionsTest.kt
+++ b/commons/commons-model-extended/src/test/kotlin/ebi/ac/uk/extended/model/ExtSubmissionExtensionsTest.kt
@@ -1,11 +1,23 @@
 package ebi.ac.uk.extended.model
 
+import arrow.core.Either.Companion.left
+import ebi.ac.uk.io.ext.md5
+import ebi.ac.uk.io.ext.size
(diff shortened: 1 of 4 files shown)


commons/commons-model-extended/src/test/kotlin/ebi/ac/uk/extended/model/ExtSubmissionExtensionsTest.kt
```diff
@@ -1,11 +1,23 @@
...
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.ExtendWith
 import java.time.OffsetDateTime
 import java.time.ZoneOffset
 
-class ExtSubmissionExtensionsTest {
+@ExtendWith(TemporaryFolderExtension::class)
+class ExtSubmissionExtensionsTest(
+    tempFolder: TemporaryFolder
```
Juan-EBI, line 16: this is only needed in  one test so we can declare both variables (files) and folder directly in test case

Juan-EBI (commented):
Please check comment also I do not see coding changes solve ticket described issue. It looks to be that all this is doing is including page tab files and file list files in the list of previos files.

Juan-EBI (approved):

PR #528, opened 2022-03-16
Pivotal ID # 180490482: Include Root Path File Not Found Error

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/180490482

Include root path in the errors related to files not being found

### DIFF
--- a/commons/commons-bio/src/main/kotlin/ebi/ac/uk/io/sources/FilesSource.kt
+++ b/commons/commons-bio/src/main/kotlin/ebi/ac/uk/io/sources/FilesSource.kt
@@ -5,9 +5,13 @@ import ebi.ac.uk.io.ext.size
 import java.io.File
 
 interface FilesSource {
+    val rootPath: String?
+
     fun exists(filePath: String): Boolean
 
     fun getFile(filePath: String): BioFile
+
+    fun fullPath(filePath: String): String = "${rootPath?.plus("/").orEmpty()}$filePath"
 }
 
 sealed class BioFile {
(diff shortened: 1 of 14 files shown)


commons/commons-bio/src/main/kotlin/ebi/ac/uk/io/sources/FilesSource.kt
```diff
@@ -5,9 +5,13 @@ import ebi.ac.uk.io.ext.size
 import java.io.File
 
 interface FilesSource {
+    val rootPath: String?
```
Juan-EBI, line 8: this looks incorrect for me, root path is not related at all with concept of source (a generic interface that allow to obtain a file). This is coupling this class with an special feature of submitter that allow use to locate files in a folder which in my opinion is not related at all. 

You can notice this is not a good modeling option as for example ComposedFileSource now have a root path which does not make sense given compose only is a file source compoused of others files not relation at all.

Juan-EBI: I think what we need to do is change interface so getFile return nullable file  `BioFile?` and callers trigger exception with proper message which actually make sense as only caller knows what error to trigger and not the individual source which is totally normal they fail to find a fail.

Juan-EBI (commented):
Please check comments also I think given ticket is about error message to user, it will be nice to have a IT testing this case and specifically error message populated to user.

jhoanmanuelms:
Close in favor of PR from @Juan-EBI

PR #532, opened 2022-03-21
Pivotal ID #180902516: FileList class should load files (created mapping classes)

Juan-EBI:
https://www.pivotaltracker.com/story/show/180902516
- Changed functions to classes to be able to use serialization service operations in future changes


jhoanmanuelms (approved):
LGTM

PR #533, opened 2022-03-24
Pivotal ID # 181664033: Fire file source

Juan-EBI:
https://www.pivotaltracker.com/story/show/181664033


commons/commons-bio/src/main/kotlin/ebi/ac/uk/io/sources/FilesListSource.kt
```diff
@@ -1,13 +1,8 @@
...
 
 class FilesListSource(private val files: List<File>) : FilesSource {
-    override fun exists(filePath: String) = files.any { it.name == filePath }
-
-    override fun getFile(filePath: String): NfsBioFile {
-        val file = files.firstOrNull { it.name == filePath } ?: throw FileNotFoundException(filePath)
-        return NfsBioFile(file)
-    }
+    override fun getFile(path: String, md5: String?): NfsBioFile? =
+        files.firstOrNull { it.name == path }?.let { NfsBioFile(it) }
```
jhoanmanuelms, line 7: Why are we removing the exception? I think in some cases the file won't exist at all

commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/service/FileListSerializer.kt
```diff
@@ -37,6 +38,7 @@ internal class FileListSerializer(
             is FireBioFile -> TODO()
             is FireDirectoryBioFile -> TODO()
             is NfsBioFile -> bioFile.file
+            null -> throw FileNotFoundException(fileList)
```
jhoanmanuelms, line 41: Ok, I see. The only bad thing about leaving the responsibility for the exception outside the sources is that it can be easily forgotten

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/helpers/ExtFileListSource.kt
```diff
@@ -16,10 +16,8 @@ class ExtFileListSource(
     private val fireWebClient: FireWebClient,
     private val files: List<ExtFile>
 ) : FilesSource {
-    override fun exists(filePath: String): Boolean = files.any { it.fileName == filePath }
-
-    override fun getFile(filePath: String): BioFile {
-        val file = files.firstOrNull { it.fileName == filePath } ?: throw FileNotFoundException(filePath)
+    override fun getFile(path: String, md5: String?): BioFile {
+        val file = files.firstOrNull { it.fileName == path } ?: throw FileNotFoundException(path)
```
jhoanmanuelms, line 20: For example here, shouldn't we normalize that either all sources throw the exception or they don't?

jhoanmanuelms (commented):
WDYT about these comments?

commons/commons-bio/src/main/kotlin/ebi/ac/uk/io/sources/FilesListSource.kt
Juan-EBI: now sources return nullable so is caller responsability to throw it or hanble it.

commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/service/FileListSerializer.kt
Juan-EBI: It can be forgotten in Java but not in Kotlin, as return type is nullable so compilter will force you. Code will not compile if you do not handle null case

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/helpers/ExtFileListSource.kt
Juan-EBI: good catch , fixing it.

jhoanmanuelms (approved):
LGTM

PR #541, opened 2022-04-06
Pivotal ID # 181814601: Submission Source

Juan-EBI:
https://www.pivotaltracker.com/story/show/181814601

- simplified source model


commons/commons-model-extended-mapping/src/main/kotlin/ebi/ac/uk/io/sources/Common.kt
```diff
@@ -0,0 +1,21 @@
...
+
+import ebi.ac.uk.extended.mapping.from.toExtAttributes
+import ebi.ac.uk.extended.model.NfsFile
+import ebi.ac.uk.io.ext.md5
+import ebi.ac.uk.io.ext.size
+import ebi.ac.uk.model.Attribute
+import ebi.ac.uk.model.constants.FILES_RESERVED_ATTRS
+import java.io.File
+
+internal fun create(path: String, file: File, calculateProperties: Boolean, attributes: List<Attribute>): NfsFile {
```
jhoanmanuelms, line 11: From what I understood, the only reason for having the `calculateProperties` flag in the interface method is to decide whether or not to calculate the md5 and size but, it's actually only used here and the only place where it's explicitly set to true is when mapping the file to ExtFile (see the other comment)

This got me thinking about a couple things:
1. The only reason for getting a null file and having to use this method is because we're mapping for the first time from to extended in order to store it as a request for it to be processed later. We shouldn't have files that are not found once we're at the processing stage.
2. Right now, the calculation flag is only used for NFS files which makes sense because we'll only need to do these calculations for files that come from the user space and that haven't been previously used in a submission
3. The use of this flag is basically for avoiding the calculation at request time but doing it at processing time
4. We're always replacing the old instance of ExtFile (whatever the type, NfsFile or FireFile) at processing time i.e. the extended submission object returned by the processing services will have the updated ExtFile objects

I'm just thinking that maybe it would make sense to avoid this `calculateProperties` flag altogether and just calculate it specifically where we actually needed, that is, when processing the files and setting the values to the new ExtFile instances that are set to the processed ExtSubmission.

With that, we would always create them here with "NOT_CALCULATED" and -1 values

commons/commons-model-extended-mapping/src/main/kotlin/ebi/ac/uk/io/sources/Common.kt
```diff
@@ -0,0 +1,21 @@
...
+    return NfsFile(
+        filePath = path,
+        relPath = "Files/$path",
+        file = file,
+        fullPath = file.absolutePath,
+        md5 = if (calculateProperties) file.md5() else "NOT_CALCULATED",
```
jhoanmanuelms, line 17: Perhaps "NOT_CALCULATED" should be a constant?

jhoanmanuelms (approved):
LGTM. It makes a lot of sense and simplifies the code a lot. I just have a couple of comments for you to consider

PR #543, opened 2022-04-11
Pivotal ID # 181523667: Inner Folder File Lists

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/181523667

Use the file path instead of the file name to persist the referenced files


Juan-EBI (approved):
LGTM

PR #545, opened 2022-04-12
Pivotal ID # 181328889: File List Extension Is Overwritten

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/181328889

- Add base path to the FIRE files source
- Add integration test for reusing previous versions file list and referenced files


Juan-EBI (approved):
LGTM

PR #546, opened 2022-04-19
Pivotal ID # 181597483: FIRE Directories Persistence

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/181597483

- Simplify model by removing FireDirectory from both the extended model and the persistence layer
- Remove unused and duplicated code


submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/helpers/FireFilesSource.kt
```diff
@@ -73,5 +77,12 @@ fun FireApiFile.asFireBioFile(path: String, attributes: List<Attribute>): FireFi
...
         md5 = objectMd5,
         size = objectSize.toLong(),
+        type = fileType,
         attributes = attributes.toExtAttributes(FILES_RESERVED_ATTRS)
     )
+
+private val FireApiFile.fileType: ExtFileType
+    get(): ExtFileType {
+        val type = metadata!!.first { it.key == FIRE_BIO_FILE_TYPE }.value
+        return if (type == FILE.value) FILE else DIR
```
Juan-EBI, line 87: I think the default need to be file and not directory.

commons/commons-model-extended-serialization/src/main/kotlin/uk/ac/ebi/extended/serialization/serializers/ExtFileSerializer.kt
```diff
@@ -53,26 +48,13 @@ class ExtFileSerializer : JsonSerializer<ExtFile>() {
...
-        writeObjectField(ATTRIBUTES, file.attributes)
-        writeStringField(EXT_TYPE, ExtType.FireDirectory.type)
-        writeStringField(FILE_TYPE, FILE_DIR_TYPE)
-        writeStringField(FILE_MD5, file.md5)
-        writeNumberField(FILE_SIZE, file.size)
-        writeEndObject()
-    }
-
-    private fun getType(file: File) = if (FileUtils.isDirectory(file)) FILE_DIR_TYPE else FILE_FILE_TYPE
+    private val FireFile.extType: String
```
Juan-EBI, line 58: not sure about this one,  basically I am not sure when is better to use an extension function, and property and or a regular function when is private as this one and others in this PR. 

on one hand sintax is nice but in the other hand an private extension property is more complex thant a simple function (well know in every single programing language). IMO the nice sintax (used only in one or two places) does not pay off the extra complexity but not 100% sure.

Juan-EBI (approved):
please check commetns before merge

Juan-EBI (commented):
looking graeat, please check comments

commons/commons-model-extended-serialization/src/testFixtures/kotlin/uk/ac/ebi/extended/test/SubmissionFactory.kt
```diff
@@ -154,30 +155,31 @@ object FireFileFactory {
...
-    )
-
-    const val FIRE_ID = "dirFireId"
-    const val FILE_PATH = "folder/file.txt"
-    const val REL_PATH = "Files/folder/file.txt"
-    const val MD5 = "md5"
-    const val SIZE = 1L
-    val ATTRIBUTES = emptyList<ExtAttribute>()
-}
+// object FireDirectoryFactory {
```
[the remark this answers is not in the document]
jhoanmanuelms: 🤦

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/helpers/FireFilesSource.kt
jhoanmanuelms: Nice catch, even better I implemented it from the enum itself. The compiler is our friend

commons/commons-model-extended-serialization/src/main/kotlin/uk/ac/ebi/extended/serialization/serializers/ExtFileSerializer.kt
jhoanmanuelms: Well, in this case, I didn't use a function since it would be a "virtual" property from the FireFile. That being said, look at how cute it looks 🤩 

```
private val FireFile.extType: String
        get() = when(type) {
            FILE -> ExtType.FireFile.type
            DIR -> ExtType.FireDirectory.type
        }
```

jhoanmanuelms: Never mind, this isn't actually necessary since we really should have the file type telling us if it's either a file or directory and the ext file type telling us whether it's on NFS or FIRE. Please check my latest changes here.

Juan-EBI (approved):
LGTM

PR #547, opened 2022-04-20
Pivotal ID # 181946003: CLI: On Behalf Submit Not Working

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/181946003

Use "onBehalf" parameter on CLI submit operations


client/bio-commandline/src/test/kotlin/uk/ac/ebi/biostd/client/cli/services/SubmissionServiceTest.kt
```diff
@@ -35,114 +34,148 @@ internal class SubmissionServiceTest {
     fun beforeEach() = mockkObject(SecurityWebClient)
 
     @Test
-    fun `submit successful`() {
+    fun `submit copying files`() {
```
Juan-EBI, line 37: as logic is exactly the same I do not think we need two test for copying and moving.  Basically is like we do a unit test for a user with name "juan" and another user with name "jhoan" as system does not take diferent actions based on this input, IMO  is does no pay off to duplicate all test. 

In the case you do want to do this I would suggest to have single test and use junit5 paramterized test with [enum](https://www.baeldung.com/parameterized-tests-junit-5#3-enum) support.

Juan-EBI (approved):

PR #548, opened 2022-04-25
Pivotal Id: #181938430 Submissions files sequence processing

Juan-EBI:
https://www.pivotaltracker.com/story/show/181938430


commons/commons-model-extended-serialization/src/test/kotlin/uk/ac/ebi/extended/serialization/serializers/ExtSectionSerializerTest.kt
```diff
@@ -128,6 +131,7 @@ class ExtSectionSerializerTest(private val tempFolder: TemporaryFolder) {
             "fileList" to jsonObj {
                 "fileName" to "file-list"
                 "filesUrl" to "/submissions/extended/S-BSST1/referencedFiles/file-list"
+                "file" to allInOneSection.fileList?.file?.absolutePath
```
jhoanmanuelms, line 134: Since it's a test, shouldn't we know beforehand that it's not null?

commons/commons-util/src/main/kotlin/ebi/ac/uk/io/Streams.kt
```diff
@@ -0,0 +1,5 @@
+package ebi.ac.uk.io
+
+inline fun <A : AutoCloseable, B : AutoCloseable> use(a: A, b: B, func: (A, B) -> Unit) {
```
jhoanmanuelms, line 3: I think this method is not very clear at a single glane. Could we add some documentation?

commons/commons-util/src/main/kotlin/ebi/ac/uk/io/ext/FileExtensions.kt
```diff
@@ -18,6 +18,10 @@ fun File.createDirectory(name: String): File = Files.createDirectory(toPath().re
 
 fun File.createNewFile(name: String): File = resolve(name).apply { createNewFile() }
 
+fun File.createTempFile(prefix: String, suffix: String): File {
```
jhoanmanuelms, line 21: one-liner

submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/service/SubmissionMongoQueryService.kt
```diff
@@ -89,22 +93,27 @@ internal class SubmissionMongoQueryService(
...
-        .map { extFile -> loadFileAttributes(extFile) }
-        .toList()
+    private fun copyFile(inputFile: File, outputFile: File): File {
+        use(inputFile.inputStream(), outputFile.outputStream()) { input, output -> loadFiles(input, output) }
+        return outputFile
+    }
+
+    private fun loadFiles(input: InputStream, output: OutputStream) {
+        val files = serializationService.deserializeList(input)
+            .onEachIndexed { index, file -> logger.info { "mapping file ${file.filePath}, ${index + 1}" } }
```
jhoanmanuelms, line 113: In order to keep track on Kibana, all logs should have the accNo and owner

jhoanmanuelms (commented):
Please consider my comments

commons/commons-model-extended-serialization/src/main/kotlin/uk/ac/ebi/extended/serialization/service/ExtFilesResolver.kt
```diff
@@ -0,0 +1,28 @@
...
+        val name = fileName.replace("/", "-")
+        return baseFolder(subAccNo, version).createTempFile(name, ".json")
+    }
+
+    private fun baseFolder(subAccNo: String, version: Int): File {
+        val now = LocalDate.now()
+        val path = extendFilesPath
+            .resolve(now.get(YEAR).toString())
+            .resolve(now.get(MONTH_OF_YEAR).toString())
+            .resolve(now.get(DAY_OF_MONTH).toString())
```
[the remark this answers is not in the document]
Juan-EBI: the problem with the accession/version pattern is that we will end with houndred of thosons files that we will not be able to list and eventually will slow down processing (when linux limit of folders). So here we can replicate logic of relpath or this other option that I think was simpler. 

relpath one is more complex but also is pedictable, this one is simpler but is not predictable without checking modification date where folder are.

commons/commons-model-extended-serialization/src/main/kotlin/uk/ac/ebi/extended/serialization/service/ExtSerializationServiceExt.kt
```diff
@@ -0,0 +1,18 @@
+package uk.ac.ebi.extended.serialization.service
```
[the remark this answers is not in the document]
Juan-EBI: I think it depends , if there are many extension for many classes  it make sense to organize them in a package but here is only for one so IMO current structure is good. is like creating controller package for single controller.

commons/commons-model-extended-serialization/src/main/kotlin/uk/ac/ebi/extended/serialization/service/ExtFilesResolver.kt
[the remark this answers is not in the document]
jhoanmanuelms: Got it. My concern is a possibility rather than a reality so if we really find this schema is problematic, we could change it later.

jhoanmanuelms (approved):
LGTM

PR #554, opened 2022-05-06
Pivotal ID # 182034592: FIRE API Version

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/182034592

Add the version to the FIRE API

### DIFF
--- a/client/fire-webclient/src/main/kotlin/uk/ac/ebi/fire/client/integration/web/FireWebClient.kt
+++ b/client/fire-webclient/src/main/kotlin/uk/ac/ebi/fire/client/integration/web/FireWebClient.kt
@@ -6,6 +6,8 @@ import org.springframework.web.util.DefaultUriBuilderFactory
 import uk.ac.ebi.fire.client.api.FireClient
 import uk.ac.ebi.fire.client.exception.FireWebClientErrorHandler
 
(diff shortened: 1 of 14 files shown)


client/fire-webclient/src/main/kotlin/uk/ac/ebi/fire/client/integration/web/FireWebClient.kt
```diff
@@ -6,20 +6,24 @@ import org.springframework.web.util.DefaultUriBuilderFactory
 import uk.ac.ebi.fire.client.api.FireClient
 import uk.ac.ebi.fire.client.exception.FireWebClientErrorHandler
 
+const val FIRE_API_BASE = "fire"
```
Juan-EBI, line 9: can be private

submission/submission-webapp/src/main/resources/application.yml
```diff
@@ -67,6 +67,7 @@ app:
     host: # FIRE host
     username: # FIRE user
     password: # FIRE password
+    version: # FIRE version
```
Juan-EBI, line 70: I think is we can safaly set a default value as it is not sensible information.

Juan-EBI (approved):
LGTM please check comments before merge.

PR #556, opened 2022-05-06
Pivotal ID # 181568867: Remove .pagetab Extension

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/181568867

Remove the .pagetab extension from all the code base

### DIFF
--- a/commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/common/TsvPagetabExtension.kt
+++ b/commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/common/TsvPagetabExtension.kt
@@ -1,0 +1,7 @@
+package ac.uk.ebi.biostd.common
+
+class TsvPagetabExtension(
+    private val featureEnabled: Boolean = false
+) {
+    fun tsvExtension(): String = if (featureEnabled) "tsv" else "pagetab.tsv"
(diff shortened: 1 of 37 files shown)


Juan-EBI (commented):
Changes Looks greath,  but I think we need to include a [feature flag ](https://martinfowler.com/articles/feature-toggles.html). The reason is that if we merge as it is we will mix formats when merge into prod for UI and it will break it. 

What I think we can do is implement something like  _ExtensionFeatureFlag_ class with _tsvExtension_ method that add pagetab extension or not depending of a system property (by default in false).

What do you think?

commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/common/TsvPagetabExtension.kt
```diff
@@ -0,0 +1,7 @@
+package ac.uk.ebi.biostd.common
+
+class TsvPagetabExtension(
+    private val featureEnabled: Boolean = true
```
Juan-EBI, line 4: I think by default it needs to be "pagetab.tsv" not the other way around. I mean the default behavior is the safest option (current behavior so system does not change.)

Juan-EBI (approved):
LGTM

PR #557, opened 2022-05-10
Pivotal ID # 181559436: Allow on behalf in file list validation endpoint

nestor-diaz:
**Title**: Pivotal ID # 181559436: Develop a simple (internal) tool for file list checking so that we could easily help users when they have file list troubles.

**Pivotal Link:** https://www.pivotaltracker.com/story/show/181559436

Please include a list of the changes:
* Extracts common on behalf function in a utils class
* Includes onbehalf user files together with super admin user files
* Expects `onBehalfRequest` in the file list validation endpoint


submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/helpers/SourceGenerator.kt
```diff
@@ -19,8 +19,19 @@ class SourceGenerator(
 ) {
     fun userSources(
         user: SecurityUser,
-        rootPath: String? = null
-    ): FilesSource = ComposedFileSource(userSourcesList(user, rootPath.orEmpty()))
+        rootPath: String? = null,
+        onBehalfUser: SecurityUser?
+    ): FilesSource {
```
jhoanmanuelms, line 24: I don't think we should alter this method but instead, create the ComposedSource wherever required. This method is intended to create a source for any single user, not handling the `onBehalf` logic

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/web/resources/FileListResource.kt
```diff
@@ -14,11 +16,17 @@ import org.springframework.web.bind.annotation.RestController
...
     fun validateFileList(
         @BioUser user: SecurityUser,
-        @RequestParam(FILE_LIST_NAME) fileListName: String
-    ) = fileListValidator.validateFileList(fileListName, sourceGenerator.userSources(user))
+        onBehalfRequest: OnBehalfRequest?,
+        @RequestParam(FILE_LIST_NAME) fileListName: String,
+    ) {
+        val onBehalfUser = onBehalfRequest?.let { onBehalfUtils.getOnBehalfUser(it) }
+
+        fileListValidator.validateFileList(fileListName, sourceGenerator.userSources(user, null, onBehalfUser))
```
jhoanmanuelms, line 30: So here, instead of leaving the onBehalf logic to the source generator, you should do something like

```
val onBehalfUser = onBehalfRequest?.let { onBehalfUtils.getOnBehalfUser(it) }
val userSources = sourceGenerator.userSources(user)
val onBehalfSources = sourceGenerator.userSources(onBehalfUser)

fileListValidator.validateFileList(fileListName, ComposedFilesSource(userSources, onBehalfUserSources))
```

jhoanmanuelms (commented):
Please consider my comments

PR #558, opened 2022-05-10
Pivotal ID # 181807913: Add MD5 To Pagetab

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/181807913

Add calculated MD5 to pagetab files

### DIFF
--- a/commons/commons-model-extended-mapping/src/main/kotlin/ebi/ac/uk/extended/mapping/to/ToFile.kt
+++ b/commons/commons-model-extended-mapping/src/main/kotlin/ebi/ac/uk/extended/mapping/to/ToFile.kt
@@ -1,5 +1,6 @@
 package ebi.ac.uk.extended.mapping.to
 
+import ebi.ac.uk.extended.model.ExtAttribute
 import ebi.ac.uk.extended.model.ExtFile
 import ebi.ac.uk.extended.model.FireFile
 import ebi.ac.uk.extended.model.NfsFile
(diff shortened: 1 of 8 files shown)


commons/commons-model-extended-mapping/src/main/kotlin/ebi/ac/uk/extended/mapping/to/ToFile.kt
```diff
@@ -4,12 +4,17 @@ import ebi.ac.uk.extended.model.ExtFile
...
 
 fun ExtFile.toFile(): BioFile =
     when (this) {
-        is NfsFile -> BioFile(filePath, file.size(), type.value, attributes.mapTo(mutableListOf()) { it.toAttribute() })
-        is FireFile -> BioFile(filePath, size, type.value, attributes.mapTo(mutableListOf()) { it.toAttribute() })
+        is NfsFile -> BioFile(filePath, file.size(), type.value, toAttributes())
+        is FireFile -> BioFile(filePath, size, type.value, toAttributes())
     }
+
+private fun ExtFile.toAttributes() =
```
Juan-EBI, line 19: `attributes.map { it.toAttribute() } + Attribute(MD5.value, md5)`

Juan-EBI, line 19: I think name `toAttributes` and function is misleading (is an extension function of file)  as we are not transforming a file into attributes, we are transforming file attributes into ext attributes. To make more clear imagine you have a class `human`  that have a `born_country` attribute and one declarate an extension `Human.getPopulation` which obtain the population of the country. 

Even function implementation can be correct, is not correct to model a human with shuch operation.

Juan-EBI (commented):
LGTM please check comments. Also I have the following concern if user is passing md5 as an attribute, are we filtering out?  it looks to me that we will end duplicating it.

Juan-EBI (approved):
LGTM

PR #560, opened 2022-05-11
Pivotal ID # 181937926: Avoid restarting ITests containers

Juan-EBI:
https://www.pivotaltracker.com/story/show/181937926

### DIFF
--- a/submission/submission-webapp/src/itest/resources/META-INF/services/org.junit.platform.launcher.TestExecutionListener
+++ b/submission/submission-webapp/src/itest/resources/META-INF/services/org.junit.platform.launcher.TestExecutionListener
@@ -1,0 +1,1 @@
+ac.uk.ebi.biostd.itest.itest.ITestListener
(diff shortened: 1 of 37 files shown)


submission/submission-webapp/src/itest/resources/META-INF/services/org.junit.platform.launcher.TestExecutionListener
```diff
@@ -0,0 +1 @@
+ac.uk.ebi.biostd.itest.itest.ITestListener
```
jhoanmanuelms, line 1: EOF

jhoanmanuelms (commented):
These changes look great and something our itests were needing for a long time. If I understood them correctly, what we're doing now is starting the containers once and running all the tests there. We're now also sharing a single instance of the temporary folder through the tests.

I only have one suggestion, should we have a dictionary with the accessions used for each test? Maybe putting them all in constants somewhere so we avoid collisions in the future.

Other than that, I only left some minor comments but everything looks great.

Juan-EBI:
> These changes look great and something our itests were needing for a long time. If I understood them correctly, what we're doing now is starting the containers once and running all the tests there. We're now also sharing a single instance of the temporary folder through the tests.
> 
> I only have one suggestion, should we have a dictionary with the accessions used for each test? Maybe putting them all in constants somewhere so we avoid collisions in the future.
> 
> Other than that, I only left some minor comments but everything looks great.

Make sense yes,  I will create several tickets to enhace itest which need a lot of re organizations.

PR #569, opened 2022-05-30
Pivotal ID # 179922370: Modify Submission Metadata

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/179922370

- Add a parameter to specify the preferred files source
- Use the preferred source to prioritize the file sources
- Remove unused code and duplicated tests

### DIFF
--- a/submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/MultipartFileSubmissionApiTest.kt
+++ b/submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/MultipartFileSubmissionApiTest.kt
@@ -23,6 +23,7 @@ import ebi.ac.uk.extended.mapping.to.ToSubmissionMapper
 import ebi.ac.uk.extended.model.ExtSubmission
 import ebi.ac.uk.extended.model.FireFile
 import ebi.ac.uk.extended.model.NfsFile
+import ebi.ac.uk.extended.model.allSectionsFiles
 import ebi.ac.uk.extended.model.createNfsFile
 import ebi.ac.uk.io.ext.createFile
 import ebi.ac.uk.io.ext.md5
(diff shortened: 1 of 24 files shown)


submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/data/web/SubmissionDraftResource.kt
```diff
@@ -85,6 +88,7 @@ internal class SubmissionDraftResource(
         @BioUser user: SecurityUser,
         onBehalfRequest: OnBehalfRequest?,
         @RequestParam(FILE_MODE, defaultValue = "COPY") mode: FileMode,
+        @RequestParam(PREFERRED_SOURCE, defaultValue = "USER_SPACE") preferredSource: PreferredSource,
```
Juan-EBI, line 91: 1. This and every submission method is getting super loaded with all this parameters, I think we need to introduce class here for example annotated with model attribute.

Juan-EBI, line 91: 1. This and every submission method is getting super loaded with all this parameters, I think we need to introduce class here for example annotated with model attribute.

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/web/model/SubmitWebRequest.kt
```diff
@@ -24,6 +25,7 @@ sealed class SubmitWebRequest(
     val format: SubFormat,
     val fileMode: FileMode,
     val files: List<File>,
+    val preferredSource: PreferredSource,
```
Juan-EBI, line 28: same as before this class in my opinion are becoming like a map full of everything.

Juan-EBI, line 28: same as before this class in my opinion are becoming like a map full of everything.

Juan-EBI (approved):
Please take a look to comments,  most of them suggestions.

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/MultipartFileSubmissionApiTest.kt
```diff
@@ -255,6 +256,70 @@ class MultipartFileSubmissionApiTest(
...
+        file5.delete()
+        fileList.delete()
+
+        tempFolder.createFile("File4.txt", "content 4 updated")
+
+        assertThat(webClient.submitSingle(submission("FileList.json"), TSV, emptyList())).isSuccessful()
+        assertThat(innerFile.toFile().readText()).isEqualTo("content 4")
+        assertThat(referencedFile.toFile().readText()).isEqualTo("content 5")
+
+        if (enableFire) {
```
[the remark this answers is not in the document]
jhoanmanuelms: The assertion is being made over the content for both cases, this validation is specifically for the fire id in order to verify the same file is reused

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/SubmissionApiTest.kt
```diff
@@ -168,50 +165,6 @@ class SubmissionApiTest(
         assertThat(webClient.submitSingle(submission, TSV)).isSuccessful()
     }
 
-    @Test
```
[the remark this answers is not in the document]
jhoanmanuelms: These tests cases are duplicated in the class `SubmissionOnBehalfTest`

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/helpers/SourceGenerator.kt
```diff
@@ -17,57 +21,79 @@ class SourceGenerator(
...
-    fun userSources(
-        user: SecurityUser,
-        rootPath: String? = null
-    ): FilesSource = ComposedFileSource(userSourcesList(user, rootPath.orEmpty()))
-
     fun submissionSources(requestSources: RequestSources): FilesSource {
-        val (owner, submitter, files, rootPath, submission) = requestSources
-        return ComposedFileSource(submissionSources(owner, submitter, files, rootPath, submission))
-    }
+        val sources = buildList {
```
Juan-EBI, line 25: Here I see an issue with this code but I am aware is really open to discussion. Let me share my thoughts. Our final goal as developer is to make code easy to mantain and we do that but creating code easy to undersdant. All others practices i.e code re use , small functions etc IMO they are just side effects of this.

Here you have a function which is split in many small ones (which in theory is good because small functions are good ...) but in my opinion code is harder to read because one need to fo to many places to answer this simple questions "Give x configuration what sources do we use".  Which is not a random question that I just imagine my self but the question that any developer comming to this class will have.

 So in my opinion here we need to break the gold rule (of small function) and favor simplicity by having code all togther. 
 

 ```
     fun submissionSources(requestSources: RequestSources): FilesSource {
        val (owner, submitter, files, rootPath, submission, preferredSources) = requestSources
        val sources = preferredSources.ifEmpty { listOf(USER_SPACE, SUBMISSION, FIRE) }
        return ComposedFileSource(submissionSources(owner, submitter, files, rootPath, submission, sources))
    }

    private fun submissionSources(
        owner: SecurityUser?,
        submitter: SecurityUser?,
        files: List<File>,
        rootPath: String?,
        sub: ExtSubmission?,
        sources: List<PreferredSource>,
    ): List<FilesSource> = buildList {
        add(FilesListSource(files))
        sources.forEach { it ->
            when (it) {
                FIRE -> if (props.persistence.enableFire) add(fireSourceFactory.createFireSource())
                USER_SPACE -> {
                    if (submitter != null) {
                        add(createPathSource(submitter, rootPath))
                        addAll(groupSources(submitter.groupsFolders))
                    }
                    if (owner != null) {
                        add(createPathSource(owner, rootPath))
                        addAll(groupSources(owner.groupsFolders))
                    }
                }
                SUBMISSION -> {
                    if (sub != null) {
                        add(fireSourceFactory.createSubmissionFireSource(sub.accNo, Paths.get("${sub.relPath}/Files")))
                        add(PathFilesSource(Paths.get(props.submissionPath).resolve(sub.relPath).resolve(FILES_PATH)))
                    }
                }
            }
        }
    }
 ```

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/web/handlers/SubmitWebHandler.kt
```diff
@@ -45,60 +46,66 @@ class SubmitWebHandler(
...
+                preferredSources = preferredSource
             )
         )
-        val submission = withAttributes(submission(request.submission, request.format, source), request.attrs)
+        val submission = withAttributes(submission(request.submission, format, source), attrs)
 
         return SubmitRequest(
             submission = submission,
-            submitter = request.submitter,
+            submitter = submitter,
```
Juan-EBI, line 68: Not part of your changes but submitter should not be nuallable

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/web/model/SubmitWebRequest.kt
```diff
@@ -12,40 +15,45 @@ class OnBehalfRequest(
...
 }
 
-@Suppress("LongParameterList")
-sealed class SubmitWebRequest(
-    val submitter: SecurityUser,
-    val onBehalfRequest: OnBehalfRequest?,
+data class SubmissionRequestParameters(
+    val fileMode: FileMode = COPY,
+    val preferredSources: List<PreferredSource> = emptyList(),
+    val attributes: List<ExtAttributeDetail> = emptyList(),
```
Juan-EBI, line 25: Even this is convenient we should not expose our extended model (domain models) to user rest layer  as then we will need to evolve them indivudally.

Juan-EBI (commented):
Great changes!, system is getting improved a lot after this.

PR #570, opened 2022-05-31
Pivotal ID # 182333701: Submissions stages

Juan-EBI:
https://www.pivotaltracker.com/story/show/182333701

- Separated command or actions classes from query ones to reduce complexity.
- Added pending notification to event publisher to avoid direct interation with rabbit mq in others classes.

### DIFF
--- a/submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/ExtSubmissionSubmitter.kt
+++ b/submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/ExtSubmissionSubmitter.kt
@@ -40,7 +40,6 @@ class ExtSubmissionSubmitter(
 }
 
 class RequestProcessor(
-    private val submissionPersistenceService: SubmissionPersistenceService,
     private val submissionPersistenceQueryService: SubmissionPersistenceQueryService,
     private val fileProcessingService: FileProcessingService,
 ) {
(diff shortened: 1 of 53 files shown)


submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/ExtSubmissionSubmitter.kt
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    private fun saveRequest(request: SubmissionRequest, owner: String): Pair<String, Int> {
+        val saved = persistenceService.saveSubmissionRequest(request)
+        request.draftKey?.let { draftService.setProcessingStatus(owner, it) }
+        return saved
+    }
+}
+
+class RequestProcessor(
```
jhoanmanuelms, line 42: I think this should be on its own file since it's a functionality class. I understand these two have kind of the same business but I feel it would make more sense.

Also, I would suggest naming it `SubmissionRequestProcessor` since we have several kind of requests on the code and finding the name `RequestProcessor`, by itself, doesn't really give me a clue of what it does

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/SubmissionSubmitter.kt
```diff
@@ -147,7 +120,7 @@ class SubmissionSubmitter(
             collections = tags.map { ExtCollection(it) },
             section = toExtSectionMapper.convert(submission.accNo, version, submission.section, source),
             attributes = submission.attributes.toExtAttributes(SUBMISSION_RESERVED_ATTRIBUTES),
-            storageMode = if (properties.persistence.enableFire) FIRE else NFS
+            storageMode = if (properties.persistence.enableFire) StorageMode.FIRE else StorageMode.NFS
```
jhoanmanuelms, line 123: Same thing about the context

PR #571, opened 2022-05-31
Pivotal # 182332696: TSV File List Serialization

jhoanmanuelms:
- Include first list in the TSV file list serialization
- Add file list assertions to the itests

### DIFF
--- a/commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/tsv/deserialization/stream/FileListTsvStreamDeserializer.kt
+++ b/commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/tsv/deserialization/stream/FileListTsvStreamDeserializer.kt
@@ -15,7 +15,7 @@ import java.io.OutputStream
 internal class FileListTsvStreamDeserializer {
     fun serializeFileList(files: Sequence<BioFile>, fileList: OutputStream) {
         val writer = fileList.bufferedWriter()
-        writeHeaderNames(files.first(), writer)
+        processFirstFile(files.first(), writer)
         files.forEach { file -> writeAttributesValues(file, writer) }
         writer.close()
     }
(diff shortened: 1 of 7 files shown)


commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/tsv/deserialization/stream/FileListTsvStreamDeserializer.kt
```diff
@@ -15,7 +15,7 @@ import java.io.OutputStream
 internal class FileListTsvStreamDeserializer {
     fun serializeFileList(files: Sequence<BioFile>, fileList: OutputStream) {
         val writer = fileList.bufferedWriter()
-        writeHeaderNames(files.first(), writer)
+        processFirstFile(files.first(), writer)
```
Juan-EBI, line 18: Something does not look correct, if you notice we are writing two times now the attributes values of the first file. Once in line 35 and another in line 19. That probablythe reason unit test are failing.

jhoanmanuelms: I think the problem was coming from the `.first()` method taking out the first element from the sequence. Then, when it starts iterating, it'll miss it since it's already processed

commons/commons-serialization/src/test/kotlin/ac/uk/ebi/biostd/tsv/FileListTsvStreamDeserializerTest.kt
```diff
@@ -55,16 +55,26 @@ class FileListTsvStreamDeserializerTest(
...
     fun `serialize - deserialize FileList`() {
-        val files = (1..20_000).map {
+        fun bioFile(it: Int) =
             BioFile("folder/file$it.txt", attributes = listOf(Attribute("Attr1", "A$it"), Attribute("Attr2", "B$it")))
-        }.asSequence()
-        val iterator = files.iterator()
+
+        var idx = -1
+        val files = object : Sequence<BioFile> {
+            override fun iterator(): Iterator<BioFile> = object : Iterator<BioFile> {
```
Juan-EBI, line 63: this way of creating a sequence looks overcomplex, why we have change previos approach (create a list and use asSequence).  You may change https://kotlinlang.org/docs/sequences.html#construct there they mention 3 recomended way to construct a sequence. 

1. From elements
2. From iterable (previos appraoch)
3. From a function
4. From chunks﻿

All this IMO looks way simpler that new approach.

Juan-EBI (approved):
LGTM

PR #572, opened 2022-06-03
Pivotal ID # 182371709: Remove submission status

Juan-EBI:
https://www.pivotaltracker.com/story/show/182371709

- Removed submission status as it is always "processed" when submission is in submission collection and it is always processing when exists a request that has not been completed.


jhoanmanuelms (approved):
LGTM

PR #578, opened 2022-06-27
Pivotal ID # 180178872: ORCID is ignored

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/180178872

- Rename auxProfile column to orcid
- Store/fetch ORCID in the user profile


infrastructure/src/main/resources/setup/mysql/Schema.sql
```diff
@@ -47,7 +47,7 @@ CREATE TABLE User
     id                   BIGINT AUTO_INCREMENT PRIMARY KEY,
     activationKey        VARCHAR(255) NULL,
     active               BIT    NOT NULL,
-    auxProfileInfo       LONGTEXT NULL,
+    orcid                LONGTEXT NULL,
```
Juan-EBI, line 50: longtext does not look appropiated for this. VARCHAR(255) should be enought.

Juan-EBI (approved):
LGTM please check comment, also we need to migrate old data before releasing this one. Maybe we can have script ready.

PR #579, opened 2022-06-28
Pivotal ID # 180490482: Include Root Path File Not Found Error

Juan-EBI:
https://www.pivotaltracker.com/story/show/180490482
- improved file missing error message.


submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/helpers/SourceGenerator.kt
```diff
@@ -59,16 +63,18 @@ class SourceGenerator(
 
             if (sub != null) {
                 add(fireSourceFactory.createSubmissionFireSource(sub.accNo, Paths.get("${sub.relPath}/Files")))
-                add(PathFilesSource(Paths.get(props.submissionPath).resolve(sub.relPath).resolve(FILES_PATH)))
+
+                val subPath = Paths.get(props.submissionPath).resolve("${sub.relPath}/$FILES_PATH")
+                add(PathSource("Submission ${sub.accNo} previous version files", subPath))
```
jhoanmanuelms, line 68: Since we already have the submission and thinking that we'll support history, I think it might add value to include the version as well so it's more specific

`Files for submission ${sub.accNo} version ${sub.version}`

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/helpers/FireFilesSource.kt
```diff
@@ -66,6 +64,8 @@ private class SubmissionFireFilesSource(
     override fun getFile(path: String, md5: String?): File? =
         if (md5 == null) fireClient.downloadByPath(subPath.resolve(path).toString())
         else fireClient.downloadByMd5(md5)
+
+    override fun description(): String = "Submission $accNo files"
```
jhoanmanuelms, line 68: Perhaps we could include the version as well?

jhoanmanuelms (commented):
Looks good in general, just a couple of comments

commons/commons-bio/src/main/kotlin/ebi/ac/uk/io/sources/FilesSources.kt
```diff
@@ -4,10 +4,10 @@ import ebi.ac.uk.extended.model.ExtFile
 import ebi.ac.uk.model.Attribute
 import java.io.File
 
-class ComposedFileSource(private val sources: List<FilesSource>) : FilesSource {
```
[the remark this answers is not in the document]
Juan-EBI: basically, because this was hidden the fact that actually, we have many sources. I notice this abstraction issue as when I was going to trigger an error only one source was able which does not represent correctly business case.

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/helpers/FireFilesSource.kt
```diff
@@ -42,6 +38,8 @@ class FireFilesSource(
 
     override fun getFile(path: String, md5: String?): File? =
         if (md5 == null) null else fireClient.downloadByMd5(md5)
+
+    override fun description(): String = "EBI internal files Archive"
```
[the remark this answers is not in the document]
Juan-EBI: Because error is for users and users does not know about FIRE

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/helpers/SourceGenerator.kt
Juan-EBI: This error is for users and they are not aware of the concept "version". That is the reason I decide to keep in generic.

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/helpers/FireFilesSource.kt
Juan-EBI: Same as above.

jhoanmanuelms (approved):
LGTM

PR #581, opened 2022-07-07
Pivotal ID # 182664544: PMC importer improvements

Juan-EBI:
https://www.pivotaltracker.com/story/show/182664544
- Added debug support
- Upgraded PMC import task dependencies
- Favor runCatching over arrow try
- Added string converter to solve slack message api error 
- Added java home system property
- Add auto-deploy for the scheduler application and tasks for prod branch


jhoanmanuelms (approved):
LGTM

PR #583, opened 2022-07-13
Pivotal ID # 182714936: Avoid Fire Metadata endpoint

Juan-EBI:
https://www.pivotaltracker.com/story/show/182714936

### DIFF
--- a/submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/fire/FireFilesService.kt
+++ b/submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/fire/FireFilesService.kt
@@ -6,6 +6,7 @@ import ebi.ac.uk.extended.model.ExtFile
 import ebi.ac.uk.extended.model.ExtSubmission
 import ebi.ac.uk.extended.model.FireFile
 import mu.KotlinLogging
+import uk.ac.ebi.extended.serialization.service.FileIteratorService
 import uk.ac.ebi.extended.serialization.service.FileProcessingService
 
 private val logger = KotlinLogging.logger {}
@@ -13,15 +14,21 @@ private val logger = KotlinLogging.logger {}
 class FireFilesService(
     private val fireService: FireService,
     private val fileProcessingService: FileProcessingService,
+    private val fileIteratorService: FileIteratorService,
 ) : FilesService {
-    override fun persistSubmissionFiles(request: FilePersistenceRequest): ExtSubmission {
-        val sub = request.submission
+    override fun persistSubmissionFiles(rqt: FilePersistenceRequest): ExtSubmission = processFiles(rqt.submission)
+    override fun cleanSubmissionFiles(sub: ExtSubmission) = cleanPreviousFiles(sub)
 
+    private fun cleanPreviousFiles(sub: ExtSubmission) {
+        fun cleanFile(file: ExtFile, index: Int) {
+            logger.debug { "${sub.accNo}, ${sub.version} Cleaning file $index, path='${file.filePath}'" }
+            fireService.cleanFile(file)
+            logger.debug { "${sub.accNo}, ${sub.version} Cleaning file $index, path='${file.filePath}'" }
+        }
+
+        logger.info { "${sub.accNo} ${sub.owner} Cleaning Current submission Folder for ${sub.accNo}" }
+        fileIteratorService.forEachFile(sub) { file, index -> cleanFile(file, index) }
         logger.info { "${sub.accNo} ${sub.owner} Cleaning Ftp Folder for ${sub.accNo}" }
-        fireService.cleanFtp(sub)
-        logger.info { "${sub.accNo} ${sub.owner} Finished Ftp Folder for ${sub.accNo}" }
-        val processed = processFiles(request.submission)
-        return processed
     }
 
     private fun processFiles(sub: ExtSubmission): ExtSubmission {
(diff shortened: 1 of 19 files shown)


submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/fire/FireFilesService.kt
```diff
@@ -6,22 +6,30 @@ import ebi.ac.uk.extended.model.ExtFile
...
     private val fileProcessingService: FileProcessingService,
+    private val serializationService: ExtSerializationService,
 ) : FilesService {
-    override fun persistSubmissionFiles(request: FilePersistenceRequest): ExtSubmission {
-        val sub = request.submission
+    override fun persistSubmissionFiles(rqt: FilePersistenceRequest): ExtSubmission = processFiles(rqt.submission)
+    override fun cleanSubmissionFiles(sub: ExtSubmission) = cleanPreviousFiles(sub)
 
+    private fun cleanPreviousFiles(sub: ExtSubmission) {
+        fun cleanFile(file: ExtFile, index: Int) {
```
jhoanmanuelms, line 24: Is this a standard we'll start using? Defining functions as variables instead of private methods?

submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/fire/FireFtpService.kt
```diff
@@ -23,36 +24,16 @@ class FireFtpService(
...
 
-    private fun cleanFtpFolder(accNo: String) {
-        fireClient
-            .findByAccNoAndPublished(accNo, true)
-            .forEach { unPublishFile(it.fireOid) }
-    }
-
-    private fun publishFile(fireId: String) {
+    private fun publishFile(sub: ExtSubmission, fireId: String, index: Int) {
+        logger.debug { "${sub.accNo}, ${sub.version} publishing file $index, fireId='$fireId'" }
```
jhoanmanuelms, line 36: I think the agreed format for the logs was accNo/owner. That is if we're still using the kibana thing

submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/fire/FireService.kt
```diff
@@ -21,14 +18,11 @@ class FireService(
...
-
-    fun cleanFtp(sub: ExtSubmission) {
-        logger.info { "Started cleaning FTP files (unset path) of files in submission ${sub.accNo}" }
-        client
-            .findByAccNo(sub.accNo)
-            .apply { logger.info { "Found $size files in submission ${sub.accNo}" } }
-            .forEach { client.unsetPath(it.fireOid) }
-        logger.info { "Finished cleaning FTP files (unset path) of files in submission ${sub.accNo}" }
+    fun cleanFile(file: ExtFile) {
+        if (file is FireFile) {
```
jhoanmanuelms, line 22: Shouldn't we filter the files before passing them to the function or maybe ask this question before? I just think this hides the fact that it won't do anything if the file is NfsFile so if I see a method that gets an ExtFile called cleanFile I'd expect it would clean both types

submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/fire/FireService.kt
```diff
@@ -40,40 +34,38 @@ class FireService(
...
-                null -> return FirePersistResult(setMetadata(sub, fireFile.fireOid, file, expectedPath), false)
-                expectedPath -> return FirePersistResult(asFireFile(file, fireFile.fireOid), false)
-            }
-        }
+        val files = client.findByMd5(file.md5)
+
+        val byPath = files.firstOrNull { it.filesystemEntry?.path == expectedPath }
+        if (byPath != null) return FirePersistResult(asFireFile(file, byPath.fireOid), false)
+
+        val noPath = files.firstOrNull { it.filesystemEntry?.path == null }
```
jhoanmanuelms, line 61: I think this one should go first since it's the most likely to happen because it's supposed that we cleaned everything in the previous step

jhoanmanuelms (approved):
LGTM. I left a couple of comments

submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/fire/FireFilesService.kt
```diff
@@ -6,22 +6,30 @@ import ebi.ac.uk.extended.model.ExtFile
...
+            logger.debug { "${sub.accNo}, ${sub.version} Cleaning file $index, path='${file.filePath}'" }
+            fireService.cleanFile(file)
```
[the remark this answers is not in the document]
Juan-EBI: yes agree

submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/fire/FireFilesService.kt
Juan-EBI: in this case I found then usefull as they can access the sub paramter in main function and we keep function signature key.

PR #586, opened 2022-07-20
Pivotal ID # 182758404: Multiple File References

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/182758404

- Find fire files including the first slash in the path to match fire format
- Update the fire wiremock to match the way FIRE handles paths
- Include the exporter artifact in the build
- Improve the cleaning logging to specify the correct version for which the files are being cleaned


submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/wiremock/FireMockDatabase.kt
```diff
@@ -28,55 +30,56 @@ class FireMockDatabase(
...
-        if (recordsById.values.firstOrNull { it.path == path } != null)
-            throw FireException("Path '$path' has already a file", BAD_REQUEST)
+        val normalizedPath = normalizePath(path)
+        val firePath = "/$normalizedPath"
+        if (recordsById.values.firstOrNull { it.path == firePath } != null)
+            throw FireException("Path '$firePath' is already allocated", CONFLICT)
 
         val record = recordsById.getValue(fireOid)
-        recordsById[fireOid] = record.copy(path = path)
+        recordsById[fireOid] = record.copy(path = firePath)
```
Juan-EBI, line 39: it looks to me that we can simple save normalized path instead of firepath and then we do not need to denormalize in every operation.

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/request/SubmissionRequestProcessor.kt
```diff
@@ -15,13 +15,13 @@ class SubmissionRequestProcessor(
 ) {
 
     fun processRequest(saveRequest: ProcessedSubmissionRequest): ExtSubmission {
-        val (sub, fileMode, draftKey) = saveRequest
+        val (sub, fileMode, draftKey, previousVersion) = saveRequest
```
Juan-EBI, line 18: nice

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/wiremock/FireMockDatabase.kt
```diff
@@ -101,6 +104,8 @@ class FireMockDatabase(
     fun getFile(fireOid: String): File {
         return dbFolder.resolve(fireOid).toFile()
     }
+
+    private fun normalizePath(path: String) = "(/*)(.*)".toPattern().match(path)!!.secondGroup()
```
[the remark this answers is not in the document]
jhoanmanuelms: We can't know if someone send us more than one slash, that's why I went with this but we could stick to prefix

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/wiremock/FireMockDatabase.kt
jhoanmanuelms: We still would need to add the extra slash when looking though

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/wiremock/FireMockDatabase.kt
```diff
@@ -28,55 +30,56 @@ class FireMockDatabase(
     }
 
     fun setPath(fireOid: String, path: String) {
-        if (recordsById.values.firstOrNull { it.path == path } != null)
-            throw FireException("Path '$path' has already a file", BAD_REQUEST)
+        val normalizedPath = normalizePath(path)
```
Juan-EBI, line 33: According to my understading fire will change the path of 

```
1. /my/path/file.txt
2. my/path/file.txt
3. ///my/path/file.txt
4. /////////my/path/file.txt
```

To `/my/path/file.txt` so if we try to search by any of the previos paths fire will not find file but if we try to set it will. We need to represent the same in the mock. so We can

1.  Save file path and system path when seting the path. Perform all search operation by path and use system path for disk operation.
2. OR Create FireFileSystem that perform operation using path (removing prefix / to make resolve cllas or)

one way or anothe only path operation should be affected as in Fire.

Juan-EBI (approved):

PR #587, opened 2022-07-21
Pivotal ID # 182720666: Inclue ALL File Sources In Files List Validation

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/182720666

- Use all the submission file resources in order to validate the file list
- Add a parameter that allows to specify an accession number to look for files
- Add integration tests
- Remove unused code


client/bio-commandline/src/main/kotlin/uk/ac/ebi/biostd/client/cli/commands/ValidateFileListCommand.kt
```diff
@@ -3,24 +3,31 @@ package uk.ac.ebi.biostd.client.cli.commands
...
-    private val user by option("-u", "--user", help = CommonParameters.USER_HELP).required()
-    private val password by option("-p", "--password", help = CommonParameters.PASSWORD_HELP).required()
-    private val onBehalf by option("-b", "--onBehalf", help = CommonParameters.ON_BEHALF_HELP)
-    private val fileListPath by option("-f", "--fileListPath", help = SubmissionParameters.FILE_LIST_PATH).required()
+    private val server by option("-s", "--server", help = SERVER_HELP).required()
+    private val user by option("-u", "--user", help = USER_HELP).required()
+    private val password by option("-p", "--password", help = PASSWORD_HELP).required()
+    private val onBehalf by option("-b", "--onBehalf", help = ON_BEHALF_HELP)
+    private val fileListPath by option("-f", "--fileListPath", help = FILE_LIST_PATH).required()
+    private val previousVersion by option("-pv", "--previousVersion", help = PREVIOUS_VERSION)
```
Juan-EBI, line 25: I think this (and http parameter) should be only accNo, which means we are validation file list inside an specific submission.

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/web/resources/FileListResource.kt
```diff
@@ -18,16 +22,27 @@ class FileListResource(
     private val sourceGenerator: SourceGenerator,
     private val fileListValidator: FileListValidator,
     private val onBehalfUtils: OnBehalfUtils,
+    private val submissionQueryService: SubmissionPersistenceQueryService,
```
Juan-EBI, line 25: calling persistence layer directly from web layer is not good.  IMO we should include all this in FileListValidator which is the service/domain class in charge of this operation.

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/FileListValidationTest.kt
```diff
@@ -72,13 +82,94 @@ class FileListValidationTest(
...
-        webClient.deleteFile(fileListFile.name)
+            line("File", "Plate2.tif")
+            line()
+        }.toString()
+
+        val file1 = tempFolder.createFile("Plate1.tif", "content-1")
+        val file2 = tempFolder.createFile("Plate2.tif", "content-2")
+        val fileListContent = tsv {
+            line("Files", "Resource")
+            line("Plate1.tif", "USER_SPACE")
```
Juan-EBI, line 100: what is the point of USER_SPACE here? SUBMISSION? this looks like a feature of the system or something like that when is a simple attribute. 

(please consider same comment for other test)

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/FileListValidationTest.kt
```diff
@@ -72,13 +82,94 @@ class FileListValidationTest(
...
-        webClient.validateFileList(fileList.name)
+            line("Study")
+            line()
 
-        webClient.deleteFile(fileListFile.name)
+            line("File", "Plate2.tif")
+            line()
+        }.toString()
+
+        val file1 = tempFolder.createFile("Plate1.tif", "content-1")
```
Juan-EBI, line 96: personally I found easier (less cognitive load) file1.txt, file2.txt and like that.

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/FileListValidationTest.kt
```diff
@@ -72,13 +82,94 @@ class FileListValidationTest(
...
+            line()
+
+            line("Study")
+            line()
+
+            line("File", "Plate5.tif")
+            line()
+        }.toString()
+
+        val file3 = tempFolder.createFile("Plate3.tif", "content-3")
```
Juan-EBI, line 130: any reason to start with the file number 3? super small detail but I think test will be way simpler if file list have a single file and is in Fire. The simpler and more focus we design the scenarios the better. 

For a complex scenarios mixed, we have the allIncluded test  (please consider same comment for other test)

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/FileListValidationTest.kt
```diff
@@ -90,19 +181,34 @@ class FileListValidationTest(
...
                           - Plate1.tif
                         List of available sources:
                           - biostudies-mgmt-filelist-v@ebi.ac.uk user files
+                          - EBI internal files Archive
                 """.trimIndent()
                 "subnodes" to jsonArray()
             }
             "status" to "FAIL"
         }
+        val expectedError = if (enableFire) expectedFireError else expectedNfsError
```
Juan-EBI, line 211: I would rather create two test, `file list with missing files when NFs` and `file list with missing files When fire`, conditional checks in test code just make it hard to follow it. (So in this case I think is better to favor simplicity over code re use)

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/validator/filelist/FileListValidator.kt
```diff
@@ -24,7 +25,7 @@ class FileListValidator(
     private fun validateFileList(stream: InputStream, format: SubFormat, filesSource: FileSourcesList) {
         serializationService
             .deserializeFileList(stream, format)
-            .filter { filesSource.getExtFile(it.path) == null }
+            .filter { filesSource.getExtFile(it.path, it.md5) == null }
```
Juan-EBI, line 28: nice catch

Juan-EBI (commented):
looking good, please check comments

PR #588, opened 2022-07-21
Pivotal ID # 182221587: Upgrade dependencies

Juan-EBI:
https://www.pivotaltracker.com/story/show/182221587

### DIFF
--- a/events/events-publisher/build.gradle.kts
+++ b/events/events-publisher/build.gradle.kts
@@ -2,17 +2,30 @@ import Dependencies.KotlinReflect
 import Dependencies.KotlinStdLib
 import Projects.CommonsBio
 import Projects.CommonsModelExtended
-import SpringBootDependencies.SpringBootAmqp
+import SpringBootDependencies.SpringBootStarterAmqp
 import TestDependencies.BaseTestCompileDependencies
 import TestDependencies.BaseTestRuntimeDependencies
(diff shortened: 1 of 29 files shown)


events/events-publisher/build.gradle.kts
```diff
@@ -2,17 +2,30 @@ import Dependencies.KotlinReflect
...
+import org.springframework.boot.gradle.plugin.SpringBootPlugin
+
+plugins {
+    id("io.spring.dependency-management") version "1.0.12.RELEASE"
+    id("org.springframework.boot") version "2.7.1" apply false
+}
+
+the<DependencyManagementExtension>().apply {
+    imports {
+        mavenBom(SpringBootPlugin.BOM_COORDINATES)
```
jhoanmanuelms, line 18: We should avoid full imports here to keep consistency

jhoanmanuelms:
The mysql driver name we're using in our configuration files is deprecated and it should be changed from `com.mysql.jdbc.Driver` to `com.mysql.cj.jdbc.Driver`

Since this PR is about updating dependencies, I think it's a good place to include it:

https://github.com/EBIBioStudies/biostudies-backend-services/blob/master/submission/submission-webapp/src/main/resources/application.yml#L21

https://github.com/EBIBioStudies/biostudies-backend-services/blob/master/submission/submission-handlers/src/main/resources/application.yml#L11

PR #590, opened 2022-07-24
Pivotal ID # 182800115: Folder Submissions handling

Juan-EBI:
https://www.pivotaltracker.com/story/show/182800115
- handled 404 when not files found by md5.
- added zip file handling logic.

### DIFF
--- a/client/fire-webclient/src/main/kotlin/uk/ac/ebi/fire/client/api/FireWebClient.kt
+++ b/client/fire-webclient/src/main/kotlin/uk/ac/ebi/fire/client/api/FireWebClient.kt
@@ -64,7 +64,7 @@ internal class FireWebClient(
     }
 
     override fun findByMd5(md5: String): List<FireApiFile> =
-        template.getForObjectOrNull<Array<FireApiFile>>("/objects/md5/$md5").orEmpty().toList()
+        template.getForObject<Array<FireApiFile>>("/objects/md5/$md5").toList()
 
     override fun findByPath(path: String): FireApiFile? = template.getForObjectOrNull("/objects/path/$path")
 
--- a/client/fire-webclient/src/main/kotlin/uk/ac/ebi/fire/client/api/FireWebClient.kt
+++ b/client/fire-webclient/src/main/kotlin/uk/ac/ebi/fire/client/api/FireWebClient.kt
@@ -6,6 +6,7 @@ import org.springframework.http.HttpHeaders
 import org.springframework.http.HttpStatus.NOT_FOUND
 import org.springframework.util.LinkedMultiValueMap
 import org.springframework.web.client.HttpClientErrorException
+import org.springframework.web.client.RestOperations
 import org.springframework.web.client.RestTemplate
 import org.springframework.web.client.getForObject
 import org.springframework.web.client.postForObject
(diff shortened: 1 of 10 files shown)


client/fire-webclient/src/main/kotlin/uk/ac/ebi/fire/client/api/FireWebClient.kt
```diff
@@ -63,17 +64,12 @@ internal class FireWebClient(
     }
 
     override fun findByMd5(md5: String): List<FireApiFile> =
-        template.getForObject<Array<FireApiFile>>("/objects/md5/$md5").toList()
+        template.getForObjectOrNull<Array<FireApiFile>>("/objects/md5/$md5").orEmpty().toList()
```
jhoanmanuelms, line 67: Here we're actually calling nonexisting endpoints which are giving us the 404. For example, if the md5 is empty we would hit `/objects/md5/` which doesn't exist. Maybe we should not call the endpoints when the values are empty and just return an empty list instead of handling the 404 for a call we know will fail? 🤔

submission/persistence-filesystem/build.gradle.kts
```diff
@@ -25,6 +25,7 @@ dependencies {
     api(project(SubmissionPersistenceCommonApi))
 
     implementation(Arrow)
+    implementation("commons-codec:commons-codec:1.15")
```
jhoanmanuelms, line 28: We should add a dependency in `Dependencies.kt`

submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/fire/FireService.kt
```diff
@@ -29,21 +33,47 @@ class FireService(
      * different path) and if so, even if the file exists in FIRE, it gets duplicated to ensure consistency.
      */
     fun getOrPersist(sub: ExtSubmission, file: ExtFile): FirePersistResult {
-        val expectedPath = "/${sub.relPath}/${file.relPath}"
```
jhoanmanuelms, line 32: I had made this change because the paths are the same in both places which I think makes clear that no matter where the file is processed, we use the same path

submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/fire/FireService.kt
```diff
@@ -29,21 +33,47 @@ class FireService(
...
+            val tempFolder = fireTempDirPath.resolve("${sub.accNo}/${sub.version}")
+            tempFolder.mkdirs()
+
+            val target = tempFolder.resolve("${file.name}.zip")
+            Files.deleteIfExists(target.toPath())
+            ZipUtil.pack(file, target, true)
+            return target
+        }
+
+        fun calculateMd5(dir: File): String {
```
jhoanmanuelms, line 60: Since we're always compressing the directory before uploading it to fire, why do we need to calculate the md5 in this way? Or is this the same value as the compressed directory?

I'm just asking because I'm not sure fire will take this md5 for the upload as valid and we'll need to recalculate it anyway and also, if we're trying to find it by md5 in order to reuse it, I'm not sure if this md5 will do

client/fire-webclient/src/main/kotlin/uk/ac/ebi/fire/client/api/FireWebClient.kt
Juan-EBI: error on terminal for me , was md5 not found.

submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/fire/FireService.kt
Juan-EBI: yes the problem is that this is not the path when file is a folder. As we need to add the .zip that is the reason I move the call after compressing the file and creating extDirectory. 

I did update the code to have single declaration please take a look again.

client/fire-webclient/src/main/kotlin/uk/ac/ebi/fire/client/api/FireWebClient.kt
jhoanmanuelms: Yeah, I thought it was coming from the standard http error but apparently they decided this design but, at least, I think we could save one request since we know for sure empty md5 or path will not return any result don't you think?

submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/fire/ZipUtil.kt
```diff
@@ -0,0 +1,34 @@
...
+import java.io.File
+import java.nio.file.Files
+import java.nio.file.Path
+import java.util.zip.ZipEntry
+import java.util.zip.ZipOutputStream
+import kotlin.io.path.isDirectory
+
+object ZipUtil {
+
+    fun pack(sourceDir: File, zipFile: File) {
```
jhoanmanuelms, line 12: 👍

jhoanmanuelms (approved):
LGTM

PR #597, opened 2022-08-04
Pivotal ID # 182829591: Submissions List Order

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/182829591

- Fix sort step in the submissions aggregation
- Remove grouping from the aggregation since it's no longer required
- Remove unused methods and properties
- Fix typos

### DIFF
--- a/submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/db/data/SubmissionDocDataRepository.kt
+++ b/submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/db/data/SubmissionDocDataRepository.kt
@@ -24,12 +24,10 @@ import org.springframework.data.domain.PageImpl
 import org.springframework.data.domain.PageRequest
 import org.springframework.data.domain.Sort
 import org.springframework.data.mongodb.core.MongoTemplate
-import org.springframework.data.mongodb.core.aggregation.Aggregation.CURRENT
 import org.springframework.data.mongodb.core.aggregation.Aggregation.group
 import org.springframework.data.mongodb.core.aggregation.Aggregation.limit
 import org.springframework.data.mongodb.core.aggregation.Aggregation.match
 import org.springframework.data.mongodb.core.aggregation.Aggregation.newAggregation
-import org.springframework.data.mongodb.core.aggregation.Aggregation.replaceRoot
 import org.springframework.data.mongodb.core.aggregation.Aggregation.skip
 import org.springframework.data.mongodb.core.aggregation.Aggregation.sort
 import org.springframework.data.mongodb.core.aggregation.AggregationOperation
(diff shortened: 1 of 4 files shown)


submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/db/data/SubmissionDocDataRepository.kt
```diff
@@ -141,14 +116,10 @@ class SubmissionDocDataRepository(
...
-                add(match(where(SUB_VERSION).gt(0).andOperator(*createQuery(filter, email))))
-                sort(Sort.Direction.DESC, SUB_MODIFICATION_TIME)
-                offsetLimit?.let { add(skip(it.first)); add(limit(it.second)) }
-                add(group(SUB_ACC_NO).first(SUB_VERSION).`as`("maxVersion").first(CURRENT).`as`(SUB_ALIAS))
-                add(replaceRoot(SUB_ALIAS))
-            }
+        ): List<AggregationOperation> = buildList {
+            add(match(where(SUB_VERSION).gt(0).andOperator(*createQuery(filter, email))))
+            add(sort(Sort.Direction.DESC, SUB_MODIFICATION_TIME))
+            offsetLimit?.let { add(skip(it.first)); add(limit(it.second)) }
```
Juan-EBI, line 122: sorry is not clear for me why we are removing the agrupation. can you explain

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/query/SubmissionListApiTest.kt
```diff
@@ -228,6 +228,17 @@ class SubmissionListApiTest(
         }
     }
 
+    @Test
+    fun `latest updated submission should appear first`() {
+        webClient.submitSingle(getSimpleSubmission(19), TSV)
```
Juan-EBI, line 233: I do not see the point of creating 19 submissions and then asserting only the first one. It will be a much better test if we create 3 and verity they are retrrieved in that order.

Juan-EBI (approved):
LGTM please check comments

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/query/SubmissionListApiTest.kt
jhoanmanuelms: We're already creating them for all the other tests. The aim of this one is to validate that the latest modified, submission 19, which is already created, is always the first appearance in the list.

It might as well be submission 2 or 24. The result should be the same

Juan-EBI: yes exactly what is the point of testing 19 usbmission?.  I think the idea behind test is to have the **simplest**  posible scenario that test what we require.  Using 19 submission instead of 2, or using 10 files instead of two 2 when they do nothing  actually does not help but rather make test harder to follow and harder to understand. 

why having more complex when simple do the job? same principle for me apply for attributes, sections, user. In general I would rather favor simplicity. Now if we are coding a performance test or a mongo documents limits test story is completly diferent but this is not what we are doing here.

jhoanmanuelms: I think the reason to have 20 submissions in this test is to validate the default limits and offsets we currently have. We could set a custom limit for the test but, I think the idea of the itests is to keep it as similar as possible to the actual use case.

Juan-EBI (approved):

PR #599, opened 2022-08-09
Pivotal ID #182583027: PMC Export Empty Files

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/182583027

The empty files were caused by the FTP connection timing out. In order to solve this, two measures were taken:
- Reconnect and retry after failure
- Remove paralelism in order to avoid multiple connections which might be killed


scheduler/tasks/exporter-task/src/main/kotlin/uk/ac/ebi/scheduler/pmc/exporter/service/PmcExporterService.kt
```diff
@@ -27,40 +24,55 @@ class PmcExporterService(
...
-        logger.info { "writing file part $part" }
-        val xml = xmlWriter.writeValueAsString(Links(links))
+        logger.info { "Writing file part $part" }
+
+        val xml = xmlWriter.writeValueAsString(Links(links)).byteInputStream()
         val path = "${appProperties.outputPath}/${String.format(appProperties.fileName, part)}"
-        ftpClient.storeFile(path, xml.byteInputStream())
+
+        try {
+            ftpClient.storeFile(path, xml)
```
Juan-EBI, line 57: I would suggest writing a wrapper class over apache FTPClient, like our own class that handles retry so we can keep the logic here clean here.  (hight specialization)

Juan-EBI (approved):
LGTM please consider changes before merge

PR #604, opened 2022-08-22
Pivotal ID # 182987585: Per submission storage mode

Juan-EBI:
https://www.pivotaltracker.com/story/show/182987585
- Improved SubmitWebHandler api
- Favor use of list instead of array
- Favor use of nullables over empty collections

### DIFF
--- a/submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/web/handlers/SubmitWebHandler.kt
+++ b/submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/web/handlers/SubmitWebHandler.kt
@@ -63,9 +63,9 @@ class SubmitWebHandler(
         val onBehalfUser = rqt.onBehalfRequest?.let { onBehalfUtils.getOnBehalfUser(it) }
 
         /**
-         * Deserialize the submission without considering files.
+         * Deserialize the submission without considering files and retrieve accNo and rootPath.
          */
-        fun submissionAttributes(): Pair<String, String?> {
+        fun deserializeSubmission(): Pair<String, String?> {
             val submission = when (rqt) {
                 is ContentSubmitWebRequest -> serializationService.deserializeSubmission(rqt.submission, rqt.format)
                 is FileSubmitWebRequest -> serializationService.deserializeSubmission(rqt.submission)
@@ -102,7 +102,7 @@ class SubmitWebHandler(
          * 4. Overridden attributes are set.
          */
         fun processSubmission(): Pair<Submission, FileSourcesList> {
-            val (accNo, rootPath) = submissionAttributes()
+            val (accNo, rootPath) = deserializeSubmission()
             require(extSubService.hasPendingRequest(accNo).not()) { throw ConcurrentSubException(accNo) }
 
             val previous = extSubService.findExtendedSubmission(accNo)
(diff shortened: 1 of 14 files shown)


submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/web/handlers/SubmitWebHandler.kt
```diff
@@ -2,122 +2,116 @@ package ac.uk.ebi.biostd.submission.web.handlers
...
-    ) = serializationService.deserializeSubmission(subFile, source)
+    private fun buildRequest(rqt: SubmitWebRequest): SubmitRequest {
+        val (submitter, attrs) = rqt.config
+        val (files, preferredSources) = rqt.filesConfig
+        val onBehalfUser = rqt.onBehalfRequest?.let { onBehalfUtils.getOnBehalfUser(it) }
+
+        /**
+         * Deserialize the submission without considering files.
+         */
+        fun submissionAttributes(): Pair<String, String?> {
```
jhoanmanuelms, line 68: I think the "without considering files" part of the doc for this function is a bit confusing I would suggest: "Deserialize the submission to get the accNo and rootPath attributes. This deserialization doesn't consider the files since they're not needed yet", something like that

I also think, after reviewing this piece of code, that maybe we should improve the names in the serialization service instead of overloading the methods which would make them clearer.

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/web/handlers/SubmitWebHandler.kt
```diff
@@ -2,122 +2,116 @@ package ac.uk.ebi.biostd.submission.web.handlers
...
-        throw ConcurrentProcessingSubmissionException(accNo)
+        /**
+         * Process the given submission:
+         *
+         * 1. AccNo, RootPath attributes are extracted from Submission.
+         * 2. Submission file sources are obtained.
+         * 3. Submission is deserialized including file sources to check both pagetab structure and file presence.
+         * 4. Overridden attributes are set.
+         */
+        fun processSubmission(): Pair<Submission, FileSourcesList> {
```
jhoanmanuelms, line 104: Way to go man, it's much clearer now 👏

jhoanmanuelms (approved):
LGTM. Please consider my comments before merging

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/web/handlers/SubmitWebHandler.kt
```diff
@@ -2,122 +2,116 @@ package ac.uk.ebi.biostd.submission.web.handlers
...
+        val rqt = buildRequest(request)
+        userFilesService.uploadFile(request.config.submitter, DIRECT_UPLOAD_PATH, request.submission)
+        subService.submitAsync(rqt)
+    }
 
-    private fun submission(
-        subFile: File,
-        source: FileSourcesList
-    ) = serializationService.deserializeSubmission(subFile, source)
+    private fun buildRequest(rqt: SubmitWebRequest): SubmitRequest {
```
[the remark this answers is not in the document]
Juan-EBI: agree but want to keep this improvement as only internal and do not change classes structure, as this is only a prepreation ticket for storage mode parameter.

PR #605, opened 2022-08-22
Pivotal ID # 183046110: Scheduler sch config

Juan-EBI:
https://www.pivotaltracker.com/story/show/183046110


jhoanmanuelms (approved):
LGTM. Just a quick comment, in this case, we would no longer need the password in the scheduler properties file right?

https://github.com/EBIBioStudies/biostudies-backend-services/blob/master/scheduler/scheduler/src/main/resources/application.yml#L13

PR #608, opened 2022-08-23
Pivotal ID # 183058373: Releaser not running

Juan-EBI:
https://www.pivotaltracker.com/story/show/183058373
- removed unecesary dependencies confusing spring boot and making it fail due a missing sql datasource.


jhoanmanuelms (approved):
LGTM. Nice catch!

PR #609, opened 2022-08-24
Pivotal ID # 182987585: Per submission storage mode

Juan-EBI:
https://www.pivotaltracker.com/story/show/182987585
- added storage mode parameter
- refactor submission submitter


jhoanmanuelms (commented):
LGTM. The only thing I couldn't find is where is the parameter being passed in the endpoints. I think this part is missing, would it be a different PR?

Juan-EBI:
> LGTM. The only thing I couldn't find is where is the parameter being passed in the endpoints. I think this part is missing, would it be a different PR?

is in here https://github.com/EBIBioStudies/biostudies-backend-services/pull/609/files#diff-fe8d398aca1eded336ed08d54e59936b2cd89945e99f80bd843eadfe6b55e3e7R25 class annotated with modelattribute

jhoanmanuelms (approved):
LGTM

PR #614, opened 2022-09-12
Pivotal ID # 183218611: Block Draft During Validation Stage

jhoanmanuelms:
- Block draft during the validation process
- Reactivate the draft in case there's any issue during validation
- Improve logs

### DIFF
--- a/submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/service/ExtSubmissionRepository.kt
+++ b/submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/service/ExtSubmissionRepository.kt
@@ -19,34 +19,33 @@ class ExtSubmissionRepository(
     private val toExtSubmissionMapper: ToExtSubmissionMapper,
     private val toDocSubmissionMapper: ToDocSubmissionMapper
 ) {
-    fun saveSubmission(submission: ExtSubmission, draftKey: String?): ExtSubmission {
-        updateCurrentRecords(submission.accNo, submission.owner, submission.submitter, draftKey)
-        return toExtSubmissionMapper.toExtSubmission(saveSubmission(submission), false)
+    fun saveSubmission(submission: ExtSubmission): ExtSubmission {
+        subDataRepository.expireActiveProcessedVersions(submission.accNo)
+        return toExtSubmissionMapper.toExtSubmission(persistSubmission(submission), false)
     }
 
-    private fun updateCurrentRecords(accNo: String, owner: String, submitter: String, draftKey: String?) {
-        subDataRepository.expireActiveProcessedVersions(accNo)
-        deleteSubmissionDrafts(accNo, owner, submitter, draftKey)
-    }
-
-    private fun deleteSubmissionDrafts(accNo: String, owner: String, submitter: String, draftKey: String?) {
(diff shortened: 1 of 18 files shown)


submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/service/ExtSubmissionRepository.kt
```diff
@@ -19,34 +19,33 @@ class ExtSubmissionRepository(
     private val toExtSubmissionMapper: ToExtSubmissionMapper,
     private val toDocSubmissionMapper: ToDocSubmissionMapper
 ) {
-    fun saveSubmission(submission: ExtSubmission, draftKey: String?): ExtSubmission {
-        updateCurrentRecords(submission.accNo, submission.owner, submission.submitter, draftKey)
-        return toExtSubmissionMapper.toExtSubmission(saveSubmission(submission), false)
+    fun saveSubmission(submission: ExtSubmission): ExtSubmission {
+        subDataRepository.expireActiveProcessedVersions(submission.accNo)
```
Juan-EBI, line 23: lets include TODO to we remeber to  remove this side effect from here too.

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/service/ExtSubmissionService.kt
```diff
@@ -59,7 +59,7 @@ class ExtSubmissionService(
     ) {
         logger.info { "${sub.accNo} $user Received async submit request for ext submission ${sub.accNo}" }
         val submission = processSubmission(user, sub)
-        val (accNo, version) = submissionSubmitter.createRequest(ExtSubmitRequest(submission))
+        val (accNo, version) = submissionSubmitter.saveRequest(ExtSubmitRequest(submission))
```
Juan-EBI, line 62: I think is not convinient to have this method named as save, the reason I choose to separate beetween create and update/save is that during creation we need to add records, update records necesary need to exists. Also in future we will need to create files records when creating only and not when saving.

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/SubmissionSubmitter.kt
```diff
@@ -43,14 +45,21 @@ class SubmissionSubmitter(
     }
 
     @Suppress("TooGenericExceptionCaught")
-    private fun process(rqt: SubmitRequest): ExtSubmission {
+    private fun processRequest(rqt: SubmitRequest): ExtSubmission {
```
Juan-EBI, line 48: I think delete draft call need to be here. I mean we are performing draft operation here already as setting processing status and setting complete status. What make sense to me is have all "draft" business case here and not spread around many clases.

Juan-EBI (approved):
LGTM please check minors comments.

submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/service/ExtSubmissionRepository.kt
jhoanmanuelms: I moved it to a separate method

PR #616, opened 2022-09-14
Pivotal ID # 182333701: Submissions Stages

jhoanmanuelms:
- Set proper status to the request after cleaning
- Add integration test to check the stages


Juan-EBI (approved):
LGTM

PR #618, opened 2022-09-26
Pivotal ID # 183351718: Active Requests Not Showing On Submissions List

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/183351718

- Validate the request processed status including all the submission stages
- Normalize methods naming

### DIFF
--- a/submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/db/data/SubmissionRequestDocDataRepository.kt
+++ b/submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/db/data/SubmissionRequestDocDataRepository.kt
@@ -1,17 +1,19 @@
 package ac.uk.ebi.biostd.persistence.doc.db.data
 
 import ac.uk.ebi.biostd.persistence.common.model.RequestStatus
-import ac.uk.ebi.biostd.persistence.common.model.RequestStatus.PROCESSED
+import ac.uk.ebi.biostd.persistence.common.model.RequestStatus.Companion.PROCESSING
 import ac.uk.ebi.biostd.persistence.common.request.SubmissionFilter
 import ac.uk.ebi.biostd.persistence.doc.db.converters.shared.DocAttributeFields.ATTRIBUTE_DOC_NAME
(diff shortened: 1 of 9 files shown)


submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/db/data/SubmissionRequestDocDataRepository.kt
```diff
@@ -68,7 +67,7 @@ class SubmissionRequestDocDataRepository(
 
     private fun criteriaArray(filter: SubmissionFilter): Array<Criteria> =
         ImmutableList.Builder<Criteria>().apply {
-            add(where("status").`is`(REQUESTED))
+            add(where("status").ne(PROCESSED))
```
Juan-EBI, line 70: I triick I have use in the past to avoid the same bug to happen again is to include static variable in enum so we do not forget to updated when adding new value. 

I suggest to do the same here. Create PROCESSING status set in enum and use here `in` query operator.

submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/service/SubmissionMongoPersistenceQueryService.kt
```diff
@@ -39,8 +39,8 @@ internal class SubmissionMongoPersistenceQueryService(
     override fun existByAccNoAndVersion(accNo: String, version: Int): Boolean =
         submissionRepo.existsByAccNoAndVersion(accNo, version)
 
-    override fun hasPendingRequest(accNo: String): Boolean =
-        requestRepository.existsByAccNoAndStatusIn(accNo, setOf(REQUESTED))
+    override fun hasActiveRequest(accNo: String): Boolean =
+        requestRepository.existsByAccNoAndStatusNot(accNo, RequestStatus.PROCESSED)
```
Juan-EBI, line 43: same suggestion as above.

Juan-EBI (approved):

Juan-EBI (approved):
LGTM

PR #620, opened 2022-09-30
Pivotal ID # 183324114: Decouple Request Files [POC]

jhoanmanuelms:

- Include properties in the request to track the number of files and the submission total files
- Granularize files operations
- Move the responsibility of iterating, tracking, and  processing the submission files up to the request processing level
- Simplify file persistence-related classes and improve naming
- Normalize request processing classes naming

Note: The unit tests for the request processing classes will be added in the next PR once we start actually decoupling the files


submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/request/SubmissionRequestProcessor.kt
```diff
@@ -23,13 +34,65 @@ class SubmissionRequestProcessor(
...
+        logger.info { "${sub.accNo} ${sub.owner} Started persisting submission files on NFS" }
+
+        val permissions = sub.permissions()
+        val subFolder = nfsFilesService.getOrCreateSubmissionFolder(sub, permissions.folder)
+        val targetFolder = nfsFilesService.createTempSubFolder(subFolder, sub.accNo)
+
+        val persisted = fileProcessingService.processFiles(sub) { file, idx ->
+            logger.info { "${sub.accNo}, ${sub.owner} Started persisting file $idx, path='${file.filePath}' on NFS" }
+
+            val request = NfsFilePersistenceRequest(file as NfsFile, subFolder, targetFolder, permissions)
```
Juan-EBI, line 83: not sure is correct to assume that every file is a NFS file what about if we are re submitting a fire re submission into FIRE. files will be in FIRE.

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/request/SubmissionRequestProcessor.kt
```diff
@@ -23,13 +34,65 @@ class SubmissionRequestProcessor(
...
+            val nfsFile = nfsFilesService.persistSubmissionFile(request)
+            persistenceService.updateRequestIndex(sub.accNo, sub.version, idx)
```
Juan-EBI, line 85: as we are not doing anything with this index I would rather include it when we actually have a use for it (also it will help to reduce this PR scope, size and delivery time).

submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/api/FilesService.kt
```diff
@@ -1,8 +1,28 @@
...
 
 internal interface FilesService {
-    fun persistSubmissionFiles(sub: ExtSubmission): ExtSubmission
-    fun cleanSubmissionFiles(sub: ExtSubmission)
+    fun persistSubmissionFile(request: FilePersistenceRequest): ExtFile
+
+    fun cleanSubmissionFile(file: ExtFile)
 }
+
+interface FilePersistenceRequest
```
Juan-EBI, line 14: rather than an maker interface I think what you are looking here is a sealed class.

submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/nfs/NfsFilesService.kt
```diff
@@ -1,85 +1,65 @@
...
-            if (target.notExist() && subFile.exists() && subFile.md5() == extFile.md5)
-                moveFile(subFile, target, permissions)
-            else if (target.notExist())
-                copyOrReplaceFile(file, target, permissions)
+        return file.copy(fullPath = subFile.absolutePath, file = subFile)
+    }
 
-            return extFile.copy(fullPath = subFile.absolutePath, file = subFile)
-        }
+    fun cleanSubmissionFiles(sub: ExtSubmission) {
```
Juan-EBI, line 45: wondering why others methods are an interface that we access using interface while this and getOrCreateSubmissionFolder. It looks super extrange that we comunicate to a class something as a interface (abtracting implementation) and some others using class directly (no abstracting implementation)

Juan-EBI (commented):
looking good for me

PR #622, opened 2022-10-03
Pivotal ID # 183448558: Allow Fire by passing

Juan-EBI:


submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/SubmissionFileSourceTest.kt
```diff
@@ -308,4 +308,38 @@ class SubmissionFileSourceTest(
...
+    fun `Submission bypassing fire`() {
+        val submission = tsv {
+            line("Submission", "S-FSTST6")
+            line("Title", "Sample Submission")
+            line()
+
+            line("Study")
+            line()
+
+            line("File", "DataFile5.txt")
```
jhoanmanuelms, line 323: I think the new rules for this feature should be documented on a confluence page so the imaging team can properly test/use it.

I think this document should also explain the rules we use to set the path since they'll need to know this in order to properly set the path.

jhoanmanuelms (approved):

PR #623, opened 2022-10-04

jhoanmanuelms:

- Decouple files for the releasing operation
- Move the files iteration to the release processor
- Normalize processor classes names
- Improve name for the files persistence configuration


commons/commons-util/src/main/kotlin/ebi/ac/uk/io/FileUtils.kt
```diff
@@ -146,21 +146,22 @@ internal object FileUtilsHelper {
     }
 
     fun createFolderHardLinks(
-        source: Path,
-        target: Path,
+        sourcePath: Path,
```
Juan-EBI, line 149: any reason to change names here? I mean is not that I disagree with name change but trying to figured out the reasoning.

Juan-EBI (approved):
LGTM :)

commons/commons-util/src/main/kotlin/ebi/ac/uk/io/FileUtils.kt
jhoanmanuelms: It's just to make it clearer that the source is the path where the hard link will be created and not the referenced file itself

PR #624, opened 2022-10-05

jhoanmanuelms:

- Decouple file persistence operation to be on file basis rather than submission basis
- Move up the responsibility to handle the submission files to the request processor
- Simplify fire persistence logic


submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/api/FilesService.kt
```diff
@@ -1,8 +1,14 @@
 package ac.uk.ebi.biostd.persistence.filesystem.api
 
+import ebi.ac.uk.extended.model.ExtFile
 import ebi.ac.uk.extended.model.ExtSubmission
 
 internal interface FilesService {
-    fun persistSubmissionFiles(sub: ExtSubmission): ExtSubmission
+    fun preProcessSubmissionFiles(sub: ExtSubmission): FilePersistenceConfig
```
Juan-EBI, line 7: I think siganutre of this methods should be

```
    fun preProcessSubmissionFiles(sub: ExtSubmission) {}
    fun persistSubmissionFile(sub: ExtSubmission, file: ExtFile): ExtFile
    fun postProcessSubmissionFiles(sub: ExtSubmission) {}
  ```
  
The problem that I see in current desing is that nothing is preventing for calling processing with the incorrect config type. With interfaces defenied as bellow we do not have that problem but on the other hand we need to recalculate on each file the target file. Personally I prefer safer code that "performance" code (calculating expecting folder is not heavy operation) but is open to discussion.

File: submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/nfs/NfsFilesService.kt [hunk 2]
```diff
@@ -10,30 +13,45 @@ import ebi.ac.uk.io.FileUtils.copyOrReplaceFile
...
-        logger.info { "${sub.accNo} ${sub.owner} Processing files of submission ${sub.accNo} on NFS" }
-        val processed = processAttachedFiles(sub, subFolder, createTempFolder(subFolder, sub.accNo), sub.permissions())
-        logger.info { "${sub.accNo} ${sub.owner} Finished processing files of submission ${sub.accNo} on NFS" }
-        return processed
+        return NfsFilePersistenceConfig(subFolder, targetFolder, permissions)
+    }
+
+    override fun persistSubmissionFile(file: ExtFile, config: FilePersistenceConfig): ExtFile {
+        val (subFolder, targetFolder, permissions) = config as NfsFilePersistenceConfig
+        val extFile = file as NfsFile
```
Juan-EBI, line 40: I think we should stop doing this and instead  `require(file is NfsFile) { "File processing handler should only handle Nfs File"}` in this way is super clear method and class expections or preconditions.

Juan-EBI (approved):
LGTM please check comments/suggestions

submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/api/FilesService.kt
Juan-EBI: yes exactly. I think in the pre process temp folder will be created

PR #627, opened 2022-10-07
Pivotal ID # 183324114: Decouple Request Files - Track Current File

jhoanmanuelms:

- Track current processed file
- Remove page tab operation from storage service since it's a separate concern

### DIFF
--- a/submission/persistence-common-api/src/main/kotlin/ac/uk/ebi/biostd/persistence/common/model/RequestStatus.kt
+++ b/submission/persistence-common-api/src/main/kotlin/ac/uk/ebi/biostd/persistence/common/model/RequestStatus.kt
@@ -1,8 +1,16 @@
 package ac.uk.ebi.biostd.persistence.common.model
 
 import ebi.ac.uk.extended.model.ExtSubmission
+import java.time.OffsetDateTime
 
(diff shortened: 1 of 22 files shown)


submission/persistence-common-api/src/main/kotlin/ac/uk/ebi/biostd/persistence/common/model/RequestStatus.kt
```diff
@@ -1,8 +1,16 @@
...
 
 import ebi.ac.uk.extended.model.ExtSubmission
+import java.time.OffsetDateTime
 
-data class SubmissionRequest(val submission: ExtSubmission, val draftKey: String?, val status: RequestStatus)
+data class SubmissionRequest(
+    val submission: ExtSubmission,
+    val draftKey: String?,
+    val status: RequestStatus,
+    val totalFiles: Int = 0,
```
Juan-EBI, line 10: rather than help IMO I think this default parameters (exception modificationTime one which IMO is safe to have default value) may allow us to create a request without calculating this fields which will generate an issue for its proccessing. Basically this allow to easily introduce a bug.

Juan-EBI (approved):
LGTM

PR #628, opened 2022-10-07

jhoanmanuelms:

- Create a new collection to store the submission files at request level
- Separate the submission data operations per functionality
- Create a separate service to handle the logic related to the submission requests
- Create a separate service to handle the logic related to the submission files
- Update classes and tests accordingly

### DIFF
--- a/submission/persistence-common-api/src/main/kotlin/ac/uk/ebi/biostd/persistence/common/service/SubmissionOperations.kt
+++ b/submission/persistence-common-api/src/main/kotlin/ac/uk/ebi/biostd/persistence/common/service/SubmissionOperations.kt
@@ -3,6 +3,7 @@ package ac.uk.ebi.biostd.persistence.common.service
 import ac.uk.ebi.biostd.persistence.common.model.BasicCollection
 import ac.uk.ebi.biostd.persistence.common.model.BasicSubmission
 import ac.uk.ebi.biostd.persistence.common.model.RequestStatus
+import ac.uk.ebi.biostd.persistence.common.model.SubmissionFile
 import ac.uk.ebi.biostd.persistence.common.model.SubmissionRequest
 import ac.uk.ebi.biostd.persistence.common.request.SubmissionFilter
 import ebi.ac.uk.extended.model.ExtFile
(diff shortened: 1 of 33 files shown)


submission/persistence-common-api/src/main/kotlin/ac/uk/ebi/biostd/persistence/common/service/SubmissionOperations.kt
```diff
@@ -55,14 +47,38 @@ interface SubmissionPersistenceQueryService {
...
+
     fun getRequestStatus(accNo: String, version: Int): RequestStatus
 }
 
+interface SubmissionFilesPersistenceService {
+    fun saveSubmissionFile(file: SubmissionFile)
+
+    fun getSubmissionFile(path: String, accNo: String, version: Int): ExtFile
+
+    fun getSubmissionFiles(accNo: String, version: Int, startingAt: Int = 0): List<Pair<ExtFile, Int>>
```
Juan-EBI, line 75: I think this kind of default parameters are dangerous, I mean IMO we should use default parameters for parameters that are optional, imagine for example operation submit optioanally may accept a on behalf user paraneter but here startAt is not optional is good that every class using the api actually provide a value and using default parameter may introduce a bug because of this.

submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/migrations/DatabaseChangeLog.kt
```diff
@@ -167,3 +172,18 @@ class ChangeLog006 {
         template.updateMulti(Query(), Update().unset(SUB_STATS), DocSubmission::class.java)
     }
 }
+
+@ChangeLog
+class ChangeLog007 {
+    @ChangeSet(order = "007", id = "Create submission files collection", author = "System")
```
Juan-EBI, line 178: *submission request

submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/db/converters/shared/ConverterCommonsFields.kt
```diff
@@ -67,6 +67,13 @@ object FileListDocFileFields {
     const val FILE_LIST_DOC_FILE_FILE_LIST_NAME = "fileListName"
 }
 
+object DocSubmissionFileFields {
```
Juan-EBI, line 70: *DocRequest

submission/persistence-common-api/src/main/kotlin/ac/uk/ebi/biostd/persistence/common/service/SubmissionOperations.kt
```diff
@@ -55,14 +47,38 @@ interface SubmissionPersistenceQueryService {
...
+
+    fun getFileListFiles(accNo: String, version: Int, fileListName: String): List<ExtFile>
```
Juan-EBI, line 77: Not sure if this is temporally but this operation need to be paginated or streamed. Loading all files in memory has been proved to  not be a good option.  We may use spring java doc stream feature https://docs.spring.io/spring-data/mongodb/docs/current/reference/html/#repositories.query-streaming.

Juan-EBI (approved):
feature is looking great! nice improved too!  Please check comments before merge, mainly related to make everything clear this is a request file and not a submission one.

submission/persistence-common-api/src/main/kotlin/ac/uk/ebi/biostd/persistence/common/service/SubmissionOperations.kt
jhoanmanuelms: Yes, the idea is to either make it a stream or a paginated one in the next (and hopefully last PR)

PR #629, opened 2022-10-16
Pivotal ID # 183448558: Allow Fire Bypassing

Juan-EBI:
- added sequence processing support
- refactor clean api to include current submission


jhoanmanuelms (approved):
LGTM

PR #632, opened 2022-10-18
Pivotal ID # 183448558: Allow Fire Bypassing

Juan-EBI:

### DIFF
--- a/submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/fire/FireFilesService.kt
+++ b/submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/fire/FireFilesService.kt
@@ -7,6 +7,7 @@ import ebi.ac.uk.extended.model.ExtFileType.FILE
 import ebi.ac.uk.extended.model.ExtSubmission
 import ebi.ac.uk.extended.model.FireFile
 import ebi.ac.uk.extended.model.NfsFile
+import ebi.ac.uk.extended.model.expectedPath
 import ebi.ac.uk.io.ext.md5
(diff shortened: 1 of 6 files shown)


submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/fire/FireFilesService.kt
```diff
@@ -69,31 +66,36 @@ class FireFilesService(
...
+                val saved = client.save(downloaded, file.md5, file.size)
+                setMetadata(saved.fireOid, file, expectedPath)
+            }
+        }
+    }
 
-        val saved = client.save(fallbackFile(), file.md5, file.size)
-        return setMetadata(saved.fireOid, file, expectedPath)
+    private fun getOrCreate(file: NfsFile, expectedPath: String): FireFile {
+        val matches = client.findByMd5(file.md5)
```
jhoanmanuelms, line 85: Given that we already searched for the files in the sources while verifying the request, it looks like it would be fair to assume that any NfsFile we get at this point is not on FIRE otherwise, we would have a fire file here instead. Maybe it's worth checking this out.

jhoanmanuelms (approved):

submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/fire/FireFilesService.kt
Juan-EBI: it can be in FIRE, if user use NFS source , we do not check if file is fire, we may do this but it will make accepting a request super slow and application will look unresponsive for user

PR #634, opened 2022-10-19
Pivotal ID # 183578315: Submitted draft should not be deleted

Juan-EBI:
https://www.pivotaltracker.com/story/show/183578315
- Introduced new state for draft so they are not deleted on submission

### DIFF
--- a/submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/model/DocSubmissionDraft.kt
+++ b/submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/model/DocSubmissionDraft.kt
@@ -26,6 +26,6 @@ data class DocSubmissionDraft(
     enum class DraftStatus {
         ACTIVE,
         PROCESSING,
-        DELETED
+        ACCEPTED
     }
 }
--- a/submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/model/DocSubmissionDraft.kt
+++ b/submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/model/DocSubmissionDraft.kt
@@ -11,7 +11,7 @@ data class DocSubmissionDraft(
     val userId: String,
     val key: String,
     val content: String,
-    val status: DraftStatus
+    val status: DraftStatus,
 ) {
     constructor(userId: String, key: String, content: String, status: DraftStatus) :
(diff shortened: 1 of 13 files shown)


submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/model/DocSubmissionDraft.kt
```diff
@@ -25,6 +25,7 @@ data class DocSubmissionDraft(
 
     enum class DraftStatus {
         ACTIVE,
-        PROCESSING
+        PROCESSING,
+        DELETED
```
jhoanmanuelms, line 29: Rather than deleted, I think we should call this "PROCESSED" or "ACCEPTED" since I think it reflects better what really happened with the draft: it was submitted and the system accepted it. Also, I think it's counterintuitive that we have something marked as deleted instead of actually deleting it.

submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/service/SubmissionDraftMongoPersistenceService.kt
```diff
@@ -21,8 +22,12 @@ class SubmissionDraftMongoPersistenceService(
         return SubmissionDraft(key, content)
     }
 
-    override fun deleteSubmissionDraft(key: String) {
-        draftDocDataRepository.deleteByKey(key)
+    override fun setDeleteStatus(key: String) {
```
jhoanmanuelms, line 25: In case we go with deleted, I think it should be `setDeletedStatus`

jhoanmanuelms (commented):
Please take a look at my comments

submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/model/DocSubmissionDraft.kt
Juan-EBI: It depends how do you see the business case and in this case a draft,  IMO it is a peace of data, like record that user can created. It can be active , block or deleted.  IMO We do not even process draft of accept them as it  like a memory that can have sometime, can be block for changes or can be deleted. 

IMO the concept of accepted, processing or submitting is related to a submission request not to a DRAFT.

jhoanmanuelms (approved):
LGTM

PR #636, opened 2022-10-31
Pivotal ID # 183522939: On-Behalf Adjustments

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/183522939

- Allow impersonation of inactive users
- Include the notification recipient as part of the submission request
- Create an integration test to validate that inactive user can't login
- Separate the submission stages handling from the submission service functions
- Remove unused code


submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/SubmissionSubmitter.kt
```diff
@@ -22,14 +22,14 @@ class SubmissionSubmitter(
...
-        val (accNo, version) = submissionSubmitter.createRequest(ExtSubmitRequest(submission, rqt.draftKey))
+        val (accNo, version) = submissionSubmitter.createRequest(ExtSubmitRequest(submission, rqt.owner, rqt.draftKey))
         submissionSubmitter.handleRequest(accNo, version)
         return submission
     }
 
     fun createRequest(rqt: SubmitRequest): ExtSubmission {
         val submission = processRequest(rqt)
-        submissionSubmitter.createRequest(ExtSubmitRequest(submission, rqt.draftKey))
+        submissionSubmitter.createRequest(ExtSubmitRequest(submission, rqt.owner, rqt.draftKey))
```
Juan-EBI, line 32: I found really confusing that at request level we are defining owner as:

` val owner: String = onBehalfUser?.email ?: submitter.email`

but when processing submission it is defined as.

` val ownerEmail = onBehalfUser?.email ?: previousVersion?.owner ?: submitter.email`

It looks to me that the first one is not the owner but rather the `notifiedUser`

Juan-EBI (approved):
LGTM please check comments

PR #638, opened 2022-11-09
Pivotal ID # 183557519: Rollback Pagetab MD5 Feature 

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/183557519

Rollback the changes related to adding the MD5 to the pagetab files

### DIFF
--- a/commons/commons-model-extended-mapping/src/main/kotlin/ebi/ac/uk/extended/mapping/to/ToFile.kt
+++ b/commons/commons-model-extended-mapping/src/main/kotlin/ebi/ac/uk/extended/mapping/to/ToFile.kt
@@ -1,6 +1,5 @@
 package ebi.ac.uk.extended.mapping.to
 
-import ebi.ac.uk.extended.model.ExtAttribute
 import ebi.ac.uk.extended.model.ExtFile
 import ebi.ac.uk.extended.model.FireFile
 import ebi.ac.uk.extended.model.NfsFile
@@ -9,10 +8,9 @@ import ebi.ac.uk.model.BioFile
 
 internal const val TO_FILE_EXTENSIONS = "ebi.ac.uk.extended.mapping.to.ToFileKt"
 
-fun ExtFile.toFile(): BioFile =
-    when (this) {
-        is NfsFile -> BioFile(filePath, file.size(), type.value, generateAttributes(attributes))
-        is FireFile -> BioFile(filePath, size, type.value, generateAttributes(attributes))
+fun ExtFile.toFile(): BioFile {
+    return when (this) {
+        is NfsFile -> BioFile(filePath, file.size(), type.value, attributes.map { it.toAttribute() })
+        is FireFile -> BioFile(filePath, size, type.value, attributes.map { it.toAttribute() })
     }
-
-private fun generateAttributes(extAttrs: List<ExtAttribute>) = extAttrs.map { it.toAttribute() }
+}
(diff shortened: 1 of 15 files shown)


commons/commons-model-extended-mapping/src/main/kotlin/ebi/ac/uk/extended/mapping/to/ToFile.kt
```diff
@@ -5,19 +5,14 @@ import ebi.ac.uk.extended.model.ExtFile
...
 import ebi.ac.uk.model.BioFile
-import ebi.ac.uk.model.constants.FileFields.MD5
 
 internal const val TO_FILE_EXTENSIONS = "ebi.ac.uk.extended.mapping.to.ToFileKt"
 
 fun ExtFile.toFile(): BioFile =
     when (this) {
-        is NfsFile -> BioFile(filePath, file.size(), type.value, generateAttributes(md5, attributes))
-        is FireFile -> BioFile(filePath, size, type.value, generateAttributes(md5, attributes))
+        is NfsFile -> BioFile(filePath, file.size(), type.value, generateAttributes(attributes))
```
Juan-EBI, line 14: inlining the  private method looks simpler for me.

```
fun ExtFile.toFile(): BioFile =
    when (this) {
        is NfsFile -> BioFile(filePath, file.size(), type.value, attributes.map { it.toAttribute() })
        is FireFile -> BioFile(filePath, size, type.value, attributes.map { it.toAttribute() })
    }
```

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/ExtSubmissionSubmitter.kt
```diff
@@ -44,7 +44,7 @@ class ExtSubmissionSubmitter(
 
     fun indexRequest(accNo: String, version: Int) = requestIndexer.indexRequest(accNo, version)
 
-    fun loadRequest(accNo: String, version: Int): ExtSubmission = requestLoader.loadRequest(accNo, version)
+    fun loadRequest(accNo: String, version: Int) = requestLoader.loadRequest(accNo, version)
```
Juan-EBI, line 47: notice others methods does not have return type but the pattern I am follow and I think we agree to is only remove return type in private methods.

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/request/SubmissionRequestLoader.kt
```diff
@@ -4,57 +4,40 @@ import ac.uk.ebi.biostd.persistence.common.model.RequestStatus.LOADED
...
-    private val fileProcessingService: FileProcessingService,
-    private val pageTabService: PageTabService,
 ) {
     /**
-     * - Calculate md5 and size for every file in submission request.
-     * - Generate submission pagetab
+     * Calculate md5 and size for every file in submission request.
      */
-    fun loadRequest(accNo: String, version: Int): ExtSubmission {
+    fun loadRequest(accNo: String, version: Int) {
```
Juan-EBI, line 25: I still think consise solution is WAY better that this partitions. clearly the  complexity per line of code is higher as  it does  the same  things but in less code but is more mantianable, which in my opinion is our final goal.  

```
    fun loadRequest(accNo: String, version: Int) {
        val request = requestService.getIndexedRequest(accNo, version)
        val submission = request.submission
        filesRequestService
            .getSubmissionRequestFiles(submission.accNo, submission.version, request.currentIndex)
            .map { it.copy(file = loadFileAttributes(it.file)) }
            .forEach {
                filesRequestService.saveSubmissionRequestFile(it)
                requestService.updateRequestIndex(submission.accNo, submission.version, it.index)
            }
        requestService.saveSubmissionRequest(request.withNewStatus(status = LOADED))
    }
```

Juan-EBI (approved):
LGTM please consider minor changes before merge.

PR #639, opened 2022-11-14
Pivotal ID # 183688205: FIRE Bypassing Publishing Delta

Juan-EBI:
https://www.pivotaltracker.com/story/show/183688205


submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/db/repositories/Repositories.kt
```diff
@@ -18,8 +18,8 @@ import org.springframework.data.mongodb.repository.Query
 import java.util.stream.Stream
 
 interface SubmissionMongoRepository : MongoRepository<DocSubmission, ObjectId> {
-    @Query(value = "{ 'accNo': '?0', 'version': { \$gte: 0 } }", fields = "{ relPath : 1 }")
-    fun getRelPath(accNo: String): SubmissionRelPath
+    @Query(value = "{ 'accNo': '?0', 'version': { \$gte: 0 } }", fields = "{ relPath : 1, released: 2 }")
+    fun getSubData(accNo: String): SubmissionProjection
```
jhoanmanuelms, line 22: 1. Maybe we should call this method `getSubmissionProjection` in order to make it match the response object
2. There's this method called [findLatestBasicByAccNo](https://github.com/EBIBioStudies/biostudies-backend-services/blob/master/submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/service/SubmissionMongoMetaQueryService.kt#L28) which is only being used [here](https://github.com/EBIBioStudies/biostudies-backend-services/blob/master/submission/submission-security/src/main/kotlin/ebi/ac/uk/security/service/UserPrivilegesService.kt#L63) to get the owner. Maybe we could add that field to this method and get rid of that other call which is very inefficient because it loads all the submission just to get the owner

submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/mapping/to/ToExtFileListMapper.kt
```diff
@@ -17,25 +17,27 @@ class ToExtFileListMapper(
      * Maps a DocFileList to corresponding Ext type. Note that empty list is used if includeFileListFiles is false as
      * files list files are not loaded as part of the submission.
      */
+    @Suppress("LongParameterList")
     fun toExtFileList(
         fileList: DocFileList,
         subAccNo: String,
         subVersion: Int,
+        released: Boolean,
```
jhoanmanuelms, line 25: I understand these two fields are required here because we currently don't have that information in the database for old files so I think we should at least add a TODO here to remind us to remove them and add them to the doc files themselves once we do the submission files decoupling.

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/request/SubmissionRequestReleaser.kt
```diff
@@ -1,37 +1,57 @@
...
     fun checkReleased(accNo: String, version: Int): ExtSubmission {
         val request = requestService.getFilesCopiedRequest(accNo, version)
-        val sub = queryService.getExtByAccNoAndVersion(accNo, version, includeFileListFiles = true)
-        if (sub.released) releaseSubmission(sub)
-        requestService.updateRequestStatus(sub.accNo, sub.version, PROCESSED)
+        if (request.submission.released) releaseRequest(request)
+        requestService.updateRequestStatus(request.submission.accNo, request.submission.version, PROCESSED)
         eventsPublisherService.submissionSubmitted(accNo, request.notifyTo)
-        return sub
+        return request.submission
```
jhoanmanuelms, line 39: I don't think it makes sense to return the submission anymore since we're not actually modifying it but just iterating through its files

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/request/SubmissionRequestReleaser.kt
```diff
@@ -1,37 +1,57 @@
...
+        val sub = request.submission
+        filesRequestService
+            .getSubmissionRequestFiles(sub.accNo, sub.version, request.currentIndex)
+            .filterNot { it.file is FireFile && (it.file as FireFile).published }
+            .forEach {
+                fileStorageService.releaseSubmissionFile(it.file, sub.relPath, sub.storageMode)
+                filesRequestService.saveSubmissionRequestFile(it)
+                requestService.updateRequestIndex(sub.accNo, sub.version, it.index)
+            }
+        persistenceService.setAsReleased(sub.accNo)
```
jhoanmanuelms, line 54: This call is not needed in this case since it will only go here if the submission in the request is already released so, in theory, it should never go here if the released field is false

jhoanmanuelms (approved):
LGTM. Please consider my comments 👍

submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/db/repositories/Repositories.kt
Juan-EBI: sorry not following are you suggesting me to change this other operation to be efficient?

Juan-EBI: personally I do not think is convinient to change a security method that is not related at all with this feautre set of changes, I am pro improve the code that you are changing but IMO changing owner is way beyond that.

 Despite this I did try to refactor for merging but this become problematic as basic submission is used other places and also Project is only part of mongo and not comon persistence, result of this I end up having many changes which IMO are not good to be included in this PR.

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/request/SubmissionRequestReleaser.kt
Juan-EBI: it help to know to the caller the last status of the extended error. For example check this code:
```
  fun handleRequest(accNo: String, version: Int): ExtSubmission {
        return when (requestService.getRequestStatus(accNo, version)) {
            REQUESTED -> completeRequest(accNo, version)
            INDEXED -> loadRequestFiles(accNo, version)
            LOADED -> cleanRequestFiles(accNo, version)
            CLEANED -> processRequestFiles(accNo, version)
            FILES_COPIED -> checkReleased(accNo, version)
            else -> throw IllegalStateException("Request accNo=$accNo, version='$version' has been already processed")
        }
    }
```

If we stop retrieving the ext submission alternative will be like quering it aside which wiill just make code less clear. Same with retrigger request endpoint as we want to know the final result if only release is pending we will need to query it asside.

submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/db/repositories/Repositories.kt
jhoanmanuelms: I understand. If you think this is a valuable change, should we create a ticket for it? Otherwise, let's just let it be 😁

PR #642, opened 2022-11-22
Pivotal ID # 183804838: publish files resumable support

Juan-EBI:
https://www.pivotaltracker.com/story/show/183804838
- updated release job to reflect FIRE response

### DIFF
--- a/submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/fire/FireFtpService.kt
+++ b/submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/fire/FireFtpService.kt
@@ -8,8 +8,9 @@ import uk.ac.ebi.fire.client.integration.web.FireClient
 class FireFtpService(
     private val fireClient: FireClient,
 ) : FtpService {
-    override fun releaseSubmissionFile(file: ExtFile, subRelPath: String) {
+    override fun releaseSubmissionFile(file: ExtFile, subRelPath: String): FireFile {
         val fireFile = file as FireFile
-        fireClient.publish(fireFile.fireId)
+        val apiFile = fireClient.publish(fireFile.fireId)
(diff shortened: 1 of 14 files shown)


submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/fire/FireFtpService.kt
```diff
@@ -8,8 +8,9 @@ import uk.ac.ebi.fire.client.integration.web.FireClient
 class FireFtpService(
     private val fireClient: FireClient,
 ) : FtpService {
-    override fun releaseSubmissionFile(file: ExtFile, subRelPath: String) {
+    override fun releaseSubmissionFile(file: ExtFile, subRelPath: String): FireFile {
         val fireFile = file as FireFile
-        fireClient.publish(fireFile.fireId)
+        val apiFile = fireClient.publish(fireFile.fireId)
+        return fireFile.copy(firePath = apiFile.path, published = apiFile.published)
```
jhoanmanuelms, line 14: Just to clarify here, would this imply that the path might have been changed on FIRE (not by the backend but by any other external factor) and we're updating our database to keep it consistent, or is this a matter of keeping the same model of returning the file after processing it and this would be a side effect?

jhoanmanuelms (approved):
LGTM. Just one question

submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/fire/FireFtpService.kt
Juan-EBI: I liltle of bouth, even s unlikle that fire does not publish somehting when we call publish and if we set path x it does not use x but this already pass to us (when using special characters) so using fire return is safe IMO.

having model consisten may help us in the future too as for example we may need to store nfs FTP file path or other information.

PR #643, opened 2022-11-22
Pivotal ID # 183804838: publish files resumable support

Juan-EBI:
https://www.pivotaltracker.com/story/show/183804838
- added new updateRequestFile operation to update file and reques. 
- created new withNewStatus to simplify request update on each stage.

### DIFF
--- a/submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/request/SubmissionRequestIndexer.kt
+++ b/submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/request/SubmissionRequestIndexer.kt
@@ -1,6 +1,5 @@
 package ac.uk.ebi.biostd.submission.submitter.request
 
-import ac.uk.ebi.biostd.persistence.common.model.RequestStatus.INDEXED
 import ac.uk.ebi.biostd.persistence.common.model.SubmissionRequestFile
 import ac.uk.ebi.biostd.persistence.common.service.SubmissionRequestFilesPersistenceService
 import ac.uk.ebi.biostd.persistence.common.service.SubmissionRequestPersistenceService
@@ -22,7 +21,7 @@ class SubmissionRequestIndexer(
 
         logger.info { "${sub.accNo} ${sub.owner} Started indexing submission files" }
         val totalFiles = indexSubmissionFiles(sub)
-        requestService.saveSubmissionRequest(request.withNewStatus(INDEXED, totalFiles))
+        requestService.saveSubmissionRequest(request.indexed(totalFiles))
 
         logger.info { "${sub.accNo} ${sub.owner} Finished indexing submission files" }
     }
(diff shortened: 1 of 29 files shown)


submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/request/SubmissionRequestIndexer.kt
```diff
@@ -22,10 +22,8 @@ class SubmissionRequestIndexer(
         val sub = request.submission
 
         logger.info { "${sub.accNo} ${sub.owner} Started indexing submission files" }
-
         val totalFiles = indexSubmissionFiles(sub)
-        requestService.updateRequestTotalFiles(accNo, version, totalFiles)
-        requestService.updateRequestStatus(accNo, version, INDEXED)
+        requestService.saveSubmissionRequest(request.withNewStatus(INDEXED, totalFiles))
```
Juan-EBI, line 25: rather that perform two individual updates status and files we can simple call save operation,

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/request/SubmissionRequestIndexer.kt
```diff
@@ -22,19 +21,16 @@ class SubmissionRequestIndexer(
...
 
         logger.info { "${sub.accNo} ${sub.owner} Finished indexing submission files" }
     }
 
     private fun indexSubmissionFiles(sub: ExtSubmission): Int {
-        val index = AtomicInteger()
         return extSerializationService
             .fileSequence(sub)
-            .map { SubmissionRequestFile(sub.accNo, sub.version, index.incrementAndGet(), it.filePath, it) }
+            .mapIndexed { idx, file -> SubmissionRequestFile(sub.accNo, sub.version, idx + 1, file.filePath, file) }
```
jhoanmanuelms, line 33: Is this actually better? I thought it was better to have AtomicInteger for this kind of operations instead of making the operation everytime

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/request/SubmissionRequestReleaser.kt
```diff
@@ -45,12 +45,11 @@ class SubmissionRequestReleaser(
         val sub = request.submission
         filesRequestService
             .getSubmissionRequestFiles(sub.accNo, sub.version, request.currentIndex)
-            .filterNot { it.file is FireFile && (it.file as FireFile).published }
-            .map { it.copy(file = fileStorageService.releaseSubmissionFile(it.file, sub.relPath, sub.storageMode)) }
-            .forEach {
-                filesRequestService.saveSubmissionRequestFile(it)
-                requestService.updateRequestIndex(sub.accNo, sub.version, it.index)
+            .map {
```
jhoanmanuelms, line 48: I just noticed we're missing the individual file logs here as in the other steps:

```
logger.info { "${sub.accNo} ${sub.owner} Started publishing file ${it.index}, path='${it.path}'" }

releaseFile

logger.info { "${sub.accNo} ${sub.owner} Finished releasing file ${it.index}, path='${it.path}'" }

```

jhoanmanuelms (commented):
I left a couple of comments

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/request/SubmissionRequestIndexer.kt
Juan-EBI: what I like of current approach is that is kotlin folks the ones that implement index logic and not ourself so I think is simpler to us as we do not mantain that part of the code. Said that IMO diference is minimun.

PR #644, opened 2022-11-25
Pivotal ID # 183573138: Reject Submission With Empty File Lists

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/183573138

- Add the number of files to the file list validations
- Use the file list validator in the submission processing stage
- Add unit tests

### DIFF
--- a/commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/exception/InvalidFileListException.kt
+++ b/commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/exception/InvalidFileListException.kt
@@ -7,5 +7,8 @@ class InvalidFileListException(
     companion object {
         fun directoryCantBeFileList(fileName: String) =
             InvalidFileListException(fileName, "A directory can't be used as File List")
+
+        fun emptyFileList(fileName: String) =
+            InvalidFileListException(fileName, "A file list should contain at least one file")
     }
 }
--- a/submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/FileListValidationTest.kt
+++ b/submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/FileListValidationTest.kt
@@ -52,14 +52,27 @@ class FileListValidationTest(
     }
 
     @Test
-    fun `empty file list`() {
-        val fileList = tempFolder.createFile("EmptyFileList.json")
+    fun `blank file list`() {
+        val fileList = tempFolder.createFile("BlankFileList.json")
 
         webClient.uploadFile(fileList)
 
         val exception = assertThrows(WebClientException::class.java) { webClient.validateFileList(fileList.name) }
         assertThat(exception.statusCode).isEqualTo(HttpStatus.INTERNAL_SERVER_ERROR)
         assertThat(exception).hasMessageContaining("Expected content to be an array")
+
+        webClient.deleteFile(fileList.name)
+    }
+
+    @Test
+    fun `empty file list`() {
+        val fileList = tempFolder.createFile("EmptyFileList.tsv", "Files\tAttr1")
+
+        webClient.uploadFile(fileList)
+
+        val exception = assertThrows(WebClientException::class.java) { webClient.validateFileList(fileList.name) }
+        assertThat(exception.statusCode).isEqualTo(HttpStatus.BAD_REQUEST)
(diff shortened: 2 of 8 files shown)


commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/exception/InvalidFileListException.kt
```diff
@@ -7,5 +7,8 @@ class InvalidFileListException(
     companion object {
         fun directoryCantBeFileList(fileName: String) =
             InvalidFileListException(fileName, "A directory can't be used as File List")
+
+        fun emptyFileList(fileName: String) =
+            InvalidFileListException(fileName, "A file list should contain at least one file")
```
Juan-EBI, line 12: I think we should include the file list failing in case user has many they can identify the one failing.

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/validator/filelist/FileListValidator.kt
```diff
@@ -35,17 +36,23 @@ class FileListValidator(
...
+        fileListFile.inputStream().use {
+            validateFileList(fileListName, it, SubFormat.fromFile(fileListFile), fileSources)
+        }
     }
 
-    private fun validateFileList(stream: InputStream, format: SubFormat, filesSource: FileSourcesList) {
-        serializationService
-            .deserializeFileList(stream, format)
+    private fun validateFileList(name: String, stream: InputStream, format: SubFormat, filesSource: FileSourcesList) {
+        val fileListFiles = serializationService.deserializeFileList(stream, format).take(fileListLimit).toList()
```
Juan-EBI, line 51: we can not do this, this will load all file list files in memory (a sequence is lazy a list it is not)

Juan-EBI, line 51: I think this is incorrectly changing how method use to work, before we were checking for the first `n` null files and reporting them, now we are checking if the first `n` has a empty file. Which means i.e the file 1001 does not exists this method will not fail. Do I am missing something?

commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/exception/InvalidFileListException.kt
jhoanmanuelms: This class already does that, the full error message would be:

```
Problem processing file list 'empty.tsv': A file list should contain at least one file
```

Juan-EBI: nice we amy need to include them file name in message test assertion.

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/FileListValidationTest.kt
```diff
@@ -255,10 +268,27 @@ class FileListValidationTest(
         webClient.deleteFile(fileList.name)
     }
 
+    @Test
+    fun `empty file list on behalf another user`() {
```
Juan-EBI, line 272: personally I think this combination of test is not good. I mean in ideal world when time and budget is not a constrain will be great to have combination test of all our features like multipart requesst on behalf or behalf submission with root path but as we are not there introducing some combination is not good unless you see particual value in this one. IMO https://github.com/EBIBioStudies/biostudies-backend-services/pull/644/files#diff-35b0dc1882619e1c82a0f167ba73be7b1521e0af195c7d095a8cfc70a74d01c5R251 is enough.

commons/commons-serialization/src/test/kotlin/ac/uk/ebi/biostd/json/deserialization/FileListDeserializerTest.kt
```diff
@@ -0,0 +1,57 @@
...
+            val files = testInstance.deserializeList<BioFile>(it).toList()
+            assertThat(files).hasSize(1)
+
+            val file = files.first()
+            assertThat(file.path).isEqualTo("File1.txt")
+        }
+    }
+
+    @Test
+    fun `sequence iterator idempotent operations`() {
```
Juan-EBI, line 35: really nice

Juan-EBI (approved):
LGTM please check comment

PR #647, opened 2022-12-05
Pivotal ID # 183512371: Update Emails With A Collection Specific URL

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/183512371

- Add a property to specify the ST url
- Include the collection specific URL to the notifications
- Simplify the login in the notifications listener class

### DIFF
--- a/submission/submission-handlers/src/main/kotlin/ac/uk/ebi/biostd/handlers/listeners/SubmissionNotificationsListener.kt
+++ b/submission/submission-handlers/src/main/kotlin/ac/uk/ebi/biostd/handlers/listeners/SubmissionNotificationsListener.kt
@@ -73,9 +73,9 @@ class SubmissionNotificationsListener(
     }
 
     private fun getUiUrl(submission: ExtSubmission): String {
-        return buildString {
-            append(notificationProps.uiUrl)
-            submission.collections.firstOrNull()?.let { append("/${it.accNo.lowercase()}") }
+        return when (val col = submission.collections.firstOrNull()) {
+            null -> notificationProps.uiUrl
+            else -> "${notificationProps.uiUrl}/${col.accNo.lowercase()}"
         }
     }
 }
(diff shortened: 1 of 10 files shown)


submission/submission-handlers/src/main/kotlin/ac/uk/ebi/biostd/handlers/listeners/SubmissionNotificationsListener.kt
```diff
@@ -20,63 +20,62 @@ import org.springframework.amqp.rabbit.core.RabbitTemplate
...
 
-    private fun logFailedNotification(message: SubmissionMessage) =
+    private fun onError(message: SubmissionMessage) {
         rabbitTemplate.convertAndSend(BIOSTUDIES_EXCHANGE, NOTIFICATIONS_FAILED_REQUEST_ROUTING_KEY, message)
+        notificationsSender.send(Alert(SYSTEM_NAME, HANDLERS_SUBSYSTEM, String.format(ERROR_MESSAGE, message.accNo)))
+    }
 
-    private fun sendErrorNotification(message: String) =
-        notificationsSender.send(Alert(SYSTEM_NAME, HANDLERS_SUBSYSTEM, message))
+    private fun getUiUrl(submission: ExtSubmission): String {
```
Juan-EBI, line 75: what about
```
    private fun getUiUrl(submission: ExtSubmission): String {
        return when (val col = submission.collections.firstOrNull()) {
            null -> notificationProps.uiUrl
            else -> "${notificationProps.uiUrl}/${col.accNo.lowercase()}"
        }
    }
```

submission/submission-handlers/src/test/kotlin/ac/uk/ebi/biostd/handlers/listeners/SubmissionNotificationsListenerTest.kt
```diff
@@ -79,7 +104,7 @@ class SubmissionNotificationsListenerTest(
 
         verify(exactly = 0) {
             webConsumer.getExtSubmission("ext-tab-url")
-            rtNotificationService.notifySuccessfulSubmission(submission, "Dr Owner", "ui-url")
+            rtNotificationService.notifySuccessfulSubmission(submission, "Dr Owner", "ui-url", "st-url")
```
Juan-EBI, line 107: as mention before I think verify never should always has specific value. problem is that imagine this is call with "st-url-2" assertion will not fail so test will pass.  I think is simpler and better for test to use any if zero interactiosn scenarios. Also I think is simple to reader to acknowledge this method should not be called rather than this method should not be called with X and Y and Z parameters.

Juan-EBI (approved):
LGTM please check minor comments

PR #648, opened 2022-12-06
Pivotal ID # 183126684: Log Submission Draft ID

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/183126684

- Include draft related information in the logs
- Change the logging level for the retry web template from error to warning

### DIFF
--- a/submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/SubmissionSubmitter.kt
+++ b/submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/SubmissionSubmitter.kt
@@ -65,16 +65,26 @@ class SubmissionSubmitter(
         try {
             logger.info { "${rqt.accNo} ${rqt.owner} Started processing submission request" }
 
-            rqt.draftKey?.let { draftService.setProcessingStatus(rqt.owner, it) }
+            rqt.draftKey?.let {
+                draftService.setProcessingStatus(rqt.owner, it)
+                logger.info { "${rqt.accNo} ${rqt.owner} Status of draft with key '$it' set to PROCESSING" }
+            }
             val processed = submissionProcessor.processSubmission(rqt)
             parentInfoService.executeCollectionValidators(processed)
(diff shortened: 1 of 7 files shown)


submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/SubmissionSubmitter.kt
```diff
@@ -65,16 +65,26 @@ class SubmissionSubmitter(
         try {
             logger.info { "${rqt.accNo} ${rqt.owner} Started processing submission request" }
 
-            rqt.draftKey?.let { draftService.setProcessingStatus(rqt.owner, it) }
+            rqt.draftKey?.let {
```
Juan-EBI, line 68: I think this is adding too much complexity to this method, we times the same conditional (if draftkey != null)  in this few lines.  Not really sure about but we may consider include accNo and owner in `draftService` operations to log on each status change.  Not much better as we will be including parameters only for logging purposes but as it is righ it looks super verbose.

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/SubmissionSubmitter.kt
```diff
@@ -65,16 +65,26 @@ class SubmissionSubmitter(
...
+                draftService.setProcessingStatus(rqt.owner, it)
+                logger.info { "${rqt.accNo} ${rqt.owner} Status of draft with key '$it' set to PROCESSING" }
+            }
             val processed = submissionProcessor.processSubmission(rqt)
             parentInfoService.executeCollectionValidators(processed)
-            rqt.draftKey?.let { draftService.setAcceptedStatus(it) }
+            rqt.draftKey?.let {
+                logger.info { "${rqt.accNo} ${rqt.owner} Assigned accNo '${processed.accNo}' to draft with key '$it'" }
```
Juan-EBI, line 75: this one I can guess can be moved to `submissionProcessor.processSubmission` where actuallty this happens (I am looking for way to reduce verbosity here )

PR #649, opened 2022-12-06
 Pivotal ID # 183971429: Enumeration itests and tests inventory

msrada94:
https://www.pivotaltracker.com/story/show/183971429

enumerate itests for tracking purposes


submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/collection/query/CollectionsListTest.kt
```diff
@@ -49,7 +49,7 @@ class CollectionsListTest(
     }
 
     @Test
-    fun `list collections for regular user`() {
+    fun `20,2 - list collections for regular user`() {
```
Juan-EBI, line 52: lets use 20-2 as this is more conventional for enumations

Juan-EBI (commented):
please check comment and build status as it is failing

Juan-EBI (approved):
LGTM

PR #652, opened 2022-12-06
Pivotal ID # 183545405: Review tab-delimited parsing, esp. line-breaks

Juan-EBI:
https://www.pivotaltracker.com/story/show/183545405


commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/tsv/serialization/TsvBuilder.kt
```diff
@@ -1,18 +1,26 @@
 package ac.uk.ebi.biostd.tsv.serialization
 
+import ebi.ac.uk.base.EMPTY
+
 internal const val TITLE_KEY = "Title"
 internal const val ACC_NO_KEY = "Submission"
 internal const val ROOT_PATH_KEY = "RootPath"
+internal val specialCharRegex = "[\n\t\"]".toRegex()
```
jhoanmanuelms, line 8: I think we could rename this to MULTILNE_BLOCK or something clearer because the current name is very generic

commons/commons-util/src/main/kotlin/ebi/ac/uk/util/file/PoiExt.kt
```diff
@@ -5,22 +5,29 @@ import org.apache.poi.ss.usermodel.Cell
...
+            }
 
-    return elements.trim()
+            elements++
+            yield(row.asString())
+        }
+    }
 }
 
+private val specialCharRegex = "[\n\t\"]".toRegex()
```
jhoanmanuelms, line 29: Same comment about the name

commons/commons-util/src/main/kotlin/ebi/ac/uk/util/file/PoiExt.kt
```diff
@@ -5,22 +5,29 @@ import org.apache.poi.ss.usermodel.Cell
...
- * rows they are completed with emtpy values.
+ * rows they are completed with empty values.
  */
-fun Sheet.asTsvList(): List<String> {
-    val elements = LinkedList<String>()
-    for (row in this) {
-        while (elements.size < row.rowNum) elements.add(EMPTY)
-        elements.add(row.asString())
-    }
+fun Sheet.asTsvSequence(): Sequence<String> {
```
jhoanmanuelms, line 13: Sequences FTW 🤘

commons/commons-util/src/test/kotlin/ebi/ac/uk/util/file/ExcelReaderTest.kt
```diff
@@ -77,7 +77,29 @@ class ExcelReaderTest(private val temporaryFolder: TemporaryFolder) {
...
+                }
+            }
+        }
+
+        val expectedTsv = tsv {
+            line("Submission")
+            line("Title", "\"Excel Submission \n with a line break\"")
+        }
+
+        assertThat(asTsv(testFile)).hasContent(expectedTsv.toString().plus("\n"))
```
jhoanmanuelms, line 102: Instead of adding `plus("\n")`, I think it would be clearer to add the new line in the expected tsv

commons/commons-serialization/src/test/kotlin/ac/uk/ebi/biostd/test/SubTestFactory.kt
```diff
@@ -22,7 +22,7 @@ fun createVenousBloodMonocyte() = submission("S-IHECRE00000919.1") {
 
     section("Study") {
         accNo = "SECT-001"
-        attribute(name = "Title", value = "venous blood, Monocyte")
+        attribute(name = "Title", value = "venous blood, \n Monocyte")
```
jhoanmanuelms, line 25: I think we should also include this scenario in the [integration test](https://github.com/EBIBioStudies/biostudies-backend-services/blob/master/submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/MultipartFileSubmissionApiTest.kt#L68) we have for XLSX

jhoanmanuelms (approved):
Nice catch man, LGTM. Please, consider my comments before merging

commons/commons-util/src/test/kotlin/ebi/ac/uk/util/file/ExcelReaderTest.kt
Juan-EBI: not sure what I was thiking when did this

commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/tsv/serialization/TsvBuilder.kt
Juan-EBI: the issue here is that not only multiline block but also a block for example with a inner tab

PR #653, opened 2022-12-07
Pivotal ID # 183817421: Migrate DEV To HL

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/183817421

- Specify terminal stdout and stderr output streams to avoid the terminal being clogged
- Remove the "sh" command from the command execution since it's not required in standard linux shell


Juan-EBI (approved):
LGTM

PR #655, opened 2022-12-14
Pivotal ID # 184014407: Multiple Reference Files In Resubmission

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/184014407

- Set the proper attributes to a file that's referenced multiple times
- Change the referenced files method in order to support multiple file references


Juan-EBI (approved):
LGTM

PR #656, opened 2022-12-14
Pivotal ID # 183944271: First Collection Submission AccNo

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/183944271

Start the new accNo sequences with 1 instead of 0


Juan-EBI (approved):
LGTM

PR #658, opened 2022-12-15
Pivotal ID # 183230982: Avoid Dots In File Paths

jhoanmanuelms:

- Move all the path related logic to a single place
- Include the logic to validate relative paths
- Use the single path validation method in all the deserializers
- Reduce error messages verbosity


jhoanmanuelms:
Closed in favor of #727

PR #660, opened 2022-12-20
Pivotal ID # 184014414: Permissions For Resubmission To Collection

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/184014414

Give priority to owner's resubmission

### DIFF
--- a/submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/security/SubmitPermissionTest.kt
+++ b/submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/security/SubmitPermissionTest.kt
@@ -46,6 +46,7 @@ class SubmitPermissionTest(
     fun init() {
         securityTestService.ensureUserRegistration(SuperUser)
         securityTestService.ensureUserRegistration(ExistingUser)
+        securityTestService.ensureUserRegistration(ImpersonatedUser)
 
         superUserWebClient = getWebClient(serverPort, SuperUser)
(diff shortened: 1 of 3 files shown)


submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/security/SubmitPermissionTest.kt
```diff
@@ -184,6 +185,36 @@ class SubmitPermissionTest(
         assertThat(regularUserWebClient.submitSingle(resubmission, SubmissionFormat.TSV)).isSuccessful()
     }
 
+    @Test
+    fun `4-8 owner resubmits without attach permission`() {
```
Juan-EBI, line 189: This test is not clear for me, if what we want to test is that "owner resubmits without attach permission" why we are using belhalf client and a imporsonated user in the test.  test structure should be something like

1.  submit with user A in project B
2. Remove Attach permission from user A
3. Re submit and validate success.

Do am I missing something?

Juan-EBI (approved):
Please check comment

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/security/SubmitPermissionTest.kt
jhoanmanuelms: No, you're not, the test case you're describing is correct. The reason why I added the onbehalf step was to recreate the scenario most used by Ahmed for AE submissions

PR #661, opened 2022-12-20
Pivotal ID # 183825532: Fix size for folders on FIRE

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/183825532

- Fix folder size calculation
- Include a folder in the AllInOneSubmission test case

### DIFF
--- a/commons/commons-util/src/main/kotlin/ebi/ac/uk/io/FileUtils.kt
+++ b/commons/commons-util/src/main/kotlin/ebi/ac/uk/io/FileUtils.kt
@@ -7,6 +7,7 @@ import ebi.ac.uk.io.FileUtilsHelper.createFolderIfNotExist
 import ebi.ac.uk.io.FileUtilsHelper.createParentDirectories
 import ebi.ac.uk.io.FileUtilsHelper.createSymLink
 import ebi.ac.uk.io.ext.notExist
+import ebi.ac.uk.io.ext.size
 import org.apache.commons.codec.digest.DigestUtils
 import java.io.File
 import java.io.InputStream
@@ -112,7 +113,7 @@ object FileUtils {
 
     fun isDirectory(file: File): Boolean = Files.isDirectory(file.toPath())
 
-    fun size(file: File): Long = Files.size(file.toPath())
+    fun size(file: File): Long = if (file.isFile) Files.size(file.toPath()) else calculateDirectorySize(file)
 
     fun md5(file: File): String = if (file.isFile) calculateMd5(file) else ""
 
@@ -131,6 +132,8 @@ object FileUtils {
         if (isDirectory(file)) Files.list(file.toPath()).map { it.toFile() }.toList() else emptyList()
 
     private fun calculateMd5(file: File): String = file.inputStream().use { DigestUtils.md5Hex(it).uppercase() }
+
+    private fun calculateDirectorySize(dir: File) = dir.walkTopDown().filter { it.isFile }.map { it.size() }.sum()
 }
 
 @Suppress("TooManyFunctions")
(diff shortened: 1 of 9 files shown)


commons/commons-util/src/main/kotlin/ebi/ac/uk/io/FileUtils.kt
```diff
@@ -112,7 +113,7 @@ object FileUtils {
 
     fun isDirectory(file: File): Boolean = Files.isDirectory(file.toPath())
 
-    fun size(file: File): Long = Files.size(file.toPath())
+    fun size(file: File): Long = if (file.isFile) Files.size(file.toPath()) else calculateDirectorySize(file)
```
Juan-EBI, line 116: Files.size is not recursive  !! great catch

commons/commons-util/src/test/kotlin/ebi/ac/uk/io/FileUtilsTest.kt
```diff
@@ -246,11 +246,24 @@ internal class FileUtilsTest(private val temporaryFolder: TemporaryFolder) {
...
+        fun `directory size`() {
+            val directory = temporaryFolder.createDirectory("dir-size-test")
+
+            temporaryFolder.createDirectory("dir-size-test/level-1")
+            temporaryFolder.createDirectory("dir-size-test/level-1/level-2")
+            temporaryFolder.createFile("dir-size-test/file1.txt", "test 1")
+            temporaryFolder.createFile("dir-size-test/level-1/file2.txt", "test 2")
+            temporaryFolder.createFile("dir-size-test/level-1/level-2/file3.txt", "test 3")
+
+            assertThat(FileUtils.size(directory)).isEqualTo(18L)
```
Juan-EBI, line 264: I am wondering if we shoud do something like 
```
        @Test
        fun `directory size`() {
            val directory = temporaryFolder.createDirectory("dir-size-test")

            temporaryFolder.createDirectory("dir-size-test/level-1")
            temporaryFolder.createDirectory("dir-size-test/level-1/level-2")
            val f1 = temporaryFolder.createFile("dir-size-test/file1.txt", "test 1")
            val f2 = temporaryFolder.createFile("dir-size-test/level-1/file2.txt", "test 2")
            val f3 = temporaryFolder.createFile("dir-size-test/level-1/level-2/file3.txt", "test 3")

            assertThat(FileUtils.size(directory)).isEqualTo(f1.size() + f2.size() + f3.size())
        }
```

I am aware we would be  asserting function `FileUtils.size` using the function `size`  but as it is right now it looks super hard to be to guess expected size is 18.

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/factory/AllInOne.kt
```diff
@@ -21,5 +21,4 @@ fun submissionsFiles(tempFolder: File): List<SubmissionFile> = listOf(
     SubmissionFile(tempFolder.createOrReplaceFile("DataFile5.txt", "content 5")),
     SubmissionFile(tempFolder.createOrReplaceFile("DataFile6.txt", "content 6"), "Folder1"),
     SubmissionFile(tempFolder.createOrReplaceFile("DataFile7.txt", "content 7")),
-    SubmissionFile(tempFolder.createOrReplaceFile("DataFile8.txt", "content 8"), "Folder1"),
```
Juan-EBI, line 24: what is the reason of this change?

jhoanmanuelms: The only reason for this file was to have a second file to reference in the file list test. Now that I included the folder itself as the second reference, I thought it no longer makes sense to have it

Juan-EBI (approved):
LGTM

PR #663, opened 2023-01-04
Pivotal ID # 184149873: Releaser should not fail on a single submission fail

Juan-EBI:
https://www.pivotaltracker.com/story/show/184149873

### DIFF
--- a/scheduler/tasks/submission-releaser-task/src/main/kotlin/uk/ac/ebi/scheduler/releaser/service/SubmissionReleaserService.kt
+++ b/scheduler/tasks/submission-releaser-task/src/main/kotlin/uk/ac/ebi/scheduler/releaser/service/SubmissionReleaserService.kt
@@ -42,7 +42,6 @@ class SubmissionReleaserService(
     }
 
     private fun releaseSafely(releaseData: ReleaseData) {
-        logger.info { "Releasing submission ${releaseData.accNo}" }
         runCatching { releaseSubmission(releaseData) }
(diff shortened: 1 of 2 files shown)


scheduler/tasks/submission-releaser-task/src/main/kotlin/uk/ac/ebi/scheduler/releaser/service/SubmissionReleaserService.kt
```diff
@@ -41,8 +41,13 @@ class SubmissionReleaserService(
             .forEach(::generateFtpLink)
     }
 
+    private fun releaseSafely(releaseData: ReleaseData) {
+        runCatching { releaseSubmission(releaseData) }
+            .onFailure { logger.info { "Failed to release submission ${releaseData.accNo}" } }
```
jhoanmanuelms, line 46: This is a great improvement, we definitively shouldn't block all the other submissions.

I wonder if we should add this to another queue or maybe send a slack notification. Otherwise, we won't actually realize there are failing submissions for release until the users complain. WDYT?

Juan-EBI: To be address in https://www.pivotaltracker.com/story/show/183688670

jhoanmanuelms (approved):
LGTM

PR #669, opened 2023-01-12
Pivotal ID # 184173951: Temp pagetab fields should not generated on root temp path

Juan-EBI:
https://www.pivotaltracker.com/story/show/184173951

### DIFF
--- a/submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/pagetab/PageTabService.kt
+++ b/submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/pagetab/PageTabService.kt
@@ -9,12 +9,15 @@ import ebi.ac.uk.io.ext.size
 import uk.ac.ebi.extended.serialization.service.TrackSection
 import uk.ac.ebi.extended.serialization.service.iterateSections
 import java.io.File
+import java.time.LocalDate
+import java.time.temporal.ChronoField
 
 class PageTabService(
-    private val tempFolder: File,
+    private val baseTempFolder: File,
     private val pageTabUtil: PageTabUtil,
 ) : PageTabService {
     override fun generatePageTab(sub: ExtSubmission): ExtSubmission {
+        val tempFolder = createTempFolder(sub.accNo, sub.version)
         val subFiles = pageTabUtil.generateSubPageTab(sub, tempFolder)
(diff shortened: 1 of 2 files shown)


submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/pagetab/PageTabService.kt
```diff
@@ -9,12 +9,15 @@ import ebi.ac.uk.io.ext.size
...
+import java.time.LocalDate
+import java.time.temporal.ChronoField
 
 class PageTabService(
-    private val tempFolder: File,
+    private val baseTempFolder: File,
     private val pageTabUtil: PageTabUtil,
 ) : PageTabService {
     override fun generatePageTab(sub: ExtSubmission): ExtSubmission {
+        val tempFolder = createTempFolder(sub.accNo, sub.version)
```
jhoanmanuelms, line 20: Nice catch

submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/pagetab/PageTabService.kt
```diff
@@ -63,4 +66,17 @@ class PageTabService(
             size = file.size(),
             attributes = emptyList()
         )
+
+
+    private fun createTempFolder(accNo: String, version: Int): File {
```
jhoanmanuelms, line 71: I noticed we already have the same logic [here](https://github.com/EBIBioStudies/biostudies-backend-services/blob/master/commons/commons-serialization-util/src/main/kotlin/uk/ac/ebi/serialization/common/FilesResolver.kt#L23). I think it might be worth it to have this in a common place, maybe something in FileUtils like `generateTempSubmissionFolder` or something like that. WDYT?

jhoanmanuelms (approved):

submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/pagetab/PageTabService.kt
Juan-EBI: The problem for me is that FileUtils is a class the contains IO file operations, copyOrReplaceFile, deleteFile, createHardLink, moveFile. All this operations does not have a relation with submission or even biostudies.

PR #677, opened 2023-01-30
Pivotal ID # 184337303: Add base path to submission relative path

msrada94:
https://www.pivotaltracker.com/story/show/184337303

Please include a list of the changes:
* Added application property base submission relative path for e2e testing purposes
* Added unit and integrations tests

### DIFF
--- a/submission/submission-config/src/main/kotlin/ac/uk/ebi/biostd/common/properties/ApplicationProperties.kt
+++ b/submission/submission-config/src/main/kotlin/ac/uk/ebi/biostd/common/properties/ApplicationProperties.kt
@@ -12,6 +12,7 @@ data class ApplicationProperties(
     val requestFilesPath: String,
     val ftpPath: String,
     val instanceBaseUrl: String,
+    val baseSubmissionRelPath: String = "",
     val consumers: Int,
     val maxConsumers: Int,
     val security: SecurityProperties,
--- a/submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/itest/ITestListener.kt
+++ b/submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/itest/ITestListener.kt
@@ -67,7 +67,7 @@ class ITestListener : TestExecutionListener {
 
     private fun appPropertiesSetup() {
         System.setProperty("app.submissionPath", nfsSubmissionPath.absolutePath)
-        System.setProperty("app.baseSubmissionRelPath", "helloworld")
+        System.setProperty("app.baseSubmissionRelPath", "")
         System.setProperty("app.ftpPath", nfsFtpPath.absolutePath)
         System.setProperty("app.fireTempDirPath", fireTempFolder.absolutePath)
         System.setProperty("app.tempDirPath", tempDirPath.absolutePath)
--- a/submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/itest/ITestListener.kt
+++ b/submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/itest/ITestListener.kt
@@ -67,6 +67,7 @@ class ITestListener : TestExecutionListener {
 
     private fun appPropertiesSetup() {
         System.setProperty("app.submissionPath", nfsSubmissionPath.absolutePath)
+        System.setProperty("app.baseSubmissionRelPath", "helloworld")
         System.setProperty("app.ftpPath", nfsFtpPath.absolutePath)
         System.setProperty("app.fireTempDirPath", fireTempFolder.absolutePath)
         System.setProperty("app.tempDirPath", tempDirPath.absolutePath)
--- a/submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/SubmissionApiTest.kt
+++ b/submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/SubmissionApiTest.kt
@@ -192,7 +192,6 @@ class SubmissionApiTest(
         assertThat(File("$ftpPath/${submitted.relPath}/Files/folder1")).isEmptyDirectory()
     }
 
-
     @Nested
     inner class SubmitBaseSubmissionRelPath {
         @Nested
--- a/submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/SubmissionApiTest.kt
+++ b/submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/SubmissionApiTest.kt
@@ -28,11 +28,13 @@ import ebi.ac.uk.model.extensions.title
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.jupiter.api.Assertions.assertThrows
 import org.junit.jupiter.api.BeforeAll
+import org.junit.jupiter.api.Nested
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.condition.EnabledIfSystemProperty
 import org.junit.jupiter.api.extension.ExtendWith
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.boot.test.context.SpringBootTest
+import org.springframework.boot.test.context.SpringBootTest.WebEnvironment.RANDOM_PORT
 import org.springframework.boot.test.web.server.LocalServerPort
 import org.springframework.context.annotation.Import
 import org.springframework.http.HttpStatus
@@ -42,13 +44,13 @@ import kotlin.test.assertFailsWith
 
 @Import(FilePersistenceConfig::class)
 @ExtendWith(SpringExtension::class)
-@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+@SpringBootTest(webEnvironment = RANDOM_PORT)
 class SubmissionApiTest(
     @Autowired val securityTestService: SecurityTestService,
     @Autowired val submissionRepository: SubmissionPersistenceQueryService,
     @Autowired val sequenceRepository: SequenceDataRepository,
     @Autowired val toSubmissionMapper: ToSubmissionMapper,
-    @LocalServerPort val serverPort: Int
+    @LocalServerPort val serverPort: Int,
 ) {
     private lateinit var webClient: BioWebClient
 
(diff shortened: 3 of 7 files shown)


submission/submission-config/src/main/kotlin/ac/uk/ebi/biostd/common/properties/ApplicationProperties.kt
```diff
@@ -12,6 +12,7 @@ data class ApplicationProperties(
     val requestFilesPath: String,
     val ftpPath: String,
     val instanceBaseUrl: String,
+    val baseSubmissionRelPath: String = "",
```
Juan-EBI, line 15: just `subBasePath` sound better for me and we can add comment documenting , instead of making it an empty string lets make a nullable field.

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/itest/ITestListener.kt
```diff
@@ -67,6 +67,7 @@ class ITestListener : TestExecutionListener {
 
     private fun appPropertiesSetup() {
         System.setProperty("app.submissionPath", nfsSubmissionPath.absolutePath)
+        System.setProperty("app.baseSubmissionRelPath", "")
```
Juan-EBI, line 70: as it becomes nullable we do not used it here

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/SubmissionApiTest.kt
```diff
@@ -190,6 +192,73 @@ class SubmissionApiTest(
...
+        inner class RelPathWithSlash(@LocalServerPort val serverPort: Int) {
+            private lateinit var webClient: BioWebClient
+
+            @BeforeAll
+            fun init() {
+                webClient = getWebClient(serverPort, SuperUser)
+            }
+
+            @Test
+            fun `16-10 submission with file when baseSubmissionRelPath has slash`() {
```
Juan-EBI, line 208: lets ignore this details in itest and just assume user set path correctly

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/service/AccNoService.kt
```diff
@@ -50,7 +51,8 @@ class AccNoService(
     internal fun getRelPath(accNo: AccNumber): String {
         val prefix = accNo.prefix
         val suffix = accNo.numericValue.orEmpty().padStart(3, '0')
-        return "$prefix/${suffix.takeLast(PATH_DIGITS)}/$accNo".removePrefix("/")
+        val base = if (baseRelPath.endsWith("/")) baseRelPath else "$baseRelPath/"
```
Juan-EBI, line 54: lets avoid this and always remove any trailing slash, you can use `trimEnd`

submission/submitter/src/test/kotlin/ac/uk/ebi/biostd/submission/service/AccNoServiceTest.kt
```diff
@@ -32,22 +32,57 @@ class AccNoServiceTest(
...
+    inner class GetRelPath {
+        private val testInstanceEmptyBase = AccNoService(service, accNoPatternUtil, privilegesService, baseRelPath = "")
+        private val testInstanceBaseWithSlash =
+            AccNoService(service, accNoPatternUtil, privilegesService, "baseRelPath/")
+        private val testInstanceBaseWithNoSlash =
+            AccNoService(service, accNoPatternUtil, privilegesService, "baseRelPath")
+
+        @ParameterizedTest(name = "prefix is {0} and numeric value is {1}")
+        @CsvSource(
+            "S-DIXA-AN-002, S-DIXA-AN-/002/S-DIXA-AN-002",
```
Juan-EBI, line 50: lets simple include the `basePath` here in the parameters instead of creating two test classes. In that way we use single class.

Juan-EBI (changes requested):
please check commments

submission/submission-config/src/main/kotlin/ac/uk/ebi/biostd/common/properties/ApplicationProperties.kt
```diff
@@ -3,6 +3,9 @@ package ac.uk.ebi.biostd.common.properties
 import org.springframework.boot.context.properties.ConfigurationProperties
 import org.springframework.boot.context.properties.ConstructorBinding
 
+/**
+ * @param subBasePath prefix of submission relPath. Should be null unless backend is used for e2e testing.
```
Juan-EBI, line 7: comment can be way better IMO.  `Should be null unless backend is used for e2e testing.` this does not explain what the parameter does but rather tell the reader what not to do. We do not want to tell the developers what to do, we want to explain so they are able to decide it well informed.  Also this comment is for humans so you may favor use relative path instead of relPath for example/ 

`@param subBasePath extra path added as a prefix of submission relative path, Allows to allocate submissions under diferent paths`

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/SubmissionApiTest.kt
```diff
@@ -190,6 +192,38 @@ class SubmissionApiTest(
         assertThat(File("$ftpPath/${submitted.relPath}/Files/folder1")).isEmptyDirectory()
     }
 
+    @Nested
+    @SpringBootTest(webEnvironment = RANDOM_PORT, properties = ["app.subBasePath=subRelPath"])
+    inner class SubmitWhenSubmissionRelPath(@LocalServerPort val serverPort: Int) {
```
Juan-EBI, line 197: `SubmitWebBasePath` this class is not related to relpath

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/SubmissionApiTest.kt
```diff
@@ -190,6 +192,38 @@ class SubmissionApiTest(
         assertThat(File("$ftpPath/${submitted.relPath}/Files/folder1")).isEmptyDirectory()
     }
 
+    @Nested
+    @SpringBootTest(webEnvironment = RANDOM_PORT, properties = ["app.subBasePath=subRelPath"])
```
Juan-EBI, line 196: this name make everything more confusing naming the base path `subRelPath` just confuse reader.  `base/path` or `submissions-base-path`

submission/submitter/src/test/kotlin/ac/uk/ebi/biostd/submission/service/AccNoServiceTest.kt
```diff
@@ -29,24 +29,38 @@ private const val PROJECT_PATTERN = "!{ABC-}"
...
+        "S-DIXA-011, basePath/S-DIXA-/011/S-DIXA-011, basePath",
+        "1-AAA, basePath/1-AAA/000/1-AAA, basePath",
+        "S-SCDT-EMBOJ-2019-103549, basePath/S-SCDT-EMBOJ-2019-/549/S-SCDT-EMBOJ-2019-103549, basePath",
+        "S-SCDT-EMBOR-2017-44445V1, basePath/S-SCDT-EMBOR-2017-44445V/001/S-SCDT-EMBOR-2017-44445V1, basePath",
+        "S-DIXA-AN-002, basePath/S-DIXA-AN-/002/S-DIXA-AN-002, basePath/",
+        "S-BSST11, basePath/S-BSST/011/S-BSST11, basePath/",
+        "S-DIXA-011, basePath/S-DIXA-/011/S-DIXA-011, basePath/",
+        "1-AAA, basePath/1-AAA/000/1-AAA, basePath/",
+        "S-SCDT-EMBOJ-2019-103549, basePath/S-SCDT-EMBOJ-2019-/549/S-SCDT-EMBOJ-2019-103549, basePath/",
+        "S-SCDT-EMBOR-2017-44445V1, basePath/S-SCDT-EMBOR-2017-44445V/001/S-SCDT-EMBOR-2017-44445V1, basePath/",
```
Juan-EBI, line 59: this number of permutations is too hard to mantain, what each case is is not clear.  Lets add only two more cases to original one.

Juan-EBI (changes requested):
plase check comments

submission/submission-config/src/main/kotlin/ac/uk/ebi/biostd/common/properties/ApplicationProperties.kt
msrada94: thanks for the comment! yes is quite clear

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/SubmissionApiTest.kt
msrada94: agree

submission/submission-webapp/src/main/resources/application.yml
```diff
@@ -65,6 +65,7 @@ app:
   requestFilesPath: # Base directory absolute path where request file list are stored.
   fireTempDirPath: # Absolute path of the temp folder for FIRE storage
   instanceBaseUrl: # Instance base url e.g. http://server.ebi.ac.uk:5541
+  subBasePath: # Extra path added as a prefix of submission relative path. Allows to allocate submissions under different paths
```
jhoanmanuelms, line 68: I'd suggest:
```If present, it will be used as a prefix to the submission relative path```

Not sure about the purpose part. Usually, you want to document what something does, not what's the purpose since, in reality, we're not actually checking that

File: submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/common/config/SubmitterConfig.kt [hunk 11]
```diff
@@ -202,16 +204,18 @@ class SubmitterConfig {
 
     @Configuration
     @Suppress("MagicNumber")
+    @EnableConfigurationProperties
     class ServiceConfig(
         private val service: PersistenceService,
         private val queryService: SubmissionMetaQueryService,
         private val userPrivilegesService: IUserPrivilegesService,
+        @Value("\${app.subBasePath:#{null}}") private val subBasePath: String?,
```
jhoanmanuelms, line 212: Why are we using a value notation instead of adding it to the [corresponding object](https://github.com/EBIBioStudies/biostudies-backend-services/blob/master/submission/submission-config/src/main/kotlin/ac/uk/ebi/biostd/common/properties/ApplicationProperties.kt#L8)?

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/service/AccNoService.kt
```diff
@@ -50,7 +51,9 @@ class AccNoService(
     internal fun getRelPath(accNo: AccNumber): String {
         val prefix = accNo.prefix
         val suffix = accNo.numericValue.orEmpty().padStart(3, '0')
-        return "$prefix/${suffix.takeLast(PATH_DIGITS)}/$accNo".removePrefix("/")
+        val basePath = subBasePath?.trim('/')
+        val basicRelPath = "$basePath/$prefix/${suffix.takeLast(PATH_DIGITS)}/$accNo".removePrefix("/")
+        return if (basePath != null) "$basePath/$basicRelPath" else basicRelPath
```
jhoanmanuelms, line 56: `basePath`is being included in `basicRelPath`. I'd suggest:

```
val relPath = "$prefix/${suffix.takeLast(PATH_DIGITS)}/$accNo".removePrefix("/")
return if (basePath != null) "${basePath.trim('/')}/$relPath" else relPath
```

submission/submitter/src/test/kotlin/ac/uk/ebi/biostd/submission/service/AccNoServiceTest.kt
```diff
@@ -29,25 +29,31 @@ private const val PROJECT_PATTERN = "!{ABC-}"
...
-        "S-BSST11, S-BSST/011/S-BSST11",
-        "S-DIXA-011, S-DIXA-/011/S-DIXA-011",
-        "1-AAA, 1-AAA/000/1-AAA",
-        "S-SCDT-EMBOJ-2019-103549, S-SCDT-EMBOJ-2019-/549/S-SCDT-EMBOJ-2019-103549",
-        "S-SCDT-EMBOR-2017-44445V1, S-SCDT-EMBOR-2017-44445V/001/S-SCDT-EMBOR-2017-44445V1"
-    )
-    fun getRelPath(value: String, expected: String) {
-        assertThat(testInstance.getRelPath(accNoPatternUtil.toAccNumber(value))).isEqualTo(expected)
+    inner class GetRelPath {
+        @ParameterizedTest(name = "prefix is {0}, numeric value is {1} and submission base path is {2}")
```
jhoanmanuelms, line 41: I noticed this isn't quite accurate since {0} is actually the "numeric value" which is actually the accession 😅. Also, {1} is the expected relPath. I think we should change it to:

"{0} = accession, {1} = base path, {2}=expected rel path"

Which is shorter, clearer and follows the parameters -> expected result pattern. WDYT?

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/SubmissionApiTest.kt
```diff
@@ -247,6 +249,38 @@ class SubmissionApiTest(
...
+    inner class SubmitWebBasePath(@LocalServerPort val serverPort: Int) {
+        private lateinit var webClient: BioWebClient
+
+        @BeforeAll
+        fun init() {
+            webClient = getWebClient(serverPort, SuperUser)
+        }
+
+        @Test
+        fun `16-10 submission when system has submission base path`() {
```
jhoanmanuelms, line 263: I'd suggest: "submission when the system has the basePath property configured"

However, I'd expect this property to be used later in subsequent tests which I think would make this particular test case obsolete

jhoanmanuelms (changes requested):
The changes are good in general. I left a couple of suggestions for you to consider.

jhoanmanuelms (approved):
LGTM

PR #679, opened 2023-02-03
Pivotal ID # 183804703: BioImages Collection - FIRE Migration

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/183804703

- Implement single file delete operations
- Add FIRE delete operation
- Improve file utils operation delete name
- Remove duplicated database operation
- Implement operation to get the latest inactive version


submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/files/service/GroupFilesService.kt
```diff
@@ -41,7 +41,7 @@ class GroupFilesService {
             .resolve(path)
             .resolve(fileName)
             .toFile()
-        FileUtils.deleteFile(file)
+        FileUtils.deleteIfExists(file)
```
Juan-EBI, line 44: I think we should not do make the api ignore error. (someone trying to delete a file that does not exists).  Is ok and IMO expected that api fail if someone request to delete a file that does not exits.

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/files/service/UserFilesService.kt
```diff
@@ -50,6 +50,6 @@ class UserFilesService {
         val userPath = user.magicFolder.path
         val userFile = userPath.resolve(path).resolve(fileName).toFile()
         require(userPath != userFile.toPath()) { "Can not delete user root folder" }
-        FileUtils.deleteFile(userFile)
+        FileUtils.deleteIfExists(userFile)
```
Juan-EBI, line 53: same as befllow

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/wiremock/FireMockDatabase.kt
```diff
@@ -47,6 +47,15 @@ class FireMockDatabase(
         }
     }
 
+    fun delete(fireOid: String) {
+        val record = recordsById.getValue(fireOid)
+        if (record.fileSystemPath != null) {
```
Juan-EBI, line 52: we should not call delete on a file that does not have path? does we? if is not a valid scenario lests not handle issue in mock but rather let it fail so we identify any incosistency in the backend

submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/db/repositories/Repositories.kt
```diff
@@ -35,7 +35,7 @@ interface SubmissionMongoRepository : MongoRepository<DocSubmission, ObjectId> {
 
     fun getByAccNoInAndVersionGreaterThan(accNo: List<String>, version: Int): List<DocSubmission>
 
-    fun findFirstByAccNoOrderByVersionDesc(accNo: String): DocSubmission?
+    fun findFirstByAccNoAndVersionLessThanOrderByVersion(accNo: String, version: Int = 0): DocSubmission?
```
Juan-EBI, line 38: why sorting by AccNo?

submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/api/FilesService.kt
```diff
@@ -9,4 +9,10 @@ internal interface FilesService {
     fun postProcessSubmissionFiles(sub: ExtSubmission)
 
     fun cleanSubmissionFiles(previous: ExtSubmission, current: ExtSubmission?)
+
+    fun deleteSubmissionFiles(sub: ExtSubmission)
+
+    fun deleteSubmissionFile(sub: ExtSubmission, file: ExtFile)
```
Juan-EBI, line 15: having the same method two times make me think that there are two clases implementing the same logic. as 

 deleteSubmissionFiles(sub: ExtSubmission) => Iterate and call deleteSubmissionFile(sub: ExtSubmission, file: ExtFile)
 
 So if we offer ` fun deleteSubmissionFiles(sub: ExtSubmission)` is because another class is doing exactly the same iteration and deleteing all files.  I hard to see it right now but it looks to me we need a single abtractions and either caller class control flow or this class control it.

Juan-EBI (commented):
Please check comments

submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/db/repositories/Repositories.kt
jhoanmanuelms: It's not sorting by accNo but by version

submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/api/FilesService.kt
jhoanmanuelms: yes, the other one will disappear in the next PR

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/wiremock/FireMockDatabase.kt
jhoanmanuelms: you're right, we in the system won't (or at least shouldn't) call delete over files without a path. However, this is a mock implementation of FIRE where you actually can call delete over any file regardless of whether it has a path or not, and isn't that the purpose of a mock? being as close as possible to the real thing? 🤔

Juan-EBI: I see your point and agree, hard to balance beetween real behavior and error detection, what I was trying to said is that is good that mock fail if we call delete on file without path. It will helps us to detect an unwanted behavior but yes is not real.

Juan-EBI (approved):
LGTM

PR #684, opened 2023-02-15
Pivotal ID # 183628268: Improve Backend Docs

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/183628268

- Hide internal operations that were public
- Separate the submission operations into internal/public


Juan-EBI (approved):
LGTM

PR #685, opened 2023-02-15
Pivotal ID # 183070739: Fix Backend Log Rotation

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/183070739

Preserve current logs on application start up


Juan-EBI (approved):
LGTM

PR #691, opened 2023-02-28
Pivotal ID # 184485190: S3 Client For FIRE Download Operations

Juan-EBI:
https://www.pivotaltracker.com/story/show/184485190
- Removed download by id operation


client/fire-webclient/src/main/kotlin/uk/ac/ebi/fire/client/api/FireWebClient.kt
```diff
@@ -48,20 +48,19 @@ internal class FireWebClient(
 
     override fun downloadByPath(
         path: String,
-    ): File? = findByPath(path)?.let {
-        downloadFireFile(path.substringAfterLast("/"), "/objects/blob/path/$path")
+    ): File? {
+        return downloadFireFile(path.substringAfterLast("/"), "/objects/blob/path/$path")
```
jhoanmanuelms, line 52: I don't think this change will solve the issue since this endpoint will also be removed. Check this [slack thread](https://biostudies.slack.com/archives/C012F607YBA/p1676371176264299?thread_ts=1676368255.206079&cid=C012F607YBA) for more info.

jhoanmanuelms (changes requested):
Please take a look at my comment

client/fire-webclient/src/main/kotlin/uk/ac/ebi/fire/client/api/FireWebClient.kt
Juan-EBI: yes I am aware of this, but I prefer to change to path and then s3 than a single jump (with more changes and risks)

jhoanmanuelms: Ok, got it

jhoanmanuelms (approved):
LGTM

PR #692, opened 2023-03-01
Pivotal ID # 184596586: Update Common Files Deletion Index

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/184596586

Update the current index for the delete common files operation


Juan-EBI (changes requested):
I got concerns about this solution,  we are showing some kind of progress but we do not know the numbers of records to process so this is equivalent of saying "we have process 10 files of unknow" which in my opinion does not add good value it reminde me Windows XP copy/download progress.

I think we need to desing a provide  most robust solution

PR #693, opened 2023-03-02
Pivotal ID # 183688670: Structure Backend Notifications

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/183688670

- Submit notifications independently for PMC / scheduler notifications
- Remove "failed log" queues
- Remove unused code


commons/commons-http/src/main/kotlin/ebi/ac/uk/commons/http/slack/NotificationsSender.kt
```diff
@@ -6,7 +6,7 @@ import org.springframework.http.MediaType.APPLICATION_JSON
 import org.springframework.web.client.RestTemplate
 import org.springframework.web.client.postForEntity
 
-class NotificationsSender(
+open class NotificationsSender(
```
Juan-EBI, line 9: I think rather than introduce extra classes and inhirantece we can use Spring @Qualifer to tag each bean diferently.

PR #699, opened 2023-03-22
Pivotal ID # 184772659: Delete only common files when required on re submission

Juan-EBI:
https://www.pivotaltracker.com/story/show/184772659

### DIFF
--- a/submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/request/SubmissionRequestCleaner.kt
+++ b/submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/request/SubmissionRequestCleaner.kt
@@ -7,6 +7,8 @@ import ac.uk.ebi.biostd.persistence.common.service.SubmissionRequestPersistenceS
 import ac.uk.ebi.biostd.persistence.filesystem.api.FileStorageService
 import ebi.ac.uk.extended.model.ExtFile
 import ebi.ac.uk.extended.model.ExtSubmission
+import ebi.ac.uk.extended.model.StorageMode
+import ebi.ac.uk.extended.model.storageMode
 import mu.KotlinLogging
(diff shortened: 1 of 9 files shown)


submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/request/SubmissionRequestCleaner.kt
```diff
@@ -39,18 +40,28 @@ class SubmissionRequestCleaner(
...
+            when (val newFile = newFiles[existing.filePath]) {
+                null -> false
+                else -> newFile.md5 != existing.md5 && new.storageMode == existing.storageMode
+            }
+
         val newFiles = newFilesMap(new)
         logger.info { "${current.accNo} ${current.owner} Started cleaning common submission files" }
         serializationService.fileSequence(current)
-            .filter { newFiles.containsKey(it.filePath) && newFiles[it.filePath] != it.md5 }
+            .filter { shouldDelete(newFiles, it) }
```
Juan-EBI, line 55: previous mechanism delete file regardless if the storage type is the same or not resulting in deletion of importants files as page tab ones.

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/request/SubmissionRequestCleaner.kt
```diff
@@ -28,7 +30,6 @@ class SubmissionRequestCleaner(
         if (current != null) {
             logger.info { "${new.accNo} ${new.owner} Started cleaning common files of version ${new.version}" }
             deleteCommonFiles(new, current)
-            storageService.deleteFtpLinks(current)
```
Juan-EBI, line 31: previos changes deleted all ftp folder resulting in files unavaialbiltiy even when not need it.

PR #701, opened 2023-03-28
Pivotal ID # 184812856: Regenerate pagetab operation 

Juan-EBI:
https://www.pivotaltracker.com/story/show/184812856
- added file list pagetab generation endpoints

### DIFF
--- a/submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/db/repositories/Repositories.kt
+++ b/submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/db/repositories/Repositories.kt
@@ -92,6 +92,7 @@ interface SubmissionDraftRepository : MongoRepository<DocSubmissionDraft, String
 }
 
 interface FileListDocFileRepository : MongoRepository<FileListDocFile, ObjectId> {
+    @Meta(cursorBatchSize = 100, flags = [CursorOption.NO_TIMEOUT])
     fun findAllBySubmissionAccNoAndSubmissionVersionGreaterThanAndFileListName(
         accNo: String,
         version: Int,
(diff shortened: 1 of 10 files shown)


submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/db/repositories/Repositories.kt
```diff
@@ -92,6 +92,7 @@ interface SubmissionDraftRepository : MongoRepository<DocSubmissionDraft, String
 }
 
 interface FileListDocFileRepository : MongoRepository<FileListDocFile, ObjectId> {
+    @Meta(cursorBatchSize = 100, flags = [CursorOption.NO_TIMEOUT])
```
jhoanmanuelms, line 95: 👍

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/service/ExtSubmissionQueryService.kt
```diff
@@ -22,7 +23,12 @@ class ExtSubmissionQueryService(
     fun getReferencedFiles(accNo: String, fileListName: String): ExtFileTable {
         val sub = submissionPersistenceQueryService.getExtByAccNo(accNo, false)
         val files = submissionPersistenceQueryService.getReferencedFiles(sub, fileListName)
-        return ExtFileTable(files)
+        return ExtFileTable(files.toList())
+    }
+
+    fun getReferencedFile(accNo: String, fileListName: String): Sequence<ExtFile> {
```
jhoanmanuelms, line 29: I think this should be called `getReferencedFilesAsSequece` or `getReferencedFilesSequece` since the only difference is the returned type and the current name (singular) gives the impression that this method is used to get a single file

Also, I know there're some places in the code were we use the function above (getReferencedFiles) just to get the list. I mean, we're calling the function, getting an `ExtFileTable` object, and then getting the list. Maybe it would be worth taking a look at these since it might make more sense to use this new function that returns a sequence and consume it accordingly.

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/web/resources/SubmissionQueryResource.kt
```diff
@@ -26,20 +32,47 @@ class SubmissionQueryResource(
...
 
     @GetMapping("/{accNo}.tsv", produces = [TEXT_PLAIN])
-    fun asTsv(@PathVariable accNo: String) = submissionService.getSubmissionAsTsv(accNo)
+    fun asTsv(@PathVariable accNo: String) = submissionService.getSubmission(accNo, SubFormat.TSV)
+
+    @GetMapping("/{accNo}/{fileList}.tsv")
+    fun asTsv(
+        @PathVariable accNo: String,
+        @PathVariable fileList: String,
+    ): ResponseEntity<Resource> = fileListFile(accNo, fileList, SubFormat.TSV)
```
jhoanmanuelms, line 47: 👍

jhoanmanuelms (approved):

PR #702, opened 2023-04-25
Pivotal ID # 184975504: S3 submission resubmission

Juan-EBI:
https://www.pivotaltracker.com/story/show/184975504

### DIFF
--- a/client/fire-webclient/src/main/kotlin/uk/ac/ebi/fire/client/api/S3Client.kt
+++ b/client/fire-webclient/src/main/kotlin/uk/ac/ebi/fire/client/api/S3Client.kt
@@ -1,0 +1,31 @@
+package uk.ac.ebi.fire.client.api
+
+import com.amazonaws.services.s3.AmazonS3
+import com.amazonaws.services.s3.model.GetObjectRequest
+import com.amazonaws.services.s3.model.S3Object
+import uk.ac.ebi.fire.client.integration.web.FireS3Client
+import java.io.File
+import java.nio.file.Files
+import java.nio.file.StandardCopyOption
+
+
+class S3Client(
+    private val amazonS3Client: AmazonS3,
+    private val bucketName: String,
+) : FireS3Client {
+    override fun downloadByPath(path: String): File? {
+        val stream = getFireObjectByPath(path).objectContent
+        val tmp = File.createTempFile(DEFAULT_PREFIX, path.substringAfterLast("/"))
+        Files.copy(stream, tmp.toPath(), StandardCopyOption.REPLACE_EXISTING)
+        return tmp
+    }
+
+    private fun getFireObjectByPath(path: String?): S3Object {
+        val getObjectRequest = GetObjectRequest(bucketName, path)
+        return amazonS3Client.getObject(getObjectRequest)
+    }
+
+    companion object {
+        const val DEFAULT_PREFIX = "tmp_s3_file"
+    }
+}
--- a/client/fire-webclient/src/main/kotlin/uk/ac/ebi/fire/client/integration/web/RetryWebClient.kt
+++ b/client/fire-webclient/src/main/kotlin/uk/ac/ebi/fire/client/integration/web/RetryWebClient.kt
@@ -7,7 +7,8 @@ import java.io.File
 
 @Suppress("TooManyFunctions")
 internal class RetryWebClient(
-    private val fireClient: FireClient,
+    private val fireClient: FireWebClient,
+    private val fireS3Client: FireS3Client,
     private val template: RetryTemplate,
 ) : FireClient {
     override fun save(file: File, md5: String, size: Long): FireApiFile {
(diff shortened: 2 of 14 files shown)


submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/SubmissionStorageModeTest.kt
```diff
@@ -142,11 +151,15 @@ class SubmissionStorageModeTest(
...
     @Test
     fun `10-4 transfer from FIRE to NFS`() {
-        val submission = createSubmission("S-STR-MODE-4")
-
+        val (submission, file, fileList, fileListFile) = createSubmission("S-STR-MODE-4")
+        webClient.uploadFiles(listOf(file, fileListFile, fileList))
         assertThat(webClient.submitSingle(submission, TSV, FIRE)).isSuccessful()
         val fireSub = submissionRepository.getExtByAccNo("S-STR-MODE-4", includeFileListFiles = true)
 
+        // We upload files manually into S3 to simulate fire S3 support.
```
jhoanmanuelms, line 159: Just to clarify, is this to validate the case where the imaging team uploads the files directly to FIRE? If so, I don't think this is the right place to validate this logic since the idea of the transfer function is for submissions that already exist and will be transferred as-is so it seems weird to me we're adding files afterward.

I think the proper test case for this would be [here](https://github.com/EBIBioStudies/biostudies-backend-services/pull/702/files#diff-f291be59aef177e9e2d92fe673f8885390fe05251ff8400078e5b6c9e97d98b8R73) like this:
1. Submit to FIRE
2. Upload extra files manually
3. Resubmit to NFS including the manually uploaded files in the new version

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/validator/collection/EuToxRiskValidator.kt
```diff
@@ -12,13 +13,17 @@ import org.springframework.http.HttpHeaders
...
+import ebi.ac.uk.extended.model.ExtFile as ExtFile1
 
 internal const val EXCEL_FILE_REQUIRED = "Excel file is required for Eu-ToxRisk submissions"
 internal const val SKIP_VALIDATION_ATTR = "QMRF-ID"
 
 class EuToxRiskValidator(
     private val restTemplate: RestTemplate,
-    private val validationProperties: ValidatorProperties
+    private val validationProperties: ValidatorProperties,
+    private val fireClient: FireClient,
```
jhoanmanuelms, line 26: 👍

client/fire-webclient/src/main/kotlin/uk/ac/ebi/fire/client/integration/web/RetryWebClient.kt
```diff
@@ -25,11 +26,6 @@ internal class RetryWebClient(
         template.execute(opt) { fireClient.unsetPath(fireOid) }
     }
 
-    override fun downloadByPath(path: String): File? {
-        val opt = "Download file path='$path'"
-        return template.execute(opt) { fireClient.downloadByPath(path) }
```
jhoanmanuelms, line 30: Since we're removing this, I think we should also remove the [http client method](https://github.com/EBIBioStudies/biostudies-backend-services/blob/master/client/fire-webclient/src/main/kotlin/uk/ac/ebi/fire/client/api/FireWebClient.kt#L49) and the related tests since it won't be required anymore

client/fire-webclient/src/main/kotlin/uk/ac/ebi/fire/client/api/S3Client.kt
```diff
@@ -0,0 +1,30 @@
...
+
+import com.amazonaws.services.s3.AmazonS3
+import com.amazonaws.services.s3.model.GetObjectRequest
+import com.amazonaws.services.s3.model.S3Object
+import uk.ac.ebi.fire.client.integration.web.FireS3Client
+import java.io.File
+import java.nio.file.Files
+import java.nio.file.StandardCopyOption
+
+class S3Client(
```
jhoanmanuelms, line 11: I think we're missing unit tests for this client

jhoanmanuelms (approved):
LGTM. Please consider my comments before merging

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/SubmissionStorageModeTest.kt
Juan-EBI: No, this is not because of this. This changes are because we need to simulate fire mechanism that make every file uploaded through http it make it avaiable in s3.

Juan-EBI: wonderinf if the comment need to be better to clarify this

PR #703, opened 2023-05-02
Pivotal ID # 184779579: Match generated zip with user folders Collapse

Juan-EBI:
https://www.pivotaltracker.com/story/show/184779579
- added file type to serialization


Juan-EBI:
@jhoanmanuelms main change here is that we will use file type attribute to search for directories with the same name in user folder when re submitting in this case we "detect" any change there when re submitting.

jhoanmanuelms (approved):
LGTM

PR #705, opened 2023-05-10
Pivotal ID # 185138199: User Files List Directories Size

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/185138199

- Add an option to avoid internal directory calculation size
- Avoid directory calculation for the user space list endpoint


Juan-EBI (approved):
LGTM please consider  comments

PR #707, opened 2023-05-11
Pivotal ID # 184822139: Delete Submission Folder After Resubmission

Juan-EBI:
https://www.pivotaltracker.com/story/show/184822139
- simplified code to have a single delete files operation.

### DIFF
--- a/submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/service/SubmissionService.kt
+++ b/submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/service/SubmissionService.kt
@@ -2,13 +2,13 @@ package ac.uk.ebi.biostd.submission.domain.service
 
 import ac.uk.ebi.biostd.persistence.common.service.SubmissionPersistenceQueryService
 import ac.uk.ebi.biostd.persistence.common.service.SubmissionPersistenceService
+import ac.uk.ebi.biostd.persistence.filesystem.api.FileStorageService
 import ac.uk.ebi.biostd.submission.exceptions.UserCanNotDelete
 import ac.uk.ebi.biostd.submission.exceptions.UserCanNotRelease
 import ac.uk.ebi.biostd.submission.model.ReleaseRequest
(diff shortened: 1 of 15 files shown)


submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/service/SubmissionService.kt
```diff
@@ -24,8 +24,8 @@ class SubmissionService(
     private val extSubmissionSubmitter: ExtSubmissionSubmitter,
     private val submissionSubmitter: SubmissionSubmitter,
     private val eventsPublisherService: EventsPublisherService,
-    private val fileStorageService: FileStorageService,
     private val submissionPersistenceService: SubmissionPersistenceService,
+    private val submissionRequestFinalizer: SubmissionRequestFinalizer,
```
jhoanmanuelms, line 28: I don't like very much the idea of including the `SubmissionRequestFinalizer` here since it's part of the submission workflow

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/request/SubmissionRequestFinalizer.kt
```diff
@@ -30,22 +30,29 @@ class SubmissionRequestFinalizer(
         return sub
     }
 
-    private fun deleteRemainingFiles(sub: ExtSubmission, previous: ExtSubmission) {
+    fun deleteSubmissionFiles(submission: ExtSubmission) {
+        deleteRemainingFiles(null, submission)
+    }
+
+    private fun deleteRemainingFiles(current: ExtSubmission?, previous: ExtSubmission) {
```
jhoanmanuelms, line 37: I think this is the logic that should be moved to the `FileStorageService` and then be used in both places since it makes more sense to me to have a storage services that deletes files that later is used by both, the RequestFinalizer and the SubmissionService than the SubmissionService using the RequestFinalizer directly

PR #709, opened 2023-05-17
Pivotal ID # 184822139: Delete Submission Folder After Resubmission

Juan-EBI:
https://www.pivotaltracker.com/story/show/184822139
- added delete files operations


commons/commons-util/src/main/kotlin/ebi/ac/uk/io/FileUtils.kt
```diff
@@ -72,6 +73,15 @@ object FileUtils {
         }
     }
 
+    fun deleteEmptyDirectories(dir: File) {
+        dir
+            .walkBottomUp()
+            .asSequence()
+            .takeWhile { it != dir }
+            .forEach { deleteEmptyDirectories(it) }
+        if (dir.isDirectory && dir.isEmpty()) Files.delete(dir.toPath())
```
jhoanmanuelms, line 82: I didn't know this way of doing recursion 👍

submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/api/FileStorageService.kt
```diff
@@ -11,5 +11,9 @@ interface FileStorageService {
 
     fun deleteSubmissionFile(sub: ExtSubmission, file: ExtFile)
 
-    fun deleteSubmissionFiles(sub: ExtSubmission)
+    fun deleteSubmissionFiles(
+        sub: ExtSubmission,
+        onEachIndexed: (Int, ExtFile) -> Unit = { _, _ -> },
```
jhoanmanuelms, line 16: I understand the idea behind this parameter but I'm not sure we should have it since I think it makes the signature and the code more complicated than it should be. I think that since the operation is specifically for deleting and we're currently only using this parameter for logging, we'd be better off leaving the logging to the implementation of this method so it would be just

```
fun deleteSubmissionFiles(sub: ExtSubmission, filter: (ExtFile) -> Boolean = { true })
```

(The comment continues below 1/3)

submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/service/StorageService.kt
```diff
@@ -49,7 +49,20 @@ class StorageService(
...
     }
 
-    override fun deleteSubmissionFiles(sub: ExtSubmission) {
-        serializationService.fileSequence(sub).forEach { file -> deleteSubmissionFile(sub, file) }
+    override fun deleteSubmissionFiles(
+        sub: ExtSubmission,
+        onEachIndexed: (Int, ExtFile) -> Unit,
+        filter: (ExtFile) -> Boolean,
+    ) {
+        serializationService.fileSequence(sub)
```
jhoanmanuelms, line 57: Here we add the logging since it's what is actually doing, deleting

(Continues below 2/3)

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/request/SubmissionRequestFinalizer.kt
```diff
@@ -31,16 +30,15 @@ class SubmissionRequestFinalizer(
...
-            logger.info { "${previous.accNo} ${previous.owner} Deleting file $index, path='${file.filePath}'" }
-            storageService.deleteSubmissionFile(previous, file)
-        }
-
         val subFiles = subFilesSet(current)
         logger.info { "${previous.accNo} ${previous.owner} Started deleting remaining submission files" }
-        serializationService.fileSequence(previous)
-            .filter { subFiles.contains(it.filePath).not() || it.storageMode != current?.storageMode }
-            .forEachIndexed { index, file -> deleteFile(index, file) }
+        storageService.deleteSubmissionFiles(
```
jhoanmanuelms, line 35: And finally here, we would just call

```
storageService.deleteSubmissionFiles(
    sub = previous, 
    filter = { subFiles.contains(it.filePath).not() || it.storageMode != current?.storageMode },
)
```

WDYT? I think it makes the code easier to follow 

(3/3)

jhoanmanuelms (approved):
LGTM. Please take my comment into consideration

PR #710, opened 2023-05-22
Pivotal # 182595703: Submission Releasing Workflow

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/182595703

- Implement the restrictions for the release date
- Create iTests for un-releasing
- Adjust iTests to use current date for public submissions instead of using a past release date


submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/factory/AllInOneSubmissionJson.kt
```diff
@@ -21,7 +23,7 @@ fun allInOneSubmissionJson(accNo: String) = jsonObj {
         },
         {
             "name" to "ReleaseDate"
-            "value" to "2021-02-12"
+            "value" to OffsetDateTime.now().toStringDate()
```
Juan-EBI, line 26: IMO using evaluated expression in itest make it harder to follow and also may generate issues depending of computer time zone and hour. I think using fixed value make easier to follow as is super clear for example that 2021 is the past of 2040 is in the future.

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/ResubmissionApiTest.kt
```diff
@@ -180,4 +187,65 @@ class ResubmissionApiTest(
...
+            line("Submission", "S-RSTST4")
+            line("Title", "Public Submission")
+            line("ReleaseDate", OffsetDateTime.now().toStringDate())
+            line()
+            line("Study")
+            line()
+        }.toString()
+
+        val regularUser = RegularUser.email
+        val onBehalfClient = SecurityWebClient
```
Juan-EBI, line 203: is not clear why we mix the onBehalf functionally here, I think we should simple use a regular use client.

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/ResubmissionApiTest.kt
```diff
@@ -180,4 +187,65 @@ class ResubmissionApiTest(
...
+        val version1 = tsv {
+            line("Submission", "S-RSTST5")
+            line("Title", "Public Submission")
+            line("ReleaseDate", OffsetDateTime.now().toStringDate())
+            line()
+            line("Study")
+            line()
+        }.toString()
+
+        val onBehalfClient = SecurityWebClient
```
Juan-EBI, line 234: I think it should be only superUserClient

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/service/TimesService.kt
```diff
@@ -14,16 +18,38 @@ import java.time.ZoneOffset.UTC
...
 
+    private fun releaseTime(
+        now: OffsetDateTime,
+        releaseDate: String,
+        request: SubmitRequest,
+    ): OffsetDateTime {
+        val releaseTime = parseDate(releaseDate)
+        val accNo = request.accNo
+        val user = request.submitter.email
+        val today = now.toInstant().asOffsetAtStartOfDay()
```
Juan-EBI, line 41: lets said is 2pm accotrding to this if we try to release to same day at 8 am (in the past) code still will fail which does not look correct.

Juan-EBI (commented):

Juan-EBI (commented):

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/factory/AllInOneSubmissionJson.kt
jhoanmanuelms: I understand your point but after adding the restriction of not allowing release dates in the past, the only way to make a submission public right away would be to use the same day. I also thought about mocking the date but then it wouldn't be a real iTest.

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/service/TimesService.kt
jhoanmanuelms: The exact time isn't taken into consideration for the release date since it's [currently parsed](https://github.com/EBIBioStudies/biostudies-backend-services/blob/master/submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/service/TimesService.kt#L30) always at the start of the day [regardless if it contains the time](https://github.com/EBIBioStudies/biostudies-backend-services/blob/master/submission/submitter/src/test/kotlin/ac/uk/ebi/biostd/submission/service/TimesServiceTest.kt#L74)

Juan-EBI: make sense to me

PR #711, opened 2023-05-22
Pivotal ID # 184812856: Regenerate pagetab operation 

Juan-EBI:
https://www.pivotaltracker.com/story/show/184812856


submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/SubmissionRefreshApiTest.kt
```diff
@@ -0,0 +1,219 @@
...
+import org.springframework.data.mongodb.core.query.Query
+import org.springframework.data.mongodb.core.query.Update.update
+import org.springframework.test.context.junit.jupiter.SpringExtension
+import java.time.LocalDate
+import java.time.ZoneOffset.UTC
+
+@Import(FilePersistenceConfig::class)
+@ExtendWith(SpringExtension::class)
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+class SubmissionRefreshApiTest(
```
jhoanmanuelms, line 63: I think we should continue using the numbered format for the tests and keep adding them to the test inventory as we create them. I've found it useful and I think it's a good practice

jhoanmanuelms (approved):

PR #713, opened 2023-05-25

jhoanmanuelms:

This is the initial PR just to move all the logic into a single package so we have everything in the same place. The plan is:

1. This PR to re-structure everything
2. Refactor out the database method used to query files from the database into a separate class so we have the database access as isolated as possible. I know having db access here is not ideal but I think removing the need for it is out of the scope of this ticket
3. Create a class that works as single entry point for the files and use only this one. All the other implementations should be internal
4. Finally, apply the validation for the file path in this single entry point

The changes contained in this PR are:
- Create a new module in order to contain all the file sources related logic
- Refactor all the file sources related logic and tests into the new module
- Remove unused code


Juan-EBI (approved):
nice refactor

PR #714, opened 2023-05-26
Pivotal ID # 185273996: Fire s3 path should be relative

Juan-EBI:
https://www.pivotaltracker.com/story/show/185273996
- removed slash prefix


Juan-EBI:
@jhoanmanuelms as Fire ignores slash prefix instead of fixing this manually we are basically sending paths without it.

jhoanmanuelms (approved):
LGTM

PR #715, opened 2023-05-31
Pivotal ID # 185165390: Avoid Empty FIRE ByPass Attributes

Juan-EBI:
https://www.pivotaltracker.com/story/show/185165390


jhoanmanuelms (approved):
I think this is a bit clearer. I think we should merge this ASAP in order to avoid issues with the FIRE bypass feature

PR #716, opened 2023-05-31
Pivotal ID # 185273996: Fire s3 path should be relative

jhoanmanuelms:
- Remove the initial slash from FIRE responses
- Normalize FIRE mock output to mimic the original service response

### DIFF
--- a/client/fire-webclient/src/main/kotlin/uk/ac/ebi/fire/client/model/FireApiFile.kt
+++ b/client/fire-webclient/src/main/kotlin/uk/ac/ebi/fire/client/model/FireApiFile.kt
@@ -10,6 +10,7 @@ data class FireApiFile(
     val createTime: String,
     val filesystemEntry: FileSystemEntry? = null,
 ) {
+    // The path shouldn't contain an initial slash in order to avoid inconsistencies with S3
     val path: String? = filesystemEntry?.path?.removePrefix("/")
     val published: Boolean = filesystemEntry?.published.orFalse()
 }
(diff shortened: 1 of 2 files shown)


client/fire-webclient/src/main/kotlin/uk/ac/ebi/fire/client/model/FireApiFile.kt
```diff
@@ -10,7 +10,7 @@ data class FireApiFile(
     val createTime: String,
     val filesystemEntry: FileSystemEntry? = null,
 ) {
-    val path: String? = filesystemEntry?.path
+    val path: String? = filesystemEntry?.path?.removePrefix("/")
```
Juan-EBI, line 13: I think is worth it to document what we do this at class level or property level.

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/wiremock/FireMockDatabase.kt
```diff
@@ -75,5 +75,5 @@ data class DbRecord(
     val firePath: String?,
     val published: Boolean,
 ) {
-    fun toFile(): FireApiFile = file.copy(filesystemEntry = FileSystemEntry(firePath, published))
+    fun toFile(): FireApiFile = file.copy(filesystemEntry = FileSystemEntry(firePath?.let { "/$it" }, published))
```
Juan-EBI, line 78: same as bellow

Juan-EBI (approved):

PR #719, opened 2023-06-07
Pivotal ID # 182595703: Submission Suppressing Workflow

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/182595703

- Have the previous release time into consideration
- Throw a single clearer exception when a regular user tries to update the release date of a public study


Juan-EBI (approved):
Looks good for me. It looks to me that we need Itest in regards this feature. Personally I think Itest give us way more value that unit test. 

1. Change release date from future to future (Allowed)
2. Change release date from past (Failure)
3. Change files in released submission (Failure)
3. Change metada in  released  submission (Allowed)

PR #720, opened 2023-06-08
Pivotal ID # 184724563: Use a specific user and server for the update script

awaisathar:


deploy.gradle
```diff
@@ -133,6 +133,8 @@ task deploy {
                 execute "sed -i -e 's~APP_PORT~$appPort~g' $deployPath/update.sh"
                 execute "sed -i -e 's~APP_NAME~$artifactName~g' $deployPath/update.sh"
                 execute "sed -i -e 's~DEBUG_PORT~$debugPort~g' $deployPath/update.sh"
+                execute "sed -i -e 's~BS_USER~`whoami`~g' $deployPath/update.sh"
```
jhoanmanuelms, line 136: Hi @awaisathar. I think we need to check the replacing command since it isn't actually replacing with the expected value in the script on the server.

You can test this by running [this gradle command](https://github.com/EBIBioStudies/biostudies-backend-services/blob/master/ci/submitter.yml#L4) to deploy on dev:

```
./gradlew :submission:submission-webapp:bootJar deploySubmitter -Penv=dev -PdeployPath=$APPS_PATH/submitter/dev
```

The caveat is you'd need to export the required variables in your local console. Another way to test would be to push the changes to gitlab and test it from the deploy task there

jhoanmanuelms (changes requested):

jhoanmanuelms (approved):
LGTM

deploy.gradle
jhoanmanuelms: Works perfectly now. Thanks @awaisathar 😁👍

PR #722, opened 2023-06-12
Pivotal ID # 185110100: Case Insensitive Submission Ownership

jhoanmanuelms:
- Ensure user email is always lowercase on registration
- Remove unnecessary data classes


Juan-EBI (approved):
LGTM

commons/commons-bio/src/main/kotlin/ebi/ac/uk/api/security/RegisterRequest.kt
```diff
@@ -19,3 +24,17 @@ class RegisterRequest(
...
+    @JsonDeserialize(using = EmailDeserializer::class)
+    @field:Email(message = "The provided email is not valid")
+    val userEmail: String,
+
+    val userName: String,
+)
+
+class EmailDeserializer : StdDeserializer<String>(String::class.java) {
+    override fun deserialize(p: JsonParser?, ctxt: DeserializationContext?): String {
+        return _parseString(p, ctxt).lowercase()
```
Juan-EBI, line 38: not sure about adding case insensitve check in web layer? should we add this servie layer directly? (so if we change how we call this, rule still in place)

Juan-EBI (commented):
please consider comment before merge

PR #725, opened 2023-06-15

jhoanmanuelms:

- Make all the FilesSource implementations package private
- Create a builder that allows the creation of a FilesSourceList
- Force usage of FilesSourceList only as the way to use the files sources


Juan-EBI (approved):
LGTM

PR #727, opened 2023-06-21

jhoanmanuelms:
Implement file path restrictions


Juan-EBI (approved):
LGTM please take a look to minor comments

PR #728, opened 2023-06-29

jhoanmanuelms:

- Migrate BioWebClient operations from RestTemplate to WebClient
- Move common operations to http-commons package and reuse it


client/bio-webclient/src/main/kotlin/ac/uk/ebi/biostd/client/extensions/WebClientExt.kt
```diff
@@ -0,0 +1,19 @@
+package ac.uk.ebi.biostd.client.extensions
+
+import ac.uk.ebi.biostd.client.integration.web.SubmissionResponse
+import ac.uk.ebi.biostd.integration.SerializationService
+import ac.uk.ebi.biostd.integration.SubFormat
+import org.springframework.web.reactive.function.client.WebClient.RequestHeadersSpec
+
+fun RequestHeadersSpec<*>.submitBlocking(
```
Juan-EBI, line 8: I do not think this extensions is the right abtractions basically because RequestHeadersSpec is a spring generic class, so we are populating the scope of this class with something specific.  Is like adding `toAccNo` to String class, clearly it will work but String class should not coupled to something like biostudies.  I think extension should live on serializacionService side. (something that do is specfiic to biostudies)

```
fun SerializationService.deserializeResponse(
    response: WebClient.ResponseSpec,
    format: SubFormat,
): Mono<SubmissionResponse> {
    fun asSubmission(response: ResponseEntity<String>): SubmissionResponse {
        val submission = deserializeSubmission(response.body!!, format)
        return SubmissionResponse(submission, response.statusCodeValue)
    }
    return response.toEntity(String::class.java).map { asSubmission(it) }
}

```

Juan-EBI (approved):
LGTM please check comment

PR #729, opened 2023-07-03

jhoanmanuelms:

- Migrate the rest of modules to WebClient
- Unify common code
- Remove unnecessary logic from the blocking common code
- Improve unit tests


Juan-EBI (approved):
LGTM only one question, should be able to remove spring web dependency now? (dependency that provide Rest Template)

PR #732, opened 2023-07-06

jhoanmanuelms:

Use constants for plugin dependencies in order to have them centralized for easy maintenance. This will be later required for upgrading Kotlin in order to use the Kotlin S3 Client

### DIFF
--- a/submission/submitter/build.gradle.kts
+++ b/submission/submitter/build.gradle.kts
@@ -23,8 +23,8 @@ import io.spring.gradle.dependencymanagement.dsl.DependencyManagementExtension
 import org.springframework.boot.gradle.plugin.SpringBootPlugin.BOM_COORDINATES
 
 plugins {
-    id("io.spring.dependency-management") version "1.0.12.RELEASE"
-    id("org.springframework.boot") version "2.7.1" apply false
+    id(Plugins.SpringBootPlugin) version PluginVersions.SpringBootPluginVersion apply false
+    id(Plugins.SpringDependencyManagementPlugin) version PluginVersions.SpringDependencyManagementPluginVersion
 }
 
 the<DependencyManagementExtension>().apply {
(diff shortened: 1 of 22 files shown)


submission/submitter/build.gradle.kts
```diff
@@ -23,8 +23,8 @@ import io.spring.gradle.dependencymanagement.dsl.DependencyManagementExtension
 import org.springframework.boot.gradle.plugin.SpringBootPlugin.BOM_COORDINATES
 
 plugins {
-    id("io.spring.dependency-management") version "1.0.12.RELEASE"
-    id("org.springframework.boot") version "2.7.1" apply false
+    id(Plugins.SpringBootPlugin) version PluginVersions.SpringBootPluginVersion apply false
```
Juan-EBI, line 26: why this apply false?

jhoanmanuelms: Not sure tbh. In this ticket, I didn't change anything, just converted everything to constants leaving it as it was. The change comes from [this PR](https://github.com/EBIBioStudies/biostudies-backend-services/commit/649177605195087fe83d53b9f9587d42d1d599f9#diff-dbfa181dbf41e4e780c28c8e99a59a499583f784a4fab4436cca2358d4edb1ceR17) and [according to the web](https://stackoverflow.com/a/50153617/2934857) this is used to actually not apply the plugin in the main project but only for sub-projects.

I tried locally removing this property and I get this error:

```
Execution failed for task ':client:fire-webclient:bootJar'.
> Error while evaluating property 'mainClass' of task ':client:fire-webclient:bootJar'
   > Failed to calculate the value of task ':client:fire-webclient:bootJar' property 'mainClass'.
      > Main class name has not been configured and it could not be resolved
```

And if I remove the plugin altogether, it can't resolve the BOM_COORDINATES so I found [this post](https://stackoverflow.com/a/55731664/2934857) which may explain why the `apply false` was used in the first place

PR #733, opened 2023-07-20

jhoanmanuelms:

- Process the files concurrently
- Include a new property to configure the number of files to be processed concurrently
- Retrieve the request files using flows instead of sequences
- Improve error logging in the retry template


submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/nfs/NfsFilesService.kt
```diff
@@ -33,11 +33,13 @@ class NfsFilesService(
     }
 
     private fun persistNfsFile(sub: ExtSubmission, file: NfsFile): ExtFile {
-        val permissions = sub.permissions()
-        val subFile = getSubFile(sub, permissions, file.relPath)
-        if (subFile.notExist()) copyOrReplaceFile(file.file, subFile, permissions)
+        return synchronized(this) {
```
Juan-EBI, line 36: why this need to be synchronized?

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/request/SubmissionRequestProcessor.kt
```diff
@@ -22,27 +30,27 @@ class SubmissionRequestProcessor(
...
+            when (val persisted = storageService.persistSubmissionFile(sub, rqtFile.file)) {
+                rqtFile.file -> requestService.updateRqtIndex(accNo, version, rqtFile.index)
+                else -> requestService.updateRqtIndex(rqtFile, persisted)
             }
+            logger.info { "$accNo ${sub.owner} Finished persisting file ${rqtFile.index}, path='${rqtFile.path}'" }
+        }
+
+        withContext(Dispatchers.Default) {
+            filesRequestService
+                .getSubmissionRequestFiles(accNo, sub.version, startingAt)
```
Juan-EBI, line 50: nice

Juan-EBI (commented):
Looking fantastic, I was expecting that you will remove all blocking operations of webclient as now code support async? is that in a future PR?

submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/nfs/NfsFilesService.kt
jhoanmanuelms: Because the operation to getOrCreate the submission folders is not thread safe

jhoanmanuelms:
Yes @Juan-EBI, the changes to remove blocking operations will come in future PRs

submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/nfs/NfsFilesService.kt
Juan-EBI: I think the problem with this functions are.

1. In FileUtils
```
    fun createFolderIfNotExist(file: Path, permissions: Set<PosixFilePermission>) {
        if (exists(file).not()) createDirectories(file, permissions)
    }
``` 
the problem is that after function check the file does not exists it try to created but other function already did it. I think we can solve this code easily by  ignoring FileAlreadyExists exception in that way function will be threadsafe. Not the cleanest solution but given JDK does not offer createIFNotExists atomic (threadsafe) function is the only thing we can do. 

2. In FileUtils
```
    private fun createDirectory(path: Path, permissions: Set<PosixFilePermission>) {
        Files.createDirectory(path)
        Files.setPosixFilePermissions(path, permissions)
    }
```
This again assume directory does not exists so we can apply the same fix as in 1.


In that way we can remove both syncronization (which will not work when multiple JVM or backends) and do the method threadsafe.

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/request/SubmissionRequestCleaner.kt
```diff
@@ -48,19 +55,26 @@ class SubmissionRequestCleaner(
                 else -> newFile.md5 != existing.md5 && new.storageMode == existing.storageMode
             }
 
-        val newFiles = newFilesMap(new)
+        val newFiles = runBlocking { newFilesMap(new) }
```
Juan-EBI, line 58: not need block function already suspend

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/request/SubmissionRequestCleaner.kt
```diff
@@ -29,15 +36,15 @@ class SubmissionRequestCleaner(
 
         if (current != null) {
             logger.info { "${new.accNo} ${new.owner} Started cleaning common files of version ${new.version}" }
-            deleteCommonFiles(new, current)
+            runBlocking { deleteCommonFiles(new, current) }
```
Juan-EBI, line 39: I think we should make all functions suspend it is already supported in  RabbitListener https://github.com/spring-projects/spring-amqp/pull/2460.  so we can suspend that one and all others downstream. 

If this delay PR we can block main funcitons but keep this one suspend so later is easier to suspend only the  main ones.  (Apply in others places )

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/request/SubmissionRequestCleaner.kt
```diff
@@ -48,19 +55,26 @@ class SubmissionRequestCleaner(
...
     }
 
-    private fun newFilesMap(new: ExtSubmission): Map<String, FileEntry> {
-        return filesRequestService
-            .getSubmissionRequestFiles(new.accNo, new.version, 0)
-            .map { it.file }
-            .associate { it.filePath to FileEntry(it.md5, new.storageMode) }
+    private suspend fun newFilesMap(new: ExtSubmission): Map<String, FileEntry> {
+        var files: Map<String, FileEntry>
+        withContext(Dispatchers.Default) {
```
Juan-EBI, line 68: not need context as function already suspended. (Apply in others places)

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/request/SubmissionRequestCleaner.kt
```diff
@@ -48,19 +55,26 @@ class SubmissionRequestCleaner(
...
+            files = filesRequestService
+                .getSubmissionRequestFiles(new.accNo, new.version, 0)
+                .map { it.file }
+                .buffer(concurrency)
```
Juan-EBI, line 72: I think Buffer here does nothing as eventually list will created with all records.

Juan-EBI (commented):
Looking great please check comments

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/wiremock/FireMockDatabase.kt
```diff
@@ -23,14 +24,18 @@ class FireMockDatabase(
...
-    fun setPath(fireOid: String, firePath: String) {
-        if (recordsById.values.any { it.firePath == firePath })
+    fun setPath(fireOid: String, firePath: String): FireApiFile {
+        if (recordsById.values.any { it.firePath == firePath }) {
             throw FireException("Path '$firePath' is already allocated", CONFLICT)
+        }
 
-        val record = recordsById.getValue(fireOid)
-        recordsById[fireOid] = record.copy(firePath = firePath)
+        val record = recordsById.getValue(fireOid).copy(firePath = firePath)
```
Juan-EBI, line 32: why this changes?  I mean I understand how this implementation will never work for a real world but in our test when we try to run code that generate issues here?

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/wiremock/FireMockDatabase.kt
```diff
@@ -8,11 +8,12 @@ import uk.ac.ebi.fire.client.model.FileSystemEntry
 import uk.ac.ebi.fire.client.model.FireApiFile
 import java.io.File
 import java.time.Instant
+import java.util.Collections
 
 class FireMockDatabase(
     private val fileSystem: FireMockFileSystem,
 ) {
-    private val recordsById: MutableMap<String, DbRecord> = mutableMapOf()
+    private val recordsById: MutableMap<String, DbRecord> = Collections.synchronizedMap(mutableMapOf())
```
Juan-EBI, line 16: I think we should use ConcurrentHashMap check https://www.geeksforgeeks.org/difference-between-concurrenthashmap-and-synchronizedhashmap/

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/wiremock/FireMockDatabase.kt
jhoanmanuelms: The real FIRE api returns a `FireApiFile` when setting the path so I'm expecting that in the `FireWebClient`.  I just made this change in order to keep the mock consistent with the real api

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/request/SubmissionRequestCleaner.kt
jhoanmanuelms: I think it's better to leave these changes for a separate PR immediately after we merge this one and the one where you made the endpoints reactive which I think will cascade up

Juan-EBI (approved):
LGTM

PR #734, opened 2023-07-21
Pivotal ID # 185460059: Retry E-Mails Only Once

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/185460059

- Catch errors on security notifications
- Use spring logs rolling back policy in order to compress the old log files

### DIFF
--- a/submission/submission-handlers/src/main/kotlin/ac/uk/ebi/biostd/handlers/listeners/SecurityNotificationListener.kt
+++ b/submission/submission-handlers/src/main/kotlin/ac/uk/ebi/biostd/handlers/listeners/SecurityNotificationListener.kt
@@ -1,6 +1,12 @@
 package ac.uk.ebi.biostd.handlers.listeners
 
+import ac.uk.ebi.biostd.common.events.BIOSTUDIES_EXCHANGE
+import ac.uk.ebi.biostd.handlers.common.HANDLERS_SUBSYSTEM
+import ac.uk.ebi.biostd.handlers.common.SYSTEM_NAME
+import ac.uk.ebi.biostd.handlers.config.NOTIFICATIONS_FAILED_REQUEST_ROUTING_KEY
 import ac.uk.ebi.biostd.handlers.config.SECURITY_NOTIFICATIONS_QUEUE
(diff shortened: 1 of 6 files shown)


submission/submission-handlers/src/main/kotlin/ac/uk/ebi/biostd/handlers/listeners/SecurityNotificationListener.kt
```diff
@@ -1,25 +1,46 @@
...
     fun receiveMessage(notification: SecurityNotification) {
-        logger.info { "processing ${notification.type} notification for user '${notification.email}'" }
+        logger.info { "Processing ${notification.type} notification for user '${notification.email}'" }
 
-        when (notification.type) {
-            ACTIVATION -> securityNotificationService.sendActivationNotification(notification)
-            ACTIVATION_BY_EMAIL -> securityNotificationService.sendActivationByEmailNotification(notification)
-            PASSWORD_RESET -> securityNotificationService.sendPasswordResetNotification(notification)
-        }
+        runCatching {
```
Juan-EBI, line 31: this is not retrying but rather sending a message when there is failure. Are this the expectaions?

jhoanmanuelms: This is the expected approach. The idea of this ticket is to use the same mechanism that we use for the release notifications to register any failure (which we've historically had very few) in the notifications-failed-log queue instead of keeping retrying indefinitely.

This is actually a very similar approach to the one you suggested where the "parking lot" queue is the failed notifications log, we would just need to add automation to it. I've created [a ticket ](https://www.pivotaltracker.com/story/show/185696865) for it.

Juan-EBI: I was thinking that we were going to actually stop sending message there as no one actually see them but this is the discussed approach that is ok

Juan-EBI (approved):
LGTM

PR #737, opened 2023-08-02
Pivotal ID # 185649200: Cater For Paths Ending With Backslash

jhoanmanuelms:
- Create a common operation to get the file path since all of them should share the same logic
- Remove any trailing slash on file names
- Add integration tests for directories
- Update itests inventory


commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/common/FileCommons.kt
```diff
@@ -0,0 +1,10 @@
+package ac.uk.ebi.biostd.common
+
+import ac.uk.ebi.biostd.validation.InvalidElementException
+import ac.uk.ebi.biostd.validation.REQUIRED_FILE_PATH
+import ebi.ac.uk.base.isNotBlank
+
+fun getFilePath(fileName: String?): String {
+    require(fileName.isNotBlank()) { throw InvalidElementException(REQUIRED_FILE_PATH) }
+    return fileName!!.removeSuffix("/")
```
Juan-EBI, line 9: this looks as a business rule rather than a serialization concern, serialization IMO is dump and only create a class for a serialized format (tsv, json, xml, xlsx).  

I think we should put this in our domain layer all together with the rules that we have (to allow only specific characters) in [FileSourcesList#getExtFile](https://github.com/EBIBioStudies/biostudies-backend-services/blob/32e4328ce1d8af7316bf6a4749bfdbf8ff0ba4f5/commons/commons-bio/src/main/kotlin/ebi/ac/uk/io/sources/FileSourcesList.kt#L34) and rather than change user input I think we should fail in the same way we fail if path contains invalid character.

Juan-EBI (commented):
please check comment, we can meet and discuss about it

Juan-EBI (approved):
LGTM

PR #739, opened 2023-08-08
Pivotal ID # 185786866: Simple automated performance test

Juan-EBI:
https://www.pivotaltracker.com/story/show/185786866

### DIFF
--- a/submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/SubmissionPerformanceTest.kt
+++ b/submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/SubmissionPerformanceTest.kt
@@ -45,7 +45,7 @@ class SubmissionPerformanceTest(
     @EnabledIfEnvironmentVariable(named = fixedDelayEnv, matches = "\\d+")
     @EnabledIfSystemProperty(named = "enableFire", matches = "true")
     fun `With many files`() {
-        val files = 1000
+        val files = 100
         val delay = System.getenv(fixedDelayEnv).toLong()
 
         val subFiles = (1..files).map { tempFolder.createFile("$it.txt") }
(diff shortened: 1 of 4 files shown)


submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/SubmissionPerformanceTest.kt
```diff
@@ -0,0 +1,73 @@
...
+        val performanceSubmission = tsv {
+            line("Submission", "SPER-1")
+            line("Title", "Performance Submission")
+            line()
+
+            line("Study")
+            line()
+
+            line("Files")
+            subFiles.forEach { line(it.name) }
```
jhoanmanuelms, line 63: The test looks awesome, I think this is worth measuring. I think it could benefit from adding the files through a file list rather than placing them directly in the submission just to add to the test case the overhead of reading the file list at each stage which is our most common case for big submissions.

Also, I think we could increase the number of files to 1000. It'll add an extra minute but the test will be closer to our actual use case.

WDYT?

Juan-EBI: yes sound good, done.

Juan-EBI: actually just test it with 1000 and runner JVM dies (and I think updating runner to match production machine and get "real performance" is out of the scope)

jhoanmanuelms: yeah, maybe the execution is too long. I'd still like to see the files in a file list but I'll leave it to you.

jhoanmanuelms (approved):
LGTM

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/SubmissionPerformanceTest.kt
Juan-EBI: files are inded in a file list already

PR #746, opened 2023-09-14
Pivotal ID # 185788503: Reactive Mongo

Juan-EBI:
https://www.pivotaltracker.com/story/show/185788503
- reactive submission operations


submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/MongoDbConfig.kt
```diff
@@ -28,7 +28,7 @@ const val CHANGE_LOG_LOCK = "submitter_mongockLock"
 @Configuration
 @EnableMongoRepositories(
     basePackageClasses = [
-        SubmissionMongoRepository::class,
```
jhoanmanuelms, line 31: Why are we removing this one? 🤔

submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/db/data/SubmissionDocDataRepository.kt
```diff
@@ -45,6 +50,23 @@ class SubmissionDocDataRepository(
     private val submissionRepository: SubmissionMongoRepository,
     private val mongoTemplate: ReactiveMongoTemplate,
 ) : SubmissionMongoRepository by submissionRepository {
+
+    suspend fun saveSubmission(docSubmission: DocSubmission): DocSubmission {
+        return submissionRepository.save(docSubmission).awaitSingle()
+    }
+
+    suspend fun saveAllSubmissions(submissions: List<DocSubmission>) {
+        submissionRepository.saveAll(submissions).awaitFirst()
```
jhoanmanuelms, line 59: I'm wondering, if this is an operation to save all the elements, shouldn't we use `awaitLast()` instead to make sure all of them were processed? Or it doesn't make a difference?

submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/db/data/SubmissionDocDataRepository.kt
```diff
@@ -45,6 +50,23 @@ class SubmissionDocDataRepository(
...
+    }
+
+    suspend fun deleteAllSubmissions() {
+        submissionRepository.deleteAll().awaitSingleOrNull()
```
jhoanmanuelms, line 63: Same here

submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/db/reactive/repositories/Repositories.kt
```diff
@@ -38,3 +41,24 @@ interface SubmissionStatsRepository : ReactiveCrudRepository<DocSubmissionStats,
...
+
+    fun getByAccNoInAndVersionGreaterThan(accNo: List<String>, version: Int): Flux<DocSubmission>
+
+    suspend fun findFirstByAccNoAndVersionLessThanOrderByVersion(accNo: String, version: Int = 0): DocSubmission?
+
+    @Query(value = "{ 'accNo' : ?0, 'version' : { \$gt: 0} }", fields = "{ 'collections.accNo':1 }")
+    suspend fun findSubmissionCollections(accNo: String): SubmissionCollections?
+}
+
+suspend fun SubmissionMongoRepository.getByAccNo(accNo: String) =
```
jhoanmanuelms, line 63: Missing return type for public function

submission/persistence-mongo/src/test/kotlin/ac/uk/ebi/biostd/persistence/doc/db/data/SubmissionDocDataRepositoryTest.kt
```diff
@@ -47,17 +49,17 @@ internal class SubmissionDocDataRepositoryTest(
 ) {
 
     @BeforeEach
-    fun beforeEach() {
-        testInstance.deleteAll()
+    fun beforeEach() = runBlocking {
```
jhoanmanuelms, line 52: I understand `runTest` is equivalent to `runBlocking` but to be used in a test context since it handles the exceptions differently

```
If the created coroutine completes with an exception, then this exception will be thrown at the end of the test
```

So, I guess it's ok to use it here as well, just to avoid mixing those two. WDYT?

submission/persistence-sql/src/main/kotlin/ac/uk/ebi/biostd/persistence/integration/services/SqlPersistenceService.kt
```diff
@@ -23,13 +24,17 @@ internal open class SqlPersistenceService(
...
+            var next = sequence.counter.count + 1
+            while (submissionPersistenceQueryService.existByAccNo("${sequence.prefix}$next")) next++
 
-        sequence.counter.count = next
-        sequenceRepository.save(sequence).counter.count
+            sequence.counter.count = next
+            val saved = sequenceRepository.save(sequence)
+            return saved.counter.count
+        }
+        return lockExecutor.executeLocking(pattern) { runBlocking { return@runBlocking getNextSequence(pattern) } }
```
jhoanmanuelms, line 37: I think it looks clearer leaving it as it's right now and just adding the `runBlocking`. Also, once we migrate to mysql reactive, we would just need to make the function suspend and remove the `runBlocking` no?

jhoanmanuelms (approved):
LGTM. Please consider my comments

submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/MongoDbConfig.kt
Juan-EBI: this repository is not any more a normal one but reactive one so this one is loaded true reactive mongo config

submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/db/data/SubmissionDocDataRepository.kt
Juan-EBI: this one does not return a list

submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/db/data/SubmissionDocDataRepository.kt
Juan-EBI: make sense yes, great catch

submission/persistence-mongo/src/test/kotlin/ac/uk/ebi/biostd/persistence/doc/db/data/SubmissionDocDataRepositoryTest.kt
Juan-EBI: to be honest I am not sure but I think given this is not a test we should not use annotation.  runTest Documentaiton said `If the created coroutine completes with an exception, then this exception will be thrown at the end of the test.` I think in before each if something fail we want inmediante fail and also error should not propagetd to any test.

PR #748, opened 2023-09-19
Pivotal ID # 185649200: FIRE Duplicated Folder Name

jhoanmanuelms:
- Preserve inner directories path in order to deal with duplicates
- Add iTest


submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/SubmissionFileSourceTest.kt
```diff
@@ -219,8 +219,8 @@ class SubmissionFileSourceTest(
                 line("Type", "test")
                 line()
 
-                line("File", "folder/inner")
-                line("Type", "inner folder")
+                line("File", "folder/inner/directory")
+                line("Type", "inner directory")
```
Juan-EBI, line 223: I think this is really specialized scenario and can be easily miss in a refactor. IMO we should have specialized test "6-3-2 submission with directories with the same name on FIRE". In that way is super clear what is that that we are testing.

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/request/SubmissionRequestLoader.kt
```diff
@@ -73,7 +73,7 @@ class SubmissionRequestLoader(
 
     private fun asCompressedFile(accNo: String, version: Int, directory: NfsFile): NfsFile {
         fun compress(file: File): File {
-            val tempFolder = fireTempDirPath.resolve("$accNo/$version")
+            val tempFolder = fireTempDirPath.resolve("$accNo/$version/${directory.filePath.substringBeforeLast("/")}")
```
Juan-EBI, line 76: really subjective but  I think the problem is not in this line but rather `tempFolder.resolve("${file.name}.zip")` the problem with that code is that we are using fixed file path rather than do that we should use `Files.createTempFile` which will atuomatically garantee to us that same name is not used.

Juan-EBI (approved):
please consider changes before merge

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/request/SubmissionRequestLoader.kt
jhoanmanuelms: It actually solves the problem but, this method uses the VM's temp folder which is only 10 GB currently which I think won't be enough for the imaging people's requirements

Juan-EBI: what I mean is use the same method over the same folder `fireTempDirPath.createTempFile("$accNo/$version")`

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/SubmissionFileSourceTest.kt
jhoanmanuelms: Done

submission/submitter/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/request/SubmissionRequestLoader.kt
jhoanmanuelms: Got it, makes sense. Done!

PR #754, opened 2023-10-01
Pivotal ID # 185788503: Reactive Mongo

Juan-EBI:
- removed flux references

### DIFF
--- a/submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/db/reactive/repositories/Repositories.kt
+++ b/submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/db/reactive/repositories/Repositories.kt
@@ -10,9 +10,9 @@ import ac.uk.ebi.biostd.persistence.doc.model.DocSubmissionRequest
 import ac.uk.ebi.biostd.persistence.doc.model.DocSubmissionRequestFile
 import ac.uk.ebi.biostd.persistence.doc.model.DocSubmissionStats
 import ac.uk.ebi.biostd.persistence.doc.model.FileListDocFile
+import kotlinx.coroutines.flow.Flow
 import org.bson.types.ObjectId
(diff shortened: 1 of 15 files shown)


submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/db/reactive/repositories/Repositories.kt
```diff
@@ -98,29 +97,31 @@ interface SubmissionRequestFilesRepository : CoroutineCrudRepository<DocSubmissi
...
         version: Int,
         index: Int,
-    ): Flux<DocSubmissionRequestFile>
+    ): Flow<DocSubmissionRequestFile>
 
     suspend fun getByPathAndAccNoAndVersion(path: String, accNo: String, version: Int): DocSubmissionRequestFile
 }
 
 interface FileListDocFileRepository : CoroutineCrudRepository<FileListDocFile, ObjectId> {
-    @Meta(cursorBatchSize = 100, flags = [org.springframework.data.mongodb.core.query.Meta.CursorOption.NO_TIMEOUT])
```
jhoanmanuelms, line 107: I remember this was added in order to avoid the cursor timeout issue when processing large file lists. I don't think we should remove it unless it doesn't have any effect when using the reactive client

jhoanmanuelms (approved):
Really nice changes to avoid all these conversions 👍. LGTM, please consider my comment

submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/db/reactive/repositories/Repositories.kt
Juan-EBI: we are not using long query now but paged query so this is not necesary.  That was the intention of this changes to remove long running cursor that may fail.

PR #755, opened 2023-10-05
Pivotal ID # 185392780: User FTP files API support

Juan-EBI:
https://www.pivotaltracker.com/story/show/185392780
-  Changed logic to transform FTP check file in NFS as process will be executed in data mover


submission/file-sources/src/main/kotlin/uk/ac/ebi/io/sources/FtpSource.kt
```diff
@@ -7,13 +7,12 @@ import ebi.ac.uk.model.Attribute
...
 import uk.ac.ebi.io.builder.createFile
 import java.io.File
-import java.nio.file.Files
 import java.nio.file.Path
-import kotlin.io.path.outputStream
 
 class FtpSource(
     override val description: String,
-    private val basePath: Path,
+    private val ftpPath: Path,
```
jhoanmanuelms, line 14: I think the names `ftpUrl` and `nfsFtpPath` would be clearer in order to differentiate which resource we're referring to

submission/file-sources/src/main/kotlin/uk/ac/ebi/io/sources/FtpSource.kt
```diff
@@ -22,15 +21,10 @@ class FtpSource(
...
+        val ftpPath = ftpPath.resolve(filePath)
         val files = ftpClient.listFiles(ftpPath)
         if (files.isEmpty()) return null
-
-        val fileName = filePath.substringAfterLast("/")
-        val target = Files.createTempFile("ftp-file", fileName)
-
-        target.outputStream().use { ftpClient.downloadFile(ftpPath, it) }
-        return target.toFile()
+        return nfsPath.resolve(filePath).toFile()
```
jhoanmanuelms, line 27: If I understood correctly, this is what we're doing:
1. We use the ftp client to list the files
2. If the file is not found we return null
3. If it exists, we return whatever is on `nfsPath.resolve(filePath)`

If we're returning the file from the nfsPath and the only reason to need the ftp client is to list the files, wouldn't this be equivalent to:

```
    val file = nfsPath.resolve(filePath)
    return if (file.exists()) file.toFile() else null
```

I just want to understand the need for the ftp client here

jhoanmanuelms (commented):
Looking great, I just have one question

jhoanmanuelms (approved):
LGTM

PR #756, opened 2023-10-06
Pivotal ID # 186180872:  Submission Migrator Task

jhoanmanuelms:

- Create a task to migrate a specific set of submissions from NFS to FIRE
- Start a separate instance of the backend in order to handle the migration requests


scheduler/tasks/submission-migrator-task/src/main/kotlin/uk/ac/ebi/scheduler/migrator/service/SubmissionMigratorService.kt
```diff
@@ -0,0 +1,49 @@
...
+                .buffer(properties.concurrency)
+                .collect { it.await() }
+        }
+    }
+
+    private fun migrateSafely(migrationData: MigrationData) {
+        fun migrate() {
+            logger.info { "Started migrating submission ${migrationData.accNo} to FIRE" }
+            bioWebClient.transferSubmission(migrationData.accNo, FIRE)
+            await()
```
Juan-EBI, line 39: I think we should use suspend instead of blocking the thread. I mean delay instead of sleep we got a function in test module but we can transfer to here.

PR #759, opened 2023-10-12
Pivotal ID # 185900074: Accept Requests With Large Directories Times Out

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/185900074

Avoid directories calculation at the moment of generating pagetab


commons/commons-model-extended-mapping/src/main/kotlin/ebi/ac/uk/extended/mapping/to/ToFile.kt
```diff
@@ -8,9 +8,11 @@ import ebi.ac.uk.model.BioFile
 
 internal const val TO_FILE_EXTENSIONS = "ebi.ac.uk.extended.mapping.to.ToFileKt"
 
-fun ExtFile.toFile(): BioFile {
+fun ExtFile.toFile(calculateDirectories: Boolean = true): BioFile {
```
Juan-EBI, line 11: is confusing for me when the files size need to be calculated and when not.  Like all operations support both calculating them and not calculating. How is posible that calculating them is some parts is good but in others not?  I think code does not reveal when is necesary to calculated and when is not

Juan-EBI: agree to always use false

PR #762, opened 2023-10-25
Pivotal ID # 186322997:  S-BIAD877 could not be loaded

Juan-EBI:
https://www.pivotaltracker.com/story/show/186322997
- refactor to use async flow rather than blocking sequence

### DIFF
--- a/commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/integration/SerializationService.kt
+++ b/commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/integration/SerializationService.kt
@@ -5,6 +5,7 @@ import ebi.ac.uk.model.BioFile
 import ebi.ac.uk.model.FilesTable
 import ebi.ac.uk.model.Submission
 import kotlinx.coroutines.flow.Flow
+import org.jetbrains.annotations.Blocking
 import java.io.File
 import java.io.InputStream
 import java.io.OutputStream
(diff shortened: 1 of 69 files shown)


commons/commons-serialization/src/main/kotlin/ac/uk/ebi/biostd/integration/SerializationService.kt
```diff
@@ -5,26 +5,31 @@ import ebi.ac.uk.model.BioFile
...
-
     fun deserializeSubmission(content: String, format: SubFormat): Submission
 
-    suspend fun deserializeSubmission(content: String, format: SubFormat, source: FileSourcesList): Submission
-
+    @Blocking
     fun deserializeSubmission(file: File): Submission
 
-    suspend fun deserializeSubmission(file: File, source: FileSourcesList): Submission
+    @Blocking
```
Juan-EBI, line 27: @jhoanmanuelms lets start using this one so we know when a code is blocking

commons/commons-serialization/src/testFixtures/kotlin/ac/uk/ebi/biostd/TestFunctions.kt
```diff
@@ -3,13 +3,15 @@ package ac.uk.ebi.biostd
...
 import java.nio.file.Files
 import kotlin.io.path.outputStream
 
-fun createFileList(vararg files: BioFile, format: SubFormat = SubFormat.JSON): File {
+fun createFileList(vararg files: BioFile, format: SubFormat = SubFormat.JSON): File = runBlocking {
     val file = Files.createTempFile("file-list", "${files.size}-files")
     val serializer = SerializationConfig.serializationService()
-    file.outputStream().use { serializer.serializeFileList(files.asSequence(), format, it) }
-    return file.toFile()
+    file.outputStream().use { serializer.serializeFileList(files.asFlow(), format, it) }
```
jhoanmanuelms, line 15: Shouldn't this also be executed in the IO context?

jhoanmanuelms (approved):
LGTM. Is looking good getting rid of all the sequences to favor flows 👍

commons/commons-serialization/src/testFixtures/kotlin/ac/uk/ebi/biostd/TestFunctions.kt
Juan-EBI: yes but it will require the function to be suspended and we use it as a top level. So because it is a test I think runBlocking will do.

PR #763, opened 2023-10-26
Pivotal ID # 186293173: Persist Files With Duplicated MD5 And Different Name

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/186293173

- Remove MD5 search
- Check for the appropriate web client error when using the find by path endpoint
- Use the actual data returned by fire when storing it to the database
- Use a mutex to avoid race conditions
- Add an integration test


submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/fire/FireFilesService.kt
```diff
@@ -34,32 +34,33 @@ class FireFilesService(
...
 
     private suspend fun getOrCreate(file: NfsFile, expectedPath: String): FireFile {
-        val matches = client.findByMd5(file.md5)
-        val apiFile = matches.find { it.path == expectedPath }
-            ?: matches.find { it.path == null }
-            ?: client.save(file.file, file.md5, file.size)
+        val previousFile = client.findByPath(expectedPath)
+        val apiFile = when {
+            previousFile != null && previousFile.objectMd5 == file.md5 -> previousFile
+            else -> client.save(file.file, file.md5, file.size)
```
Juan-EBI, line 50: this does not look correct for me if the file md5 is not the same we wll not succed to save file in the same path so there is not point to saved it. So IMO if file is there at this point is because is because is the same otherwize it should have clean in a previous stage.

submission/submission-core/src/main/kotlin/ac/uk/ebi/biostd/submission/submitter/request/SubmissionRequestProcessor.kt
```diff
@@ -43,12 +45,16 @@ class SubmissionRequestProcessor(
             logger.info { "$accNo ${sub.owner} Finished persisting file ${rqtFile.index}, path='${rqtFile.path}'" }
         }
 
+        val mutexMap = ConcurrentHashMap<String, Mutex>()
```
Juan-EBI, line 48: I do not think this is necesary any more. I mean we will only re use file if it is in the same path so there is not race condition posibility.

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/SubmissionFileSourceTest.kt
```diff
@@ -727,4 +727,44 @@ class SubmissionFileSourceTest(
             assertThat(firstVersionFireId).isEqualTo(secondVersionFireId)
         }
     }
+
+    @Test
+    @EnabledIfSystemProperty(named = "enableFire", matches = "true")
+    fun `6-8 submission with files with the same md5 and different path`() = runTest {
```
Juan-EBI, line 733: nice test addition.

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/SubmissionFileSourceTest.kt
```diff
@@ -727,4 +727,44 @@ class SubmissionFileSourceTest(
...
+        files.forEach { it.delete() }
+        fileList.delete()
+
+        val submission = submissionRepository.getExtByAccNo("S-FSTST8")
+        val duplicates = filesRepository.getReferencedFiles(submission, "DuplicatedFiles")
+            .toList()
+            .groupBy { (it as FireFile).fireId }
+            .filter { it.value.size > 1 }
+
+        assertThat(duplicates.isEmpty()).isTrue()
```
Juan-EBI, line 768: maybe 
```
        assertThat(duplicates).allSatisfy { id, fireFiles ->
            assertThat(fireFiles)
                .hasSize(1)
                .withFailMessage("duplicated fire file id='{}'", id)
        }
```

Juan-EBI (commented):
please check comments

Juan-EBI (approved):
LGTM

PR #767, opened 2023-11-07
Pivotal ID # 186180872: Submission Migrator Task

Juan-EBI:
- added yaml based itest.
- refactor to use rabbit mq one.


buildSrc/src/main/kotlin/Dependencies.kt
```diff
@@ -133,6 +133,7 @@ object TestDependencies {
     const val JaxbApi = "javax.xml.bind:jaxb-api:$JaxbApiVersion"
     const val XmlUnitCore = "org.xmlunit:xmlunit-core:$XmlUnitVersion"
     const val XmlUnitMatchers = "org.xmlunit:xmlunit-matchers:$XmlUnitVersion"
+    const val SnakeYaml = "org.yaml:snakeyaml"
```
jhoanmanuelms, line 136: Should we set the specific version?

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/itest/PropertyHolder.kt
```diff
@@ -0,0 +1,44 @@
+package ac.uk.ebi.biostd.itest.itest
+
+import org.yaml.snakeyaml.DumperOptions
+import org.yaml.snakeyaml.Yaml
+import java.io.File
+
+/**
+ * Holds application properties and write them into its YAML representation.
+ */
+class PropertyHolder {
```
jhoanmanuelms, line 10: I'd suggest `ConfigurationPropertiesHolder`, `ConfigurationPropertiesWriter` or `ConfigurationPropertiesHelper` which is a little bit more specific to what it actually does

jhoanmanuelms (approved):
LGTM. Nice changes! Please check my comments before merging

buildSrc/src/main/kotlin/Dependencies.kt
Juan-EBI: no, it resolved by spring boot so in that way we use the same version as them and there is not risk of imcompability. Actually I was having some issues when using the last version.

PR #768, opened 2023-11-07
Pivotal ID # 184977702: Assign DOIs Automatically

jhoanmanuelms:
- Make DOI properties case insensitive
- Include section title as valid submission title
- Add "Organisation" as valid tag
- Add these cases to the unit tests
- Update application.yml to match the configuration property


submission/submission-core/src/main/kotlin/ac/uk/ebi/biostd/submission/service/DoiService.kt
```diff
@@ -55,7 +55,7 @@ class DoiService(
     private fun registerDoi(accNo: String, rqt: SubmitRequest): String {
         val sub = rqt.submission
         val timestamp = Instant.now().epochSecond.toString()
-        val title = requireNotNull(sub.title) { throw MissingTitleException() }
+        val title = rqt.submission.find(TITLE) ?: rqt.submission.section.find(TITLE) ?: throw MissingTitleException()
```
Juan-EBI, line 58: in the first call we search in case sensite and in the second it is not. Are these the expectations?

submission/submission-core/src/main/kotlin/ac/uk/ebi/biostd/submission/service/DoiService.kt
```diff
@@ -121,16 +122,20 @@ class DoiService(
...
+    private fun Section.findAttr(attribute: String): String? =
+        attributes.find { it.name.lowercase() == attribute.lowercase() }?.value
+
     companion object {
         internal const val AFFILIATION_ATTR = "Affiliation"
         internal const val NAME_ATTR = "Name"
         internal const val ORCID_ATTR = "ORCID"
 
-        internal const val ORG_TYPE = "Organization"
+        internal const val ORG_TYPE_1 = "Organization"
```
Juan-EBI, line 137: rather that defined constant with capaital and then calling lowercase I think is simpler and efficient just have constant in lower case

submission/submission-core/src/main/kotlin/ac/uk/ebi/biostd/submission/service/DoiService.kt
```diff
@@ -107,10 +107,11 @@ class DoiService(
     }
 
     private fun getOrganizations(sub: Submission): Map<String, String> {
-        val organizations = sub.allSections().filter { it.type == ORG_TYPE }
+        val organizations = sub.allSections()
+            .filter { it.type.lowercase() == ORG_TYPE_1.lowercase() || it.type.lowercase() == ORG_TYPE_2.lowercase() }
```
Juan-EBI, line 111: I would rather defined a Set contant and do `validOrds.contains(it.type)`

Juan-EBI (approved):
LGTM please consider commnets before merge

submission/submission-core/src/main/kotlin/ac/uk/ebi/biostd/submission/service/DoiService.kt
jhoanmanuelms: Nice catch

PR #772, opened 2023-11-10
Pivotal ID # 186180872: Submission Migrator Task

Juan-EBI:
- added migrator task and task mode


submission/submission-task/src/main/kotlin/ac/uk/ebi/biostd/submission/task/SubmissionApp.kt
```diff
@@ -0,0 +1,64 @@
+package ac.uk.ebi.biostd.submission.task
```
jhoanmanuelms, line 1: Since this is a new project, I think we should use the right package domain naming since the beginning: `uk.ac.ebi`

submission/submission-webapp/gradle.properties
```diff
@@ -1,2 +1,3 @@
 coverage=0.7
 enableFire=true
+enableTask=false
```
jhoanmanuelms, line 3: I think we should name it `enableTaskMode` just to keep it consistent with the actual application's yml

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/SubmissionAsyncTest.kt
```diff
@@ -59,7 +59,8 @@ class SubmissionAsyncTest(
     }
 
     @Test
-    fun `19-1 simple submit async`() = runTest {
+    fun `19-1 simple submit async`() = runTest() {
```
jhoanmanuelms, line 62: Why do we need these parenthesis here?

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/service/DynamicExtSubmissionSubmitter.kt
```diff
@@ -0,0 +1,46 @@
...
+import ac.uk.ebi.biostd.submission.domain.extended.LocalExtSubmissionSubmitter
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+import mu.KotlinLogging
+import java.io.File
+import java.util.UUID
+
+private val logger = KotlinLogging.logger {}
+
+class DynamicExtSubmissionSubmitter(
```
jhoanmanuelms, line 14: Why isn't this implementation located in the same place as the Local one? I think both are concerns of the submission core, not the web layer

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/service/DynamicExtSubmissionSubmitter.kt
```diff
@@ -0,0 +1,46 @@
...
+    private val extSubmissionSubmitter: LocalExtSubmissionSubmitter,
```
jhoanmanuelms, line 15: I'm not very on board with the idea of the dynamic version containing the local one. I think we should have a handler that decides which one to use based on the configuration

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/service/DynamicExtSubmissionSubmitter.kt
```diff
@@ -0,0 +1,46 @@
...
+    private val hostProperties: TaskHostProperties,
+) : ExtSubmissionSubmitter by extSubmissionSubmitter {
+
+    override suspend fun processRequest(accNo: String, version: Int) {
+        when (hostProperties.enableTaskMode) {
+            true -> copyRequestFilesRemotly(accNo, version)
```
jhoanmanuelms, line 21: I think the dynamic submitter should be able to execute any stage, not only copying the files. Or at least that's what I understood from the previous code.

jhoanmanuelms (commented):
The changes look great. I left a couple of comments for you to consider

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/service/DynamicExtSubmissionSubmitter.kt
Juan-EBI: given the fact that executing other steps has other implications I do not see the point make implementation more complex by increaasing the scope of the feature  to "be able to execute any stage in the data mover". Right now is only the persitence

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/service/DynamicExtSubmissionSubmitter.kt
Juan-EBI: the problem is that we do not support yet any phase execution in submitter, only persistence. I guess we may in the future but I do not see the reason to complicate the implementation more  at this point.

jhoanmanuelms (approved):
LGTM

PR #775, opened 2023-11-21

jhoanmanuelms:

Fix regex expression to escape the hyphen character and ensure relative paths are avoided

### DIFF
--- a/commons/commons-bio/src/main/kotlin/ebi/ac/uk/io/sources/FileSourcesList.kt
+++ b/commons/commons-bio/src/main/kotlin/ebi/ac/uk/io/sources/FileSourcesList.kt
@@ -23,7 +23,16 @@ import java.io.File
  *     - Open parenthesis ( ( )
  *     - Close parenthesis ( ) )
  */
-private val validPathPattern = "^(?!\\./)(?!.*/\$)(?!.*\\.\\./)[0-9A-Za-z!\\-_*'(). /]+(?!/\$)\$".toRegex()
+
+private val validPathPattern = buildString {
+    append("^")                        // Start of the line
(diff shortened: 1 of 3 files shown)


commons/commons-bio/src/main/kotlin/ebi/ac/uk/io/sources/FileSourcesList.kt
```diff
@@ -23,7 +23,7 @@ import java.io.File
  *     - Open parenthesis ( ( )
  *     - Close parenthesis ( ) )
  */
-private val validPathPattern = "^(?!.*\\./)[0-9A-Za-z!-_*'(). ]+(?<!/)\$".toRegex()
+private val validPathPattern = "^(?!\\./)(?!.*/\$)(?!.*\\.\\./)[0-9A-Za-z!\\-_*'(). /]+(?!/\$)\$".toRegex()
```
Juan-EBI, line 26: what about of:

```
private val validPathPattern = """
        ^                       # Start of the string
        (?!.*(?:\./|\.\./))     # Negative lookahead to disallow './' or '../'
        (?!.*\/$)               # Negative lookahead to disallow trailing '/'
        [0-9A-Za-z!\-_*'(). /]+ # Match one or more of the allowed characters
        $                       # End of the string
    """.trimIndent().toRegex(RegexOption.COMMENTS)
```

with some help of chat GPT

Juan-EBI (approved):

commons/commons-bio/src/main/kotlin/ebi/ac/uk/io/sources/FileSourcesList.kt
jhoanmanuelms: really nice recommendation man! 👏 🤯

PR #776, opened 2023-11-21
Pivotal ID # 186529137: Extend cursor timeout

Juan-EBI:
https://www.pivotaltracker.com/story/show/186529137


jhoanmanuelms (approved):
LGTM

PR #781, opened 2023-11-29
Pivotal ID #186578721: Allow to execute test jobs in cluster API

Juan-EBI:
https://www.pivotaltracker.com/story/show/186578721
- added new api


submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/cluster/web/ClusterOperationsResource.kt
```diff
@@ -1,22 +1,45 @@
...
+            is Try.Success -> response.value
+        }
+    }
+
+    @GetMapping("/jobs/{jobId}/status")
+    suspend fun jobStatus(@PathVariable jobId: String): String {
+        return clusterOperations.jobStatus(jobId)
+    }
+
+    data class JobSpecDto(val command: String, val queue: String, val megaBytes: Int) {
```
jhoanmanuelms, line 40: I think the last parameter should be called `memory`, `ramSize` or even just `ram` like in the Job class because `megaBytes` is not very clear at communicating what the value is actually used for

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/cluster/web/ClusterOperationsResource.kt
```diff
@@ -1,22 +1,45 @@
...
 ) {
     @PostMapping("/health")
     suspend fun clusterHealthCheck(): Job {
         val spec = JobSpec(command = "echo 'hello world'")
-        return clusterOperations.awaitJob(spec)
+        return clusterOperations.runJob(spec)
+    }
+
+    @PostMapping("/submit")
+    suspend fun submitJob(@RequestBody job: JobSpecDto): Job {
```
jhoanmanuelms, line 28: I think these endpoints will be super useful

jhoanmanuelms (commented):
It's looking good, I only have one concern which is related to the private key in the repo

submission/submission-webapp/src/itest/resources/key
```diff
@@ -0,0 +1,39 @@
+-----BEGIN RSA PRIVATE KEY-----
```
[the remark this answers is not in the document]
Juan-EBI: is a dummy key required for itest to run

jhoanmanuelms (approved):
LGTM

PR #788, opened 2023-12-08
Pivotal ID # 186080414: Execute Submission Persistence On DMs

jhoanmanuelms:
- Add a new stage for page tab generation
- Generate the page tab files after the request has been loaded to avoid md5/size recalculation


submission/submission-core/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/request/SubmissionRequestLoader.kt
```diff
@@ -98,4 +100,16 @@ class SubmissionRequestLoader(
             type = ExtFileType.DIR
         )
     }
+
+    private suspend fun assembleLoadedSubmission(sub: ExtSubmission): ExtSubmission {
```
Juan-EBI, line 104: this is super slow as for each file we are hitting the database. Why we are doing this?

jhoanmanuelms:
Closed in favor of #789

PR #790, opened 2023-12-12
Pivotal ID # 186655077: FTP calculate file size

Juan-EBI:
- added file size calculation


jhoanmanuelms (approved):
LGTM

PR #793, opened 2023-12-15
Pivotal ID # 186657295: Retrigger Submission Task

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/186657295

- Handle re-triggering for submissions executed remotely
- Remove unused code

### DIFF
--- a/submission/submission-core/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/submitter/ExtSubmissionSubmitter.kt
+++ b/submission/submission-core/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/submitter/ExtSubmissionSubmitter.kt
@@ -5,11 +5,11 @@ import ebi.ac.uk.extended.model.ExtSubmission
 
 interface ExtSubmissionSubmitter {
     suspend fun createRequest(rqt: ExtSubmitRequest): Pair<String, Int>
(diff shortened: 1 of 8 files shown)


submission/submission-core/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/submitter/ExtSubmissionSubmitter.kt
```diff
@@ -5,11 +5,11 @@ import ebi.ac.uk.extended.model.ExtSubmission
 
 interface ExtSubmissionSubmitter {
     suspend fun createRequest(rqt: ExtSubmitRequest): Pair<String, Int>
-    suspend fun indexRequest(accNo: String, version: Int): Unit
-    suspend fun loadRequest(accNo: String, version: Int): Unit
-    suspend fun cleanRequest(accNo: String, version: Int): Unit
-    suspend fun processRequest(accNo: String, version: Int): Unit
-    suspend fun checkReleased(accNo: String, version: Int): Unit
+    suspend fun indexRequest(accNo: String, version: Int)
```
Juan-EBI, line 8: i start adding explicit return type only to public api/functions as for example if you call a method in the last line of function body the return value becomes something and may break unit test for example.

Juan-EBI (approved):
LGTM

PR #794, opened 2023-12-19
Pivotal ID # 186629008: Slack Message For Failed Notifications

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/186629008

- Add message converter to the rabbit template
- Configure custom exception handler to catch failing messages


Juan-EBI (approved):
LGTM

PR #795, opened 2023-12-21
Pivotal ID # 186232047: Manager Should Not Be Able To Delete

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/186232047

- Separate admin from regular permissions
- Allow deletion only for specific permission or author of private submission
- Improve integration testing
- Add unit testing


Juan-EBI (approved):
LGTM

PR #796, opened 2023-12-21
Pivotal ID # 186657310: Task Completion Publish Message

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/186657310

- Delegate sending the completion message to the task itself
- Convert the call to the submission task into an async triggering
- Sync submit waits for the async execution to finish since it's intended for small submissions


submission/submission-core/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/submission/SubmissionService.kt
```diff
@@ -30,7 +32,12 @@ class SubmissionService(
 ) {
     suspend fun submit(rqt: SubmitRequest): ExtSubmission {
         logger.info { "${rqt.accNo} ${rqt.owner} Received sync submit request with draft key '${rqt.draftKey}'" }
-        return submissionSubmitter.submit(rqt)
+
+        val (accNo, version) = submissionSubmitter.createRequest(rqt)
+        eventsPublisherService.requestCreated(accNo, version)
+        waitUntil(duration = ofMinutes(SYNC_SUBMIT_TIMEOUT)) { queryService.existByAccNoAndVersion(accNo, version) }
+
+        return queryService.getExtByAccNo(accNo)
```
Juan-EBI, line 40: I will suggest we create the function 
```
suspend fun <T> waitUntil(
    duration: Duration,
    interval: Duration = Duration.ofMillis(300),
    conditionEvaluator: suspend () -> Boolean,
    processFunction: suspend () -> T,
): T {
    waitUntil(duration, interval, conditionEvaluator)
    return processFunction()
}
```

and then here
```
        return waitUntil(
            ofMinutes(SYNC_SUBMIT_TIMEOUT),
            conditionEvaluator = { queryService.existByAccNoAndVersion(accNo, version) },
            processFunction = { queryService.getExtByAccNoAndVersion(accNo, version) })
```

wait and having a return type looks like a coomon need .

Juan-EBI (approved):
LGTM please consider minor comment

PR #797, opened 2024-01-11
Pivotal ID # 186609885: Access cluster logs admin API

Juan-EBI:
https://www.pivotaltracker.com/story/show/186609885
- added logs api

### DIFF
--- a/client/cluster-client/src/main/kotlin/uk/ac/ebi/biostd/client/cluster/api/LocalClusterClient.kt
+++ b/client/cluster-client/src/main/kotlin/uk/ac/ebi/biostd/client/cluster/api/LocalClusterClient.kt
@@ -11,11 +11,14 @@ import kotlin.io.path.createTempFile
 
 class LocalClusterClient : ClusterClient {
     private val activeProcess = ConcurrentHashMap<Long, Process>()
+    private val jobLogs = ConcurrentHashMap<Long, File>()
 
     override suspend fun triggerJobAsync(jobSpec: JobSpec): Try<Job> {
         return withContext(Dispatchers.IO) {
             val logFile = createTempFile().toFile()
             val processId = executeProcess(jobSpec.command, logFile)
+
+            jobLogs[processId] = logFile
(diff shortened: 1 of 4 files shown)


client/cluster-client/src/main/kotlin/uk/ac/ebi/biostd/client/cluster/api/LocalClusterClient.kt
```diff
@@ -34,6 +37,10 @@ class LocalClusterClient : ClusterClient {
         return if (process.isAlive) "RUNNING" else "DONE"
     }
 
+    override suspend fun jobLogs(jobId: String): String {
+        return jobLogs.getValue(jobId.toLong()).readText()
```
jhoanmanuelms, line 41: I'm just curious, why do you use `getValue` instead of `jobLogs[jobId.toLong()] `?

client/cluster-client/src/main/kotlin/uk/ac/ebi/biostd/client/cluster/api/RemoteClusterClient.kt
```diff
@@ -49,6 +49,13 @@ class RemoteClusterClient(
         }
     }
 
+    override suspend fun jobLogs(jobId: String): String {
+        return runInSession {
+            val (_, response) = executeCommand("cat $logsPath/${jobId}_OUT")
```
jhoanmanuelms, line 54: And there we were, thinking about datamovers and separate jobs 🤣 

Well done man 👌

jhoanmanuelms (approved):
LGTM

client/cluster-client/src/main/kotlin/uk/ac/ebi/biostd/client/cluster/api/LocalClusterClient.kt
Juan-EBI: nullable vs not nullable

PR #802, opened 2024-01-30
Pivotal ID # 186941052: Check Ftp Connections Before Borrow

Juan-EBI:
https://www.pivotaltracker.com/story/show/186941052
- enabled test on borrow


jhoanmanuelms (approved):
LGTM

PR #805, opened 2024-02-14
Pivotal ID # 187038254: File List Name Character Validation

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/187038254

Validate the file list name to avoid invalid characters


Juan-EBI (approved):
LGTM

PR #808, opened 2024-02-22
Pivotal ID # 186992651: NFS Private Submissions Persistence

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/186992651

- Move submission persistence related properties to the persistence
- Move fire temp dir property to the fire properties section
- Update the property names according to the new feature


Juan-EBI (approved):
LGTM

PR #811, opened 2024-03-01
Pivotal ID # 186944806: Deprecate XML Pagetab

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/186944806

- Remove XML submission format
- Remove XML submit endpoints and logic
- Remove XML related documentation


Juan-EBI (approved):
do. we want to remove xml endpoint too?

jhoanmanuelms:
Yes, I think it's for the best to remove the related XML support altogether instead of leaving it just for some things. This was also validated with the team and no one is actually using it.

PR #813, opened 2024-03-06
Pivotal ID # 186553341: Avoid Retriggering Submission Publishing

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/186553341

- Keep track of the published flag per file in order to be able to continue the release process
- Remove the release parameter from the mapping functions
- Apply database patch to set the released flag for the currently existing files


submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/migrations/DatabaseChangeLog.kt
```diff
@@ -185,3 +200,40 @@ suspend fun ReactiveMongoOperations.ensureStatsIndexes() {
...
+        match(where(SUB_VERSION).gt(0).andOperator(where(STORAGE_MODE).`is`(StorageMode.FIRE.value))),
+        lookup(FILE_LIST_DOC_FILE_COLLECTION_NAME, SUB_ACC_NO, FILE_LIST_DOC_FILE_SUBMISSION_ACC_NO, "result"),
+        match(where("result").not().size(0)),
+        project(SUB_ACC_NO, SUB_VERSION, SUB_RELEASED)
+    )
+
+    aggregate(aggregation, Submission::class.java)
+        .asFlow()
+        .map { setReleasedFlag(it) }
+        .collect()
```
Juan-EBI, line 232: are we not planing to set flag for internal files but only to file list files?

Juan-EBI:
Looking good, said that I think we are missing some parts

1. NfsFtpService need to set release flag too after saving the file.
2. SubmissionRequestReleaser need to somehow persist changes of released file otherwize we will not keep track of it. 


Also it is not clear why this changes are need if release process consider files `getSubmissionRequestFiles(sub.accNo, sub.version, startingAt)` why the process will start again?

submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/migrations/DatabaseChangeLog.kt
jhoanmanuelms: Good catch

jhoanmanuelms:
The `startingAt` works at the request level but when a submission is picked up by the releaser, there's no request but it just goes through the files and releases them all trying to [filter by the published flag](https://github.com/EBIBioStudies/biostudies-backend-services/blob/master/submission/submission-core/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/request/SubmissionRequestReleaser.kt#L111) to avoid re-publishing

PR #829, opened 2024-04-23
Pivotal ID #187440833: Monitor Handlers App Notifications

jhoanmanuelms:
- Enable actuator for the scheduler and handlers apps
- Add complete path for the logs folder to the update script


Juan-EBI (approved):
LGTM

PR #834, opened 2024-05-15
Pivotal ID # 187277504: Submission Level Permission

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/187277504

- Fix logic to provide permissions over submissions
- Create an operation to provide submission level permissions
- Update resources and operation names
- Remove unused code


client/bio-commandline/src/main/kotlin/uk/ac/ebi/biostd/client/cli/services/SecurityService.kt
```diff
@@ -3,14 +3,23 @@ package uk.ac.ebi.biostd.client.cli.services
 import uk.ac.ebi.biostd.client.cli.dto.PermissionRequest
 
 internal class SecurityService {
-    fun grantPermission(request: PermissionRequest) = performRequest { permission(request) }
+    fun grantCollectionPermission(request: PermissionRequest) =
```
Juan-EBI, line 6: wondering if this really is a collection, basically we give permission to access No, like a submission I am wondering if command, documentaiton and this method name should reflect that.

client/bio-webclient/src/main/kotlin/ac/uk/ebi/biostd/client/integration/web/SubmitClient.kt
```diff
@@ -226,9 +226,15 @@ interface ExtSubmissionOperations {
...
 interface PermissionOperations {
-    fun givePermissionToUser(
+    fun grantCollectionPermission(
         user: String,
-        accessTagName: String,
+        accNo: String,
+        accessType: String,
+    )
+
+    fun grantSubmissionPermission(
```
Juan-EBI, line 235: I think we should not mantain two operations but a single one. grantPermision depending if accNo is a collection or a submission I do not think should be relevant. `userPermissionsService.hasPermission` does not care already

Juan-EBI (commented):
Looking good please check comment, it may be subjetive so we can discuss it

jhoanmanuelms:
Closed in favor of #844

PR #837, opened 2024-05-16
Pivotal ID # 187456681: Submission request file status

Juan-EBI:
https://www.pivotaltracker.com/story/show/187456681
- added cleaning stage


submission/persistence-common-api/src/main/kotlin/ac/uk/ebi/biostd/persistence/common/model/SubmissionRequest.kt
```diff
@@ -18,16 +18,22 @@ data class SubmissionRequest(
     val notifyTo: String,
     val status: RequestStatus,
     val totalFiles: Int,
+    val conflictedFiles: Int,
```
jhoanmanuelms, line 21: I'd suggest conflicting instead of conflicted

submission/persistence-common-api/src/main/kotlin/ac/uk/ebi/biostd/persistence/common/model/SubmissionRequest.kt
```diff
@@ -18,16 +18,22 @@ data class SubmissionRequest(
...
+    val deprecatedFiles: Int,
     val currentIndex: Int,
     val modificationTime: OffsetDateTime,
+    val previousVersion: Int?,
```
jhoanmanuelms, line 25: I just realized we have the `version` field in the `DocSubmissionRequest` class but we don't map it here, I'm not sure why but I think we should add it

Also, I noticed the class `RequestStatusChanges` (just above this one) appears as unused. Maybe we should delete it or is it intended to be used in a later PR?

submission/persistence-common-api/src/main/kotlin/ac/uk/ebi/biostd/persistence/common/model/SubmissionRequest.kt
```diff
@@ -85,7 +115,8 @@ val RequestStatus.action: String
         return when (this) {
             RequestStatus.REQUESTED -> "Indexing"
             RequestStatus.INDEXED -> "Loading"
-            RequestStatus.LOADED -> "Cleaning"
+            RequestStatus.LOADED -> "Indexing To Clean Files"
```
jhoanmanuelms, line 118: I'd suggest: Indexing Files To Clean

submission/submission-core/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/request/SubmissionRequestCleaner.kt
```diff
@@ -1,104 +1,91 @@
...
+     * persisted at this point.
+     */
+    suspend fun finalizeRequest(
+        accNo: String,
+        version: Int,
+        processId: String,
+    ) {
+        requestService.onRequest(accNo, version, PERSISTED, processId, {
+            val previousVersion = it.previousVersion
+            if (previousVersion != null) cleanFiles(accNo, version, previousVersion = -previousVersion, DEPRECATED)
```
jhoanmanuelms, line 59: I thought we were storing deprecated files in the current submission request

submission/submission-core/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/submitter/RemoteExtSubmissionSubmitter.kt
```diff
@@ -79,6 +79,13 @@ class RemoteExtSubmissionSubmitter(
         TODO("Remote execution not required")
     }
 
+    override suspend fun indexToCleanRequest(
+        accNo: String,
+        version: Int,
+    ) {
+        TODO("Remote execution not required")
```
jhoanmanuelms, line 86: Since this might be dealing with already existing submissions (i.e. the files might be in the new NFS), shouldn't this also be executed remotely?

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/stats/SubmissionStatsTest.kt
```diff
@@ -201,9 +203,24 @@ class SubmissionStatsTest(
 
             assertThat(webClient.submitSingle(version2, TSV, NFS)).isSuccessful()
             waitUntil(TEN_SECONDS) { statsDataService.findByAccNo("S-STTS2").first().value != 574L }
+
+            val sub = submissionRepository.getCoreInfoByAccNoAndVersion("S-STTS2", 2)
```
jhoanmanuelms, line 207: Nice improvement 👍

jhoanmanuelms (commented):
Changes looking great. I just have a couple of comments for us to discuss

submission/persistence-common-api/src/main/kotlin/ac/uk/ebi/biostd/persistence/common/model/SubmissionRequest.kt
Juan-EBI: we do not include the version here because is kind of in the submission object. I mean version =  submission.version. Still if you think is there any use for it a root level I can added it but is redudant.

submission/submission-core/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/request/SubmissionRequestCleaner.kt
Juan-EBI: not following this comment. As submission is deprecated and we need to load submission data (ralpath and secret) that is the reason we search for submission with negative version

submission/submission-core/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/submitter/RemoteExtSubmissionSubmitter.kt
Juan-EBI: make sense yes

PR #846, opened 2024-06-05
Pivotal ID # 187484499: Make Files Private On Resubmission

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/187484499

- Update the un-publish fire operation to return the un-published file
- Make files private when a submission becomes private


submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/nfs/NfsFtpService.kt
```diff
@@ -53,5 +55,41 @@ class NfsFtpService(
         return nfsFile
     }
 
-    private fun getPublicFolder(relPath: String): Path = FileUtils.getOrCreateFolder(folderResolver.getPublicSubFolder(relPath), RWXR_XR_X)
+    override suspend fun suppressSubmissionFile(
```
Juan-EBI, line 58: I am not the best with names or english but I simple do not relate `supress` remove published files. unrelease, unpublish sounds way better and natural for me. Maybe is worth it ask team about this (as is basically the domain language)

submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/nfs/NfsFtpService.kt
```diff
@@ -53,5 +55,41 @@ class NfsFtpService(
...
+        file: ExtFile,
+    ): ExtFile =
+        withContext(Dispatchers.IO) {
+            val nfsFile = file as NfsFile
+            val publicSubFolder = getPublicFolder(sub.relPath)
+            val privateSubFolder = folderResolver.getPrivateSubFolder(sub.secretKey, sub.relPath)
+
+            when (releaseMode) {
+                MOVE -> moveSuppress(nfsFile, privateSubFolder)
+                HARD_LINKS -> hardLinkSuppress(nfsFile, publicSubFolder)
```
Juan-EBI, line 69: you are only using the publicSubFolder in HARD_LINKS mode and privateSubFolder MOVE mode. I would suggest move variables declaration there rather that populate this scope with them where they are not needed.

submission/submission-core/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/request/SubmissionRequestReleaser.kt
```diff
@@ -111,6 +119,54 @@ class SubmissionRequestReleaser(
...
+    private suspend fun suppressSubmissionFiles(sub: ExtSubmission) {
+        suspend fun suppressFile(reqFile: SubmissionRequestFile) {
+            when (val file = reqFile.file) {
+                is NfsFile -> {
+                    val released = reqFile.copy(file = suppress(sub, reqFile.index, file), status = SUPPRESSED)
+                    rqtService.updateRqtFile(released)
+                }
+
+                is FireFile -> {
+                    if (file.published) {
```
Juan-EBI, line 141: how it be posible to have a released submission with not published files? even here after we un publish the file file will remain as `published = true` it looks to me that we do not keep this flag consistent as our source of truth is submission.

submission/submission-core/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/request/SubmissionRequestReleaser.kt
```diff
@@ -111,6 +119,54 @@ class SubmissionRequestReleaser(
...
+                    } else {
+                        rqtService.updateRqtFile(reqFile.copy(status = SUPPRESSED))
+                    }
+                }
+            }
+        }
+
+        supervisorScope {
+            filesRequestService
+                .getSubmissionRequestFiles(sub.accNo, sub.version, REUSED)
```
Juan-EBI, line 153: all other operations include `.withIndex()` to them include logs with the current index. I found that usefull personally so I will suggest to include it here.

submission/submission-core/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/request/SubmissionRequestReleaser.kt
```diff
@@ -111,6 +119,54 @@ class SubmissionRequestReleaser(
...
+        logger.info { "$accNo ${sub.owner} Started suppressing submission files over ${sub.storageMode}" }
+        suppressSubmissionFiles(sub)
+        logger.info { "$accNo ${sub.owner} Finished suppressing submission files over ${sub.storageMode}" }
+    }
+
+    private suspend fun suppressSubmissionFiles(sub: ExtSubmission) {
+        suspend fun suppressFile(reqFile: SubmissionRequestFile) {
+            when (val file = reqFile.file) {
+                is NfsFile -> {
+                    val released = reqFile.copy(file = suppress(sub, reqFile.index, file), status = SUPPRESSED)
```
Juan-EBI, line 136: I think is incorrect to use this index as this is the index in the submission not the index in the total files we need to supress

submission/submission-core/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/request/SubmissionRequestReleaser.kt
```diff
@@ -46,10 +48,16 @@ class SubmissionRequestReleaser(
         version: Int,
         processId: String,
     ) {
-        rqtService.onRequest(accNo, version, FILES_COPIED, processId, {
-            if (it.submission.released) releaseRequest(accNo, it)
+        rqtService.onRequest(accNo, version, FILES_COPIED, processId) {
+            if (it.submission.released) {
+                releaseRequest(accNo, it)
+            } else {
+                val current = queryService.findExtByAccNo(accNo, includeFileListFiles = false)
```
Juan-EBI, line 55: it looks too much to query all the submission info (except files) for only determinaning if submission has a previous version. We can query only that or use CoreInfo queery operation.

submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/nfs/NfsFtpService.kt
jhoanmanuelms: I agree with you, I just went with this name because it was the name that was given to the operation of take a public submission and make it private

submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/nfs/NfsFtpService.kt
jhoanmanuelms: Nice catch 👍

submission/submission-core/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/request/SubmissionRequestReleaser.kt
jhoanmanuelms: I check all the other operations and they use `concurrently`. Could you please share an example?

submission/submission-core/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/request/SubmissionRequestReleaser.kt
jhoanmanuelms: For a public submission that goes private, the amount of files that need to be un-released are the total files of the submission aren't they?

submission/submission-core/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/request/SubmissionRequestReleaser.kt
jhoanmanuelms: I think this was coming from when we didn't update the status per file but at the request level so we added this in order to avoid making unnecessary calls to FIRE for files that were already published. Now that we're querying by status, this should no longer be needed.

submission/submission-core/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/request/SubmissionRequestReleaser.kt
Juan-EBI: yes that is correct for that case but is not correct for others for example if a submission if modifieid and re submitted (to make files private).

client/fire-webclient/src/main/kotlin/uk/ac/ebi/fire/client/api/FireWebClient.kt
```diff
@@ -68,8 +69,8 @@ internal class FireWebClient(
         return client.putForObjectAsync<FireApiFile>("/objects/$fireOid/publish")
     }
 
-    override suspend fun unpublish(fireOid: String) {
-        client.deleteAsync("/objects/$fireOid/publish")
+    override suspend fun unpublish(fireOid: String): FireApiFile {
+        return client.deleteForObjectAsync<FireApiFile>("/objects/$fireOid/publish")
```
Juan-EBI, line 73: I think given api is already native async we do not need any more extensions to faciliated. I think we can simple do now 
```
   override suspend fun unpublish(fireOid: String): FireApiFile {
        return client.delete().uri("/objects/$fireOid/publish").retrieve().awaitBody()
    }
```

I mean extensions were good during transition period from  rest template to webclient but now IMO we can start getting familiar with real API

submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/api/FileStorageService.kt
```diff
@@ -11,6 +11,11 @@ interface FileStorageService {
         file: ExtFile,
     ): ExtFile
 
+    suspend fun suppressSubmissionFile(
```
Juan-EBI, line 14: missing file renaming here

submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/nfs/NfsFtpService.kt
```diff
@@ -25,13 +27,13 @@ class NfsFtpService(
         file: ExtFile,
     ): ExtFile =
         withContext(Dispatchers.IO) {
-            val nfsFile = file as NfsFile
+            require(file is NfsFile) { "NfsFtpService should only handle NfsFile" }
```
Juan-EBI, line 30: nice, defensive code

submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/nfs/NfsFtpService.kt
```diff
@@ -53,5 +55,39 @@ class NfsFtpService(
...
+    ): ExtFile =
+        withContext(Dispatchers.IO) {
+            require(file is NfsFile) { "NfsFtpService should only handle NfsFile" }
+            when (releaseMode) {
+                MOVE -> moveSuppress(file, sub)
+                HARD_LINKS -> hardLinkSuppress(file, sub)
+            }
+        }
+
+    private fun moveSuppress(
```
Juan-EBI, line 70: missing renaming here

submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/service/StorageService.kt
```diff
@@ -41,6 +41,16 @@ class StorageService(
         }
     }
 
+    override suspend fun suppressSubmissionFile(
```
Juan-EBI, line 44: missing renaming here

submission/persistence-filesystem/src/test/kotlin/ac/uk/ebi/biostd/persistence/filesystem/fire/FireFtpServiceTest.kt
```diff
@@ -28,37 +28,66 @@ class FireFtpServiceTest(
...
 
-        assertThat(file.published).isTrue()
-        assertThat(file.firePath).isEqualTo("fire-path")
-    }
+            assertThat(file.published).isTrue()
+            assertThat(file.firePath).isEqualTo("fire-path")
+        }
+
+    @Test
+    fun `suppress submission file`() =
```
Juan-EBI, line 63: `unrelease submission file`

submission/persistence-filesystem/src/test/kotlin/ac/uk/ebi/biostd/persistence/filesystem/nfs/NfsFtpServiceTest.kt
```diff
@@ -114,12 +137,36 @@ internal class NfsFtpServiceTest(
                 assertThat(result).content().isEqualTo("file content")
                 assertThat(file).doesNotExist()
             }
-        }
+
+        @Test
+        fun `suppress submission file`() =
```
Juan-EBI, line 142: `unrelease`

submission/persistence-filesystem/src/test/kotlin/ac/uk/ebi/biostd/persistence/filesystem/nfs/NfsFtpServiceTest.kt
```diff
@@ -114,12 +137,36 @@ internal class NfsFtpServiceTest(
...
+                        relPath = "Files/move-test.txt",
+                        file = file,
+                        fullPath = file.absolutePath,
+                        md5 = file.md5(),
+                        size = file.size(),
+                    )
+
+                val privateFolder = folderResolver.getPrivateSubFolder(SECRET_KEY, REL_PATH).resolve(FILES_PATH)
+                val suppressedPath = privateFolder.resolve("move-test.txt")
+                val suppressed = testInstance.unReleaseSubmissionFile(subInfo, nfsFile) as NfsFile
```
Juan-EBI, line 159: I realize I never commended this but something that I do in test in separate the classic test part by line break (arrange, action and assert). I would suggest we addopt this. 

https://automationpanda.com/2020/07/07/arrange-act-assert-a-pattern-for-writing-good-tests/

submission/submission-core/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/request/SubmissionRequestCleanIndexer.kt
```diff
@@ -63,6 +64,7 @@ class SubmissionRequestCleanIndexer(
         newFiles: FilesRecords,
         current: ExtSubmission,
     ): Triple<Int, Int, Int> {
+        val reusedIdx = AtomicInteger(0)
```
Juan-EBI, line 67: if you noticed we return conflict and deprecated values so latter we know what is the amount of files to delete so we can acurrectly calculate the progress of procress. With reused files we are not so when we are actually unreleased we do not know how many files are pending.  


Not saying this need to be included in this ticket which was already a huge effort but I think having precise estimation of pending activities is something we do not want to lose.  Create a ticket to address that after this?

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/ResubmissionApiTest.kt
```diff
@@ -245,36 +247,65 @@ class ResubmissionApiTest(
...
-        val version1 =
-            tsv {
-                line("Submission", "S-RSTST5")
-                line("Title", "Public Submission")
-                line("ReleaseDate", OffsetDateTime.now().toStringDate())
-                line()
-                line("Study")
-                line()
-            }.toString()
+    fun `5-5 super user suppresses submission from another user`() =
```
Juan-EBI, line 250: unrelease

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/ResubmissionApiTest.kt
```diff
@@ -245,36 +247,65 @@ class ResubmissionApiTest(
...
+                    line()
+                    line("File", "file_5-5-1.txt")
+                    line()
+                }.toString()
+
+            assertThat(onBehalfClient.submitSingle(version2, TSV)).isSuccessful()
+
+            val submittedV2 = submissionRepository.getExtByAccNo("S-RSTST5")
+
+            assertThat(File("$ftpPath/${submittedV2.relPath}/Files/file_5-5-1.txt")).doesNotExist()
```
Juan-EBI, line 302: maybe we should test that the whole folder is empty

```
assertThat(File("$ftpPath/${submittedV2.relPath}/Files")).isEmptyDirectory()
```

Juan-EBI (approved):
LGTM please consider comments before merge

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/ResubmissionApiTest.kt
jhoanmanuelms: I don't think we should change it in this case since the integration tests represent the business level operation and in that level it was decided the name for the operation to make a public submission private would be "suppress"

submission/submission-core/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/request/SubmissionRequestCleanIndexer.kt
jhoanmanuelms: you're right, I added the changes 👍

PR #852, opened 2024-07-02
Pivotal ID #187764871: SLURM Migration

Juan-EBI:
https://www.pivotaltracker.com/story/show/187764871
- added slurm support


client/cluster-client/src/main/kotlin/uk/ac/ebi/biostd/client/cluster/api/SlurmClusterClient.kt
```diff
@@ -0,0 +1,119 @@
...
+            if (status.isBlank()) return null else return status
+        }
+
+        suspend fun CommandRunner.historicalJobStatus(): String {
+            val command = "sacct --noheader --format=JobID,State --jobs=$jobId | grep \"^$jobId \" | awk '{print \$2}'"
+            return executeCommand(command).second
+        }
+
+        return sshClient.runInSession {
+            val status = runningJobStatus() ?: historicalJobStatus()
```
jhoanmanuelms, line 60: I think we should handle the case when the given job id is not found

client/cluster-client/src/test/kotlin/uk/ac/ebi/biostd/client/cluster/common/JobResponseParserTest.kt
```diff
@@ -10,4 +10,11 @@ internal class JobResponseParserTest {
         assertThat(job.id).isEqualTo("9148559")
         assertThat(job.queue).isEqualTo("research-rh7")
     }
+
+    @Test
+    fun toJobSlurmRespone() {
```
jhoanmanuelms, line 15: Missing negative test

client/cluster-client/src/main/kotlin/uk/ac/ebi/biostd/client/cluster/api/SlurmClusterClient.kt
```diff
@@ -0,0 +1,119 @@
...
+import uk.ac.ebi.biostd.client.cluster.model.JobSpec
+import java.time.Duration.ofSeconds
+import kotlin.Result.Companion.failure
+import kotlin.Result.Companion.success
+
+private val logger = KotlinLogging.logger {}
+
+class SlurmClusterClient(
+    private val logsPath: String,
+    private val sshMachine: String,
```
jhoanmanuelms, line 18: We should change this parameter to `sshServer` to make it consistent with the properties in the config file. WDYT?

submission/submission-config/src/main/kotlin/ac/uk/ebi/biostd/common/properties/ClusterProperties.kt
```diff
@@ -11,6 +11,24 @@ data class ClusterProperties(
     val enabled: Boolean,
     val user: String,
     val key: String,
-    val server: String,
```
jhoanmanuelms, line 14: Since we're executing the tasks in either one cluster or the other, I think we shouldn't change this parameter but just set the corresponding property in the config files.

Another reason that makes me think this is that, eventually, we'll get rid of LSF entirely or add another cluster implementation so we would need to keep adding them here

submission/submission-config/src/main/kotlin/ac/uk/ebi/biostd/common/properties/ClusterProperties.kt
```diff
@@ -11,6 +11,24 @@ data class ClusterProperties(
...
+    val lsfServer: String,
+    val slurmServer: String,
     val logsPath: String,
+    val default: Cluster,
```
jhoanmanuelms, line 17: Following my previous comment, this property should be named something `type` or `mode` since this is decided at config level rather than runtime

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/cluster/web/ClusterExecutor.kt
```diff
@@ -1,17 +1,20 @@
...
+import ac.uk.ebi.biostd.common.properties.Cluster
+import ac.uk.ebi.biostd.submission.config.GeneralConfig
 import uk.ac.ebi.biostd.client.cluster.api.LsfClusterClient
+import uk.ac.ebi.biostd.client.cluster.api.SlurmClusterClient
 import uk.ac.ebi.biostd.client.cluster.model.Job
 import uk.ac.ebi.biostd.client.cluster.model.JobSpec
 
-class ClusterExecutor private constructor(val lsfClusterClient: LsfClusterClient) {
+class ClusterExecutor private constructor(
+    val lsfClusterClient: LsfClusterClient,
```
jhoanmanuelms, line 12: And finally, we could use the interface and pass the correct implementation based on the cluster mode rather than checking it here

jhoanmanuelms (commented):
Changes looking great. I have added a couple of comments

client/cluster-client/src/main/kotlin/uk/ac/ebi/biostd/client/cluster/api/SlurmClusterClient.kt
Juan-EBI: I have no particular stance on this so change looks good for me.

submission/submission-config/src/main/kotlin/ac/uk/ebi/biostd/common/properties/ClusterProperties.kt
Juan-EBI: I see your point but I think during transition period better two have an option for both to quickly change. During my testing of slurm infra team did find issues and server was not functionall so I suspect having both options for a while will be usefull. Once feature (and service in the EBI) stabilize I do agree better to a single one.

submission/submission-config/src/main/kotlin/ac/uk/ebi/biostd/common/properties/ClusterProperties.kt
Juan-EBI: sure I will rename I just think transition will not be a move and forget and even team is still migrated resources from cluster and fixing issues.

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/cluster/web/ClusterExecutor.kt
Juan-EBI: I create this class with the main purpose of actually supporting both cluster and being able to test both, I think will this class will be useless once codon disapeer completly but I do think it will be usefull during transition period.

client/cluster-client/src/main/kotlin/uk/ac/ebi/biostd/client/cluster/api/SlurmClusterClient.kt
Juan-EBI: if job is not found bash command output is empty strimg so status will be "". By handling do mean trigger an exception in this case or does it look for you that emtpy is correct.

client/cluster-client/src/test/kotlin/uk/ac/ebi/biostd/client/cluster/common/JobResponseParserTest.kt
Juan-EBI: added test

submission/submission-config/src/main/kotlin/ac/uk/ebi/biostd/common/properties/ClusterProperties.kt
jhoanmanuelms: I agree, let's then create a ticket for removing it in the future

jhoanmanuelms (approved):
LGTM

Juan-EBI:
create ticket https://www.pivotaltracker.com/story/show/187907319

PR #855, opened 2024-07-12
Pivotal ID # 185361632: Avoid Files Modification On Public Studies

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/185361632

- Create a new stage to validate file changes
- Add new permission to allow file modifications on public studies
- Mark the submission as invalid in case the file validation fails
- Add integration tests


submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/ResubmissionApiTest.kt
```diff
@@ -383,4 +394,82 @@ class ResubmissionApiTest(
...
+                }.toString()
+
+            webClient.uploadFile(tempFolder.createOrReplaceFile("file_5-7-1.txt", "5-7-1 file updated content"))
+            webClient.submitAsync(version2, TSV)
+
+            waitUntil(timeout = TWO_SECONDS) { requestRepository.getRequestStatus("S-RSTST8", 2) == INVALID }
+        }
+
+    @Test
+    fun `5-8 authorized user updates public submission files`() =
```
Juan-EBI, line 438: I think we need  3  more scenarios. 
1. Allow pagetab files to be modified (without specific permisions).
2. Allow file list metadata to be modified (without specific permisions).
3. File list files can not be modified

submission/submission-core/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/request/SubmissionRequestValidator.kt
```diff
@@ -0,0 +1,66 @@
...
+
+        requestService.onRequest(accNo, version, RequestStatus.INDEXED_CLEANED, processId) {
+            requestStatus = validateRequest(it)
+            RqtUpdate(it.withNewStatus(requestStatus))
+        }
+
+        if (requestStatus == VALIDATED) eventsPublisherService.requestValidated(accNo, version)
+    }
+
+    internal suspend fun validateRequest(rqt: SubmissionRequest): RequestStatus {
```
Juan-EBI, line 38: smalI detail but I do not see any value in this method to be honest,  single line that call another method with basically doing anything

Juan-EBI (commented):

submission/submission-core/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/request/SubmissionRequestValidator.kt
```diff
@@ -0,0 +1,66 @@
...
+    suspend fun validateRequest(
+        accNo: String,
+        version: Int,
+        processId: String,
+    ) {
+        var requestStatus = INVALID
+
+        requestService.onRequest(accNo, version, RequestStatus.INDEXED_CLEANED, processId) {
+            requestStatus = validateRequest(it)
+            RqtUpdate(it.withNewStatus(requestStatus))
```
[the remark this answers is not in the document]
jhoanmanuelms: I agree, this is better to handle in another ticket. Actually, this will be part of the async validation implementation. The aim of the current PR is just to avoid files modifications

submission/submission-core/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/request/SubmissionRequestValidator.kt
jhoanmanuelms: Yeah, I scratched my head with this one for a while. I left it there because I wanted to leave the template for the future when we plug in the other validations but it's not required now and we can refactor when we actually need it. I'll remove it 👍

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/ResubmissionApiTest.kt
jhoanmanuelms: Number 1 is [already there](https://github.com/EBIBioStudies/biostudies-backend-services/blob/master/submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/ResubmissionApiTest.kt#L312)

Number 2 and 3 were added to the latest commit

Juan-EBI (approved):
LGTM please consider minor changes before merge

PR #863, opened 2024-08-12
Pivotal ID # 187970878: Allow to suppress submission published/partial update notification for specific  submission

Juan-EBI:
- refactor code to use same classes in backend and web client
- refactor code to use parameter class rather than individual pameters


jhoanmanuelms (approved):
LGTM

PR #865, opened 2024-08-13
Pivotal ID # 187898943: Submission Files Integrity Check

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/187898943

- Check integrity of files and directories persisted on NFS
- Check integrity of generated zip files
- Add tests
- Remove unused code

### DIFF
--- a/commons/commons-util/src/main/kotlin/ebi/ac/uk/io/FileUtils.kt
+++ b/commons/commons-util/src/main/kotlin/ebi/ac/uk/io/FileUtils.kt
@@ -1,5 +1,6 @@
 package ebi.ac.uk.io
 
+import ebi.ac.uk.exception.CorruptedFileException
 import ebi.ac.uk.io.FileUtilsHelper.createDirectories
 import ebi.ac.uk.io.FileUtilsHelper.createFileHardLink
 import ebi.ac.uk.io.FileUtilsHelper.createFolderHardLinks
@@ -7,6 +8,7 @@ import ebi.ac.uk.io.FileUtilsHelper.createFolderIfNotExist
 import ebi.ac.uk.io.FileUtilsHelper.createParentDirectories
 import ebi.ac.uk.io.FileUtilsHelper.createSymLink
 import ebi.ac.uk.io.ext.isEmpty
+import ebi.ac.uk.io.ext.md5
 import ebi.ac.uk.io.ext.notExist
 import ebi.ac.uk.io.ext.size
 import mu.KotlinLogging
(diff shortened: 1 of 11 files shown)


commons/commons-util/src/main/kotlin/ebi/ac/uk/io/FileUtils.kt
```diff
@@ -104,6 +106,18 @@ object FileUtils {
         }
     }
 
+    fun checkFileIntegrity(target: File, expectedMd5: String) {
+        require(target.md5() == expectedMd5) { throw CorruptedFileException(target) }
+    }
+
+    fun checkDirectoryIntegrity(
```
Juan-EBI, line 113: same as above

submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/fire/ZipUtil.kt
```diff
@@ -39,4 +46,20 @@ object ZipUtil {
...
+            listAllFilesWithRelPaths(sourceDir).forEach {
+                val (path, sourceFile) = it
+                val entry = zip.getEntry(path)
+
+                if (entry == null || sourceFile.md5() != zip.md5(entry)) throw CorruptedFileException(sourceFile)
+            }
+        }
+    }
+
+    private fun ZipFile.md5(entry: ZipEntry): String = getInputStream(entry).use { DigestUtils.md5Hex(it).uppercase() }
```
Juan-EBI, line 64: with this change we are calculating the md5 twice, this is impacting significantly the zip time for large folder. It looks like a bad idea IMO, still if that is what need it well we can do it but IMO this is doing more harm than good to the system. Ugis will have lass optinion

submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/fire/ZipUtil.kt
```diff
@@ -22,6 +27,8 @@ object ZipUtil {
                     zs.closeEntry()
                 }
         }
+
+        checkZipIntegrity(sourceDir, zipFile)
```
Juan-EBI, line 31: this is a generic zip util I think we should not include the double file check as part of the utility but rather do it in biostudies business code.

Juan-EBI (commented):

jhoanmanuelms:
We should come back to this PR when we actually start working on the ticket

PR #866, opened 2024-08-14
Pivotal ID # 188114384: Scheduler SLURM Migration

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/188114384

- Create common Cluster config class
- Add option to execute scheduler jobs in SLURM
- Remove unused code

### DIFF
--- a/client/cluster-client/src/main/kotlin/uk/ac/ebi/biostd/client/cluster/api/LsfClusterClient.kt
+++ b/client/cluster-client/src/main/kotlin/uk/ac/ebi/biostd/client/cluster/api/LsfClusterClient.kt
@@ -90,7 +90,6 @@ class LsfClusterClient(
     companion object {
         const val DONE_STATUS = "DONE"
         const val EXIT_STATUS = "EXIT"
-        const val PEND_STATUS = "PEND"
 
         fun create(
             sshKey: String,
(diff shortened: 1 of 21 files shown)


client/cluster-client/src/main/kotlin/uk/ac/ebi/biostd/client/cluster/api/LsfClusterClient.kt
```diff
@@ -90,7 +90,6 @@ class LsfClusterClient(
     companion object {
         const val DONE_STATUS = "DONE"
         const val EXIT_STATUS = "EXIT"
-        const val PEND_STATUS = "PEND"
```
Juan-EBI, line 93: why we have remove this one? (I remember seeing as a valid state)

Juan-EBI (approved):
LGTM

client/cluster-client/src/main/kotlin/uk/ac/ebi/biostd/client/cluster/api/LsfClusterClient.kt
jhoanmanuelms: It was never used. Maybe we need to do something with it

Juan-EBI: yes is not used in a particular test but it is used when obtained job status using API

PR #871, opened 2024-08-22
Pivotal ID # 188156440}: re submission when INVALID state

Juan-EBI:
https://www.pivotaltracker.com/story/show/188156440
- added invalid as a processing state


jhoanmanuelms (approved):
LGTM. Nice catch! 👍

PR #872, opened 2024-08-23
Pivotal ID # 188035630: Request Status CLI Operation

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/188035630

- Create command to get a request status
- Fix operation method in the web client
- Handle scenario for request not found

### DIFF
--- a/client/bio-webclient/src/main/kotlin/ac/uk/ebi/biostd/client/api/SubmissionRequestClient.kt
+++ b/client/bio-webclient/src/main/kotlin/ac/uk/ebi/biostd/client/api/SubmissionRequestClient.kt
@@ -16,7 +16,7 @@ class SubmissionRequestClient(
         version: Int,
     ): RequestStatus =
         client
-            .post()
+            .get()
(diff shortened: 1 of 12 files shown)


Juan-EBI (approved):
LGTM

client/bio-webclient/src/main/kotlin/ac/uk/ebi/biostd/client/api/SubmissionRequestClient.kt
```diff
@@ -16,7 +16,7 @@ class SubmissionRequestClient(
         version: Int,
     ): RequestStatus =
         client
-            .post()
+            .get()
```
Juan-EBI, line 19: looks like we need an itest for this one. I had created https://www.pivotaltracker.com/story/show/188165943

jhoanmanuelms: I just added the itest as part of this ticket 👍

PR #877, opened 2024-09-06
Pivotal ID # 187642945: Partition Cluster Jobs Logs Path

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/187642945

- Create a wrapper to execute commands in the cluster
- Use the job id to provide the dynamic output path
- Add a new property to specify the wrapper script location

### DIFF
--- a/ci/slurm_wrapper.sh
+++ b/ci/slurm_wrapper.sh
@@ -2,7 +2,7 @@
 
 #
 # Executes the given command in the SLURM cluster placing the files in a dynamic location based on the job id last
-# three digits.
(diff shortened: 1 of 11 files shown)


ci/slurm_wrapper.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+#
+# Executes the given command in the SLURM cluster placing the files in a dynamic location based on the job id last
+# three digits.
```
Juan-EBI, line 5: will be nice to include an example of where the files will be located for a job with id 12345

PR #881, opened 2024-09-17
Pivotal ID # 181219386: Clean Submission Temp Files

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/181219386

- Add a method to clean folders
- Generate pagetab in the corresponding submission folder
- Clean submission temp folder when it's archived
- Set a scheduler to clean up temp folders every week


submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/archive/ArchiveScheduler.kt
```diff
@@ -22,9 +28,20 @@ class ArchiveScheduler(
             persistenceService
                 .findAllProcessed()
                 .onEach { (accNo, version) -> logger.info { "Archiving request $accNo, $version" } }
-                .collect({ (accNo, version) ->
+                .collect { (accNo, version) ->
+                    val subTempFolder = Paths.get(props.fire.tempDirPath).resolve("$accNo/$version").toFile()
+                    FileUtils.deleteFile(subTempFolder)
```
Juan-EBI, line 33: I do not think we should fix mix archieve with delete submission temporally files. Specially because file deletion can take long time and is a blocking Io operation while archieve is not blocking and is database one. 

Personally I think we should delete files older than X and include this in cleanTempFolders operation.

Juan-EBI (commented):
Please check comment

Juan-EBI (approved):
LGTM

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/archive/ArchiveScheduler.kt
jhoanmanuelms: What I meant was that I agree to avoid blocking the operation, I will create a separate method for this so we archive in one place and delete in the other. Sorry for the miscommunication.

PR #882, opened 2024-09-18
Pivotal ID # 185645993: Kotlin AWS S3 Client

Juan-EBI:
https://www.pivotaltracker.com/story/show/185645993
- migrated code to kotlin sdk


client/fire-webclient/src/main/kotlin/uk/ac/ebi/fire/client/api/S3JClient.kt
```diff
@@ -3,21 +3,24 @@ package uk.ac.ebi.fire.client.api
...
 import com.amazonaws.services.s3.model.S3Object
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
 import uk.ac.ebi.fire.client.integration.web.FireS3Client
 import java.io.File
 import java.nio.file.Files
 import java.nio.file.StandardCopyOption
 
-class S3Client(
+class S3JClient(
```
jhoanmanuelms, line 13: Are you considering adding a feature flag in another PR to choose between one client or the other? Otherwise, we could delete this class since, after this change, it will only be used by tests

jhoanmanuelms (commented):
Looking good! Just a couple of questions

jhoanmanuelms (approved):
LGTM

PR #884, opened 2024-09-25
Pivotal ID # 188325315: Single cluster job mode

Juan-EBI:
https://www.pivotaltracker.com/story/show/188325315
- refactor event publish out of request handlers to faciliate full submission process execution.
- added new complete mode to finish submission in a single task


commons/commons-bio/src/main/kotlin/ebi/ac/uk/api/SubmitWebRequest.kt
```diff
@@ -20,6 +20,7 @@ data class SubmitParameters(
     val attributes: List<SubmitAttribute> = emptyList(),
     val storageMode: StorageMode? = null,
     val silentMode: Boolean? = null,
+    val processAll: Boolean? = null,
```
jhoanmanuelms, line 23: I would suggest `sigleJob`. The name `processAll` gives me the impression there are several submissions in the request and we're deciding whether to process them all

submission/submission-core/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/submitter/LocalExtSubmissionSubmitter.kt
```diff
@@ -63,67 +65,143 @@ class LocalExtSubmissionSubmitter(
         return requestService.createRequest(request)
     }
 
+    @Suppress("CyclomaticComplexMethod")
+    override suspend fun completeRqt(
+        accNo: String,
+        version: Int,
+    ) {
+        suspend fun fromSavedSubmission() {
```
jhoanmanuelms, line 73: Nice abstraction! 👍 

Could we move these to a different location? Maybe in the companion? Even though it clarifies what's going on and where the request is starting from, it was a bit hard to read at one glance. I had to read the code a couple of times to understand where the request was being processed as a whole. 

Having these methods in a separate place, would make it cleaner and clearer for someone trying to understand the flow starting from [here](https://github.com/EBIBioStudies/biostudies-backend-services/pull/884/files#diff-cce2eacbefada70c4273ebfc61a1bf521a71ed09054adb7287f259b748f2bbb8R119)

jhoanmanuelms (approved):
LGTM! Nice improvements, I think the stages will be easier to control now.

Please consider testing a couple of cluster submissions on BETA before merging since we currently don't have itests that cover the cluster flow  👍

submission/submission-core/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/submitter/LocalExtSubmissionSubmitter.kt
JuanCamiloRada: they can not go in the companion as they use the requests handlers and companion only allow static, another option is to put them in the class itself but IMO they will add too much noise as this methods only are related to this operation. Any suggestion is welcome

PR #889, opened 2024-10-15
Pivotal ID # 188311113: Crossref "Contributor" Element Registration

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/188311113

- Allow single name authors for DOI registration
- If a single name is provided, it will be used as surname


Juan-EBI (approved):
LGTM

PR #890, opened 2024-10-15
Pivotal ID # 181219386: Clean Submission Temp Files

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/181219386

- Delete files in the request-files folder
- Create temp files for simple submission read in the system temp folder rather than the request folder

### DIFF
--- a/commons/commons-serialization-util/src/main/kotlin/uk/ac/ebi/serialization/common/FilesResolver.kt
+++ b/commons/commons-serialization-util/src/main/kotlin/uk/ac/ebi/serialization/common/FilesResolver.kt
@@ -1,12 +1,8 @@
 package uk.ac.ebi.serialization.common
 
 import ebi.ac.uk.io.ext.createTempFile
-import ebi.ac.uk.io.ext.resolveMany
 import java.io.File
-import java.time.LocalDate
-import java.time.temporal.ChronoField.DAY_OF_MONTH
-import java.time.temporal.ChronoField.MONTH_OF_YEAR
-import java.time.temporal.ChronoField.YEAR
+import java.nio.file.Files
 
 class FilesResolver(private val basePath: File) {
     fun createExtEmptyFile(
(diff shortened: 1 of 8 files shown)


commons/commons-serialization-util/src/main/kotlin/uk/ac/ebi/serialization/common/FilesResolver.kt
```diff
@@ -15,23 +11,14 @@ class FilesResolver(private val basePath: File) {
...
-        val now = LocalDate.now()
-        val path =
-            basePath
-                .resolve(now.get(YEAR).toString())
-                .resolve(now.get(MONTH_OF_YEAR).toString())
-                .resolve(now.get(DAY_OF_MONTH).toString())
-                .resolveMany(*keys)
-        path.mkdirs()
-        return path
+        return Files.createTempFile(name, ".json").toFile()
```
Juan-EBI, line 22: why we are migrating out of a deterministic folder and start using machine temp files?

commons/commons-serialization-util/src/main/kotlin/uk/ac/ebi/serialization/common/FilesResolver.kt
```diff
@@ -15,23 +11,14 @@ class FilesResolver(private val basePath: File) {
         fileName: String,
     ): File {
         val name = fileName.replace("/", "-")
-        return baseFolder(subAccNo, version.toString()).createTempFile(name, ".json")
+        val baseFolder = basePath.resolve("$subAccNo/$version")
+        baseFolder.mkdirs()
+
+        return baseFolder.createTempFile(name, ".json")
```
Juan-EBI, line 17: what was the problem with previous base folder calculation? why remove date?

Juan-EBI (commented):
I do not undertsdant changes intention may you clarify it please

commons/commons-serialization-util/src/main/kotlin/uk/ac/ebi/serialization/common/FilesResolver.kt
jhoanmanuelms: 1. it's easier to identify which files belong to a specific submission request based on the accNo/version combination
2. it's the same approach we're using for temporary fire files generation (such as zip)
3. since we're using the default temp file creation mechanism, they will still have a unique id

commons/commons-serialization-util/src/main/kotlin/uk/ac/ebi/serialization/common/FilesResolver.kt
Juan-EBI: ahh got it, maybe them docunent this method is only for being use for random temporally json files and can be deleted without previous files.

commons/commons-serialization-util/src/main/kotlin/uk/ac/ebi/serialization/common/FilesResolver.kt
Juan-EBI: maybe reanme function to create request tmp file?

Juan-EBI (approved):
LGTM please consider minor comments

PR #895, opened 2024-11-04
Pivotal ID # 188521286: OnBehalf with diferent submission should update owner

Juan-EBI:
- added test


submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/SubmissionOnBehalfTest.kt
```diff
@@ -181,4 +181,48 @@ class SubmissionOnBehalfTest(
             assertThat(testFile).exists()
             assertThat(testFile).hasContent("submitter data")
         }
+
+    @Test
+    fun `14-5 On behalf with manager with another user submission update user Owner`() =
```
Juan-EBI, line 186: @jhoanmanuelms this is the expected behavoir according to ugis so I added test to garantee we do not break this

jhoanmanuelms (approved):
LGTM

PR #898, opened 2024-11-04
Pivotal ID # 188522739: Submission stats include number of directories

Juan-EBI:
- extracted task remote executor to be able to execute other submitter task


jhoanmanuelms (approved):
LGTM 👍

PR #902, opened 2024-11-06
Pivotal ID # 188531654: Use AccNo For Temp Files

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/188531654

Use calculated accNo to generate temp files


submission/submission-core/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/submission/SubmissionProcessor.kt
```diff
@@ -68,7 +68,7 @@ class SubmissionProcessor(
             creationTime = creationTime,
             tags = submission.tags.map { ExtTag(it.first, it.second) },
             collections = tags.map { ExtCollection(it) },
-            section = toExtSectionMapper.convert(submission.accNo, version, submission.section, sources),
+            section = toExtSectionMapper.convert(accNoString, version, submission.section, sources),
```
Juan-EBI, line 71: wondering how this line that has not been changed like in more than a year is generating an issue now. Kindly may you elaborate?

Maybe it will help if we create a test case that actually reproduce the error we are trying to solve.

jhoanmanuelms: done

Juan-EBI (approved):
LGTM

PR #906, opened 2024-11-14
Pivotal ID # 188526717: Improve Drafts Timeline Structure

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/188526717

- Add modification date to the drafts
- Use readable timestamp as the draft key
- Use reactive web client for the draft operations
- Remove unused code


Juan-EBI (approved):
LGTM

PR #908, opened 2024-11-27
Pivotal ID # 188624657: Aws s3 use CRT client

Juan-EBI:
https://www.pivotaltracker.com/story/show/188624657
- added crt client
- added retrigger async support


client/fire-webclient/build.gradle.kts
```diff
@@ -27,8 +27,10 @@ dependencies {
     api(project(CommonsUtil))
     api(project(CommonsHttp))
 
-    implementation(AwsS3K)
-    implementation(AwsS3)
+    implementation(AwsS3K) {
+        exclude("com.squareup.okhttp3:okhttp")
```
jhoanmanuelms, line 31: Is it possible to use a constant for the exclude section as well? Or is it better to leave it explicit for these cases?

jhoanmanuelms (approved):
LGTM 👍. I left a question just out of curiosity

client/fire-webclient/build.gradle.kts
Juan-EBI: we can have constant name but then as your name include version we will need to remove it so having two contains one with name and one without

PR #910, opened 2024-12-04
Pivotal ID # 184033688: Async Validation Process

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/184033688

- Set the request processing field as nullable
- Add the draft related status
- Create the SubmissionRequestDraftService which will replace the SubmissionDraftService when drafts and requests are merged

### DIFF
--- a/commons/commons-bio/src/main/kotlin/ebi/ac/uk/model/RequestStatus.kt
+++ b/commons/commons-bio/src/main/kotlin/ebi/ac/uk/model/RequestStatus.kt
@@ -1,6 +1,9 @@
 package ebi.ac.uk.model
 
 enum class RequestStatus {
+    DRAFT,
+    SUBMITTED,
+    ACCEPTED,
     REQUESTED,
     INDEXED,
     LOADED,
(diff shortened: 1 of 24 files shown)


commons/commons-bio/src/main/kotlin/ebi/ac/uk/model/RequestStatus.kt
```diff
@@ -1,6 +1,9 @@
 package ebi.ac.uk.model
 
 enum class RequestStatus {
+    DRAFT,
+    SUBMITTED,
+    ACCEPTED,
```
Juan-EBI, line 6: a draft accepted is the same as request requrested is not it?

submission/submission-core/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/request/SubmissionRequestCleanIndexer.kt
```diff
@@ -43,7 +43,7 @@ class SubmissionRequestCleanIndexer(
         processId: String,
     ): SubmissionRequest =
         requestService.onRequest(accNo, version, RequestStatus.LOADED, processId) {
-            val indexedRequest = indexRequest(it.process.submission)
+            val indexedRequest = indexRequest(it.process!!.submission)
```
Juan-EBI, line 46: there is something missing in the modeling IMO that we need to use so much the it.process!!, I am afraid I can not see it now but we can take a look and improve after the completion of this feature

Juan-EBI (commented):
LGTM please check minor comment

submission/submission-core/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/request/SubmissionRequestCleanIndexer.kt
jhoanmanuelms: I completely agree. Ideally we should be able to have a class with and without without too much mapping

commons/commons-bio/src/main/kotlin/ebi/ac/uk/model/RequestStatus.kt
jhoanmanuelms: Yes, we don't need this stage anymore

Juan-EBI (approved):
LGTM

PR #912, opened 2024-12-15
Pivotal ID # 188643993: Migrate user folder / migrate stats

Juan-EBI:
https://www.pivotaltracker.com/story/show/188643993
- added stats endpoint
- added migrate endpoint


submission/submission-security/src/main/kotlin/ebi/ac/uk/security/integration/components/ISecurityService.kt
```diff
@@ -6,10 +6,12 @@ import ebi.ac.uk.api.security.LoginRequest
 import ebi.ac.uk.api.security.RegisterRequest
 import ebi.ac.uk.api.security.ResetPasswordRequest
 import ebi.ac.uk.api.security.RetryActivationRequest
+import ebi.ac.uk.model.MigrateHomeOptions
 import ebi.ac.uk.model.User
 import ebi.ac.uk.security.integration.model.api.SecurityUser
 import ebi.ac.uk.security.integration.model.api.UserInfo
 
+@Suppress("TooManyFunctions")
```
jhoanmanuelms, line 14: I've noticed we have a lot of these. Should we consider incrementing the number in the config instead of suppressing it everywhere? 🤔 

It looks like our design methods need more functions 😁

submission/submission-security/src/main/kotlin/ebi/ac/uk/security/service/ProfileService.kt
```diff
@@ -39,6 +43,38 @@ class ProfileService(
...
+        }
+    }
+
+    private fun calculateFolderStats(nfsUserFolder: File): FolderStats {
+        var totalFiles = 0
+        var totalSize = 0L
+        var totalDirectories = 0
+        var latestModificationTime = 0L
+
+        nfsUserFolder
```
jhoanmanuelms, line 60: Kotlin 👌

submission/submission-security/src/main/kotlin/ebi/ac/uk/security/service/SecurityQueryService.kt
```diff
@@ -19,24 +20,28 @@ class SecurityQueryService(
     override fun existsByEmail(
         email: String,
         onlyActive: Boolean,
-    ): Boolean {
```
jhoanmanuelms, line 22: I've been putting these on braces in my last PRs and you've been putting them in single liners 😆 

Do you think we should agree on this? I prefer braces because, in my experience, the debugger acts weirdly in single-liners

commons/commons-bio/src/main/kotlin/ebi/ac/uk/model/MigrateHomeOptions.kt
```diff
@@ -0,0 +1,6 @@
+package ebi.ac.uk.model
+
+data class MigrateHomeOptions(
+    val storageMode: String,
+    val onlyEmptyFolder: Boolean = true,
```
jhoanmanuelms, line 5: I'd suggest `onlyIfEmptyFolder` since the current name suggests the task will only migrate the empty folders and not that it will fail based on whether the folder is empty

jhoanmanuelms (approved):
LGTM. Just a couple of suggestions

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/user/UserAdminApiTest.kt
```diff
@@ -0,0 +1,103 @@
...
+import org.junit.jupiter.api.extension.ExtendWith
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.boot.test.context.SpringBootTest
+import org.springframework.boot.test.context.SpringBootTest.WebEnvironment.RANDOM_PORT
+import org.springframework.boot.test.web.server.LocalServerPort
+import org.springframework.test.context.junit.jupiter.SpringExtension
+
+@ExtendWith(SpringExtension::class)
+@SpringBootTest(webEnvironment = RANDOM_PORT)
+class UserAdminApiTest(
```
jhoanmanuelms, line 29: These tests should be added to the inventory

commons/commons-bio/src/main/kotlin/ebi/ac/uk/model/MigrateHomeOptions.kt
Juan-EBI: totally agree

submission/submission-security/src/main/kotlin/ebi/ac/uk/security/integration/components/ISecurityService.kt
Juan-EBI: make sense

PR #917, opened 2025-01-15
Pivotal ID # 184033688: Async Validation Process

jhoanmanuelms:
- Move the accNo / version calculation to the moment when the request is built
- Store errors in the request
- Add SUBMITTED to the processing status

### DIFF
--- a/submission/submission-core/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/submission/SubmissionService.kt
+++ b/submission/submission-core/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/submission/SubmissionService.kt
@@ -32,17 +32,15 @@ class SubmissionService(
         return submitter.handleRequest(accNo, version)
     }
 
-    suspend fun submitAsync(rqt: SubmitRequest): SubmissionId {
+    suspend fun submitAsync(rqt: SubmitRequest) {
         logger.info { "${rqt.accNo} ${rqt.owner} Received async submit request with accNo '${rqt.accNo}'" }
         val (accNo, version) = submitter.processRequestDraft(rqt)
         submitter.handleRequestAsync(accNo, version)
-        return SubmissionId(accNo, version)
     }
 
-    suspend fun submitAsync(rqt: List<SubmitRequest>): List<SubmissionId> {
+    suspend fun submitAsync(rqt: List<SubmitRequest>) {
         val requests = rqt.map { submitter.processRequestDraft(it) }.map { SubmissionId(it.accNo, it.version) }
         submitter.handleManyAsync(requests)
-        return requests
     }
 
     suspend fun deleteSubmission(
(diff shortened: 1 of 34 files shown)


submission/submission-core/src/main/kotlin/ac/uk/ebi/biostd/submission/domain/submission/SubmissionService.kt
```diff
@@ -32,17 +32,15 @@ class SubmissionService(
         return submitter.handleRequest(accNo, version)
     }
 
-    suspend fun submitAsync(rqt: SubmitRequest): SubmissionId {
+    suspend fun submitAsync(rqt: SubmitRequest) {
```
Juan-EBI, line 35: why are we removing the return type? this is actually used by api user to know which version was the one being submitted.

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/SubmissionDraftApiTest.kt
```diff
@@ -243,4 +244,48 @@ class SubmissionDraftApiTest(
...
+                        jsonObj {
+                            "type" to "Another"
+                        }
+                    "type" to "submission"
+                }.toString()
+            val response = webClient.submit(subV2)
+            assertThat(response).isSuccessful()
+
+            val updatedSubmission = response.body
+            assertThat(updatedSubmission.section.type).isEqualTo("Another")
```
Juan-EBI, line 289: should something happened here with the draft? like should it be deleted?

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/web/resources/SubmissionDraftResource.kt
```diff
@@ -82,28 +85,28 @@ internal class SubmissionDraftResource(
...
     ) {
         val buildRequest = SubmitBuilderRequest(user, onBehalfRequest, parameters)
-        val request = submitRequestBuilder.buildDraftRequest(key, user.email, buildRequest)
+        val request = submitRequestBuilder.buildDraftRequest(accNo, user.email, buildRequest)
 
         submitWebHandler.submitAsync(request)
     }
 
-    @PostMapping("/{key}/submit/sync")
+    @PostMapping("/{accNo}/submit/sync")
```
Juan-EBI, line 101: at the end how the accNo will be generated for a new draft?

Juan-EBI (commented):
Looking good please check comments/Questions

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/test/submission/submit/SubmissionDraftApiTest.kt
jhoanmanuelms: It should be marked as `PROCESSED`. The concept of deleting the draft after deletion no longer applies since now it's a single entity for the whole process. I didn't verify it because it's already checked in other tests.

submission/submission-webapp/src/main/kotlin/ac/uk/ebi/biostd/submission/web/resources/SubmissionDraftResource.kt
jhoanmanuelms: It hasn't changed, when it's a new draft the accNo will be `TMP_{Date}`. I'm just getting rid of the `key` word since that concept no longer applies.

PR #921, opened 2025-02-08
Pivotal ID # 188804317: Ignore Invalid Stats

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/188804317

- Ignore invalid stats entries
- Remove the response for the register and increment endpoints
- Stream the stats file instead of reading it all in order to be able to process big files
- Remove redundant tests
- Remove redundant dependencies


submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/service/StatsMongoDataService.kt
```diff
@@ -61,22 +68,24 @@ class StatsMongoDataService(
...
+            .groupBy { it.accNo.uppercase() }
+            .mapLatest { it }
+            .flatMapMerge { it }
+            .filter { submissionsRepository.existsByAccNo(it.accNo) }
+            .onEach { updateOrRegister(it) }
+            .collect()
+    }
 
-    override suspend fun incrementAll(stats: List<SubmissionStat>): List<SubmissionStat> =
+    override suspend fun incrementAll(stats: Flow<SubmissionStat>) {
```
Juan-EBI, line 81: I still have my doubsts about checking each stats exixts? like what is the business case we have that we select from a request what works and ignores others records?. Like everytime we recieve a list if we accept (200 response ) is because everything works but if it does not we fail all. 

accepting partially a request looks like a really extrange pattern as usually a customer request is either valid or invalid, not both. May you elaborate why we require this?

PR #923, opened 2025-02-18
Pivotal ID # 188526597: Deprecated Submission Versions Modification Date

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/188526597

- Avoid changing modification date for deprecated versions
- Add iTest


Juan-EBI (approved):
LGTM

PR #924, opened 2025-02-18
Pivotal ID # 188904728: Downloaded ftp binary files are corrupted

Juan-EBI:
https://www.pivotaltracker.com/story/show/188904728
- changed ftp client to use binary file type


submission/submission-config/src/main/kotlin/ac/uk/ebi/biostd/common/properties/TaskProperties.kt
```diff
@@ -7,7 +7,7 @@ import org.springframework.boot.context.properties.ConstructorBinding
 @ConstructorBinding
 @ConfigurationProperties
 data class TaskProperties(
-    val submissions: List<SubmissionId>,
+    val submissions: List<SubmissionId> = emptyList(),
```
Juan-EBI, line 10: added as calculate stats does not provide a list of submissions

jhoanmanuelms (approved):
LGTM

PR #926, opened 2025-02-19
Pivotal ID # 188522739: Submission stats include number of directories

Juan-EBI:
https://www.pivotaltracker.com/story/show/188522739
- added indexes


jhoanmanuelms (approved):
LGTM

PR #930, opened 2025-03-06
Pivotal ID # 187898907: Double Blind Review PageTab Generation

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/187898907

- Remove Author and Org sections from pagetab for private submissions
- Ext function to indentify Author and Org sections
- Reuse ext function in the DOI generation logic


submission/persistence-filesystem/src/main/kotlin/ac/uk/ebi/biostd/persistence/filesystem/pagetab/PageTabUtil.kt
```diff
@@ -20,28 +26,49 @@ class PageTabUtil(
...
-        sub: ExtSubmission,
+        extSub: ExtSubmission,
         target: File,
     ): PageTabFiles {
-        val element = toSubmissionMapper.toSimpleSubmission(sub)
-        val permissions = sub.permissions()
+        val sub = toSubmissionMapper.toSimpleSubmission(extSub)
+
+        if (extSub.released.not() && sub.reviewType == DOUBLE_BLIND) {
+            sub.section = filterBlindReview(sub.section) ?: Section()
```
Juan-EBI, line 35: This logic looks good but I think it does not bellow here, if you notice is in the toSimpleSubmission method that we filter the values not required by simple submission format, like removing public access tag. So I think we should keep that logic there.

WDYT?

Juan-EBI (approved):
LGTM please check minor comment

PR #933, opened 2025-03-18
Pivotal ID # 188930952: SUBMITTED Requests Not Showing In Submissions List

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/188930952

- Use the owner from the request itself to filter
- Use the submission from the draft when the process is still not set
- Make the draft not nullable in the request


submission/persistence-mongo/src/main/kotlin/ac/uk/ebi/biostd/persistence/doc/db/data/SubmissionRequestDocDataRepository.kt
```diff
@@ -241,7 +240,7 @@ class SubmissionRequestDocDataRepository(
 
     @Suppress("SpreadOperator")
     private fun createQuery(filter: SubmissionListFilter): Criteria =
-        where("$RQT_PROCESS.$SUB.$SUB_OWNER")
+        where(RQT_OWNER)
```
Juan-EBI, line 243: can we havea test for this. We can insert using persistence classes submission request in diferent stages and validate they are (and not) viisible when api call. 

Not a unit test, not a integration test but IMO best we can do

PR #934, opened 2025-03-18
Pivotal ID # 188973319: File presence check stage

Juan-EBI:
https://www.pivotaltracker.com/story/show/188973319
- added request properties to be able to recreate sources 
- added request file class to represent a file that has not been processed yet


jhoanmanuelms (approved):
LGTM

PR #935, opened 2025-03-20
Pivotal ID # 188973319: File presence check stage

Juan-EBI:
https://www.pivotaltracker.com/story/show/188973319
- added new stage


commons/commons-bio/src/main/kotlin/ebi/ac/uk/io/sources/FileSourcesList.kt
```diff
@@ -35,25 +36,56 @@ private val validPathPattern =
...
+        type: String,
+        attributes: List<ExtAttribute>,
+    ): ExtFile
+
+    fun sourcesDescription(): String
+
+    suspend fun getFileList(path: String): File?
+}
+
+class ByPassSourceList(
```
Juan-EBI, line 57: this is the source used for async support

commons/commons-bio/src/main/kotlin/ebi/ac/uk/model/RequestStatus.kt
```diff
@@ -4,6 +4,7 @@ enum class RequestStatus {
     DRAFT,
     SUBMITTED,
     REQUESTED,
+    FILES_VALIDATED,
```
Juan-EBI, line 7: the new stage

Juan-EBI:
@jhoanmanuelms what is pending after this, is include error field in submission list api response and create single submission status endpoint

PR #937, opened 2025-04-08
Pivotal ID # 189048268: Index new submission request

Juan-EBI:
https://www.pivotaltracker.com/story/show/189048268
- added missing indexes


jhoanmanuelms (approved):
LGTM

PR #939, opened 2025-04-08
Pivotal ID # 188773137: Remove Check Files Path Property

jhoanmanuelms:
https://www.pivotaltracker.com/story/show/188773137

- Remove checkFilesProperty feature flag
- The feature will always be enabled


Juan-EBI (approved):
LGTM

PR #941, opened 2025-05-02
Pivotal ID # 188976783: Links List PageTab Support

jhoanmanuelms:
- Replicate the File List feature to add support for the Link List so the PR is mostly boilerplate
- Add an iTest to test the Link List support


Juan-EBI (approved):
LGTM

PR #942, opened 2025-05-07
Pivotal # 188837373: Move Applications To Production NFS

jhoanmanuelms:
Fix BETA submitter deploy path


deploy.gradle
```diff
@@ -191,12 +191,18 @@ task deploy {
 
                 def user = execute "whoami"
                 def host = execute "hostname -s"
+                def filesPath = System.getenv("FILES_DIR_PATH")
```
Juan-EBI, line 194: why enviornment variable? can we simple not hardcode in each job config  as we other task param. I mean this is not a secret so IMO the simpler and easier is the best

deploy.gradle
```diff
@@ -191,12 +191,18 @@ task deploy {
...
+                def javaPath = System.getenv("JAVA_DIR_PATH")
+                def temp = "$filesPath/$env/tmp"
```
Juan-EBI, line 196: I would said that we should avoid calculated values and simple said that the tmp folder for beta is x, dev is y and prod is z.

Juan-EBI (commented):
please check comments

Juan-EBI (approved):
LGTM

PR #943, opened 2025-05-07
BIOSTD-21: FTP out file retrival support

Juan-EBI:
https://embl.atlassian.net/browse/BIOSTD-21
- added ftp out support


submission/submission-config/src/main/kotlin/ac/uk/ebi/biostd/common/properties/SecurityProperties.kt
```diff
@@ -26,14 +26,16 @@ data class FilesProperties(
     val defaultMode: StorageMode,
     val filesDirPath: String,
     val magicDirPath: String,
+    var userFtpDirPath: String,
```
jhoanmanuelms, line 29: it would be clearer to use `ftpInDirPath` for user files upload related code and `ftpOut` for the location where the submission files will be stored

submission/submission-webapp/gradle.properties
```diff
@@ -1,5 +1,5 @@
 coverage=0.7
-enableFire=true
+enableFire=false
```
jhoanmanuelms, line 2: This property should be set left as `true`

submission/submission-webapp/src/itest/kotlin/ac/uk/ebi/biostd/itest/itest/ITestListener.kt
```diff
@@ -54,7 +54,7 @@ class ITestListener : TestExecutionListener {
         mysqlContainer.stop()
         rabbitMQContainer.stop()
         fireServer.stop()
-        ftpServer.stop()
+        userFtpServer.stop()
```
jhoanmanuelms, line 57: Same comment about the ftp naming, I think we should use `FtpIn` and `FtpOut`

jhoanmanuelms (approved):
LGTM. Please check my comments specially the one about the test mode property

PR #1046, opened 2026-05-08
BIOSTD-276: Change Password Operation Should Not Recreate User Folder 

EliotRagueneau:
**Title**: Faster change password operation

* Checking creation of folder and setting chmod and group every 2s instead of every 30s (should be quite fast)
* Create in parallel the 2 magic folders (2 * 30s => 1 * 2s minimum time)
* Only run the activation when changing password if the user isn't already activated
* Update tests 

It should be fine to run the 2 folder creation in parallel because `mkdir -p` creates intermediate paths, so the order of execution is irrelevant.


submission/submission-security/src/main/kotlin/ebi/ac/uk/security/service/SecurityService.kt
```diff
@@ -233,8 +234,10 @@ open class SecurityService(
     }
 
     private suspend fun createFtpMagicFolder(ftpFolder: FtpUserFolder) {
-        createClusterFolder(ftpFolder.path.parent, UNIX_RWX__X___)
-        createClusterFolder(ftpFolder.path, UNIX_RWXRWX___)
+        coroutineScope {
+            launch { createClusterFolder(ftpFolder.path.parent, UNIX_RWX__X___) }
```
Juan-EBI: I think the folder creation is slow because job orchestation in the cluster rather than the folder creation itself, I think the real optimization here will be create both folders in a single cluster job

